PR #4, opened 2019-03-13
Temperature models

fhagemann:
Major update on the file `ChargeDriftModels.jl`

- Added new abstract type: `TemperatureModels{T}` 
to rescale the drift velocity to the crystal temperature based on a temperature model.
Every `TemperatureModels` contains the actual `temperature` of the crystal, the `reftemperature` at which referenced values were determined and fitting parameters for both electron and holes, and both the <100> and <111> axes. When defining a  `TemperatureModels`, the parameters are read from the (specified) `drift_velocity_config.json` file and depending on the model, the function `scale_to_given_temperature` will be used to scale the drift velocity in the functions `get_electron_drift_field` and `get_hole_drift_field`.
Four models are implemented so far:
  - `LinearModel.jl`: for rise times fitted with p1 + p2*T
  - `BoltzmannModel.jl`: for rise times fitted with p1 + p2 exp(-p3/T)
  - `PowerLawModel.jl`: for rise times fitted with p1*T^(3/2)
  - `VacuumModel.jl`: this does not rescale the drift velocity. This will be called as default if no temperature model is given.

- The actual `temperature` of the crystal can be passed directly when calling `ADLChargeDriftModel()`, otherwise it will be read from the `drift_velocity_config.json`.

- Default type for calculation in `ADLChargeDriftModel()` was changed from `Float64` to `Float32` to save computational time.

- The <110> axis will now be oriented according to the parameter `phi110` given in the `drift_velocity_config.json` files.

For more information about the physics behind the calculation, please refer to [this short overview and references](https://github.com/JuliaHEP/SolidStateDetectors.jl/files/2963423/TemperatureModels.pdf).


azsigmon:
Thank you for introducing the phi110 to the electrons. In the first version [here](https://github.com/mppmu/GePulseSimSandbox.jl/commit/63c883b89eea63bacc2e0806bdaa3e169dc7b20a#diff-c371fa9ba05510c141b78a81f533e16cR83) it was still there but then it disappeared in my changes.

We can merge from my side...

PR #14, opened 2019-10-30
Add SigGen read-in tool

ffischer42:
Added SigGenInterface.jl including two functions.
1. Open and read a SigGen *.config file and pass all parameteres to a dictionary
2. Take the dictionary and create the detector. Output is a dictionary comparable to the *.json files already in use

The public inverted Coax detector is added to the examples as *.config file with the name ":SigGen" and is now included to the runtest.


src/examples.jl
```diff
@@ -21,9 +21,9 @@ SSD_examples[:BEGe] = (
...
     joinpath(@__DIR__, "../examples/example_detector_config_files/public_CGD_config.json")
 )
-
 SSD_examples[:Spherical] = (
     joinpath(@__DIR__, "../examples/example_detector_config_files/public_spherical_detector_config.json")
 )
-
-
+SSD_examples[:SigGen] = (
+    joinpath(@__DIR__, "../examples/example_detector_config_files/public_ivc_config_SigGen.config")
```
lmh91: See first comment

examples/example_detector_config_files/public_ivc_config_SigGen.config
```diff
@@ -0,0 +1,51 @@
+# Example configuration file for fieldgen/siggen, by David Radford
```
lmh91: This is not an IVC detector. It's an PPC.
I guess you did this for the plot recipes, but I think we should name this correctly.
We should actually improve the general plot recipes instead.

ffischer42: Good point. II've changed everthing including the filename, example link and runtest to ppc.

PR #15, opened 2019-11-29
Lot's of changes

lmh91:


oschulz:
Hm, why do tests fail on 1.3?

Consider using Travis for 32-bit tests insteaf of Appveyor (see BAT.jl) - saves time, Tavis allows more parallel build jobs. See ValueShapes.jl for configuration.

lmh91:
> Hm, why do tests fail on 1.3?
> 
> Consider using Travis for 32-bit tests insteaf of Appveyor (see BAT.jl) - saves time, Tavis allows more parallel build jobs. See ValueShapes.jl for configuration.

I increased the default initial grid size for the field calculation in one of the commits. So the old test script might have run into memory / calculation time issues on the test servers. I hope i fixed that.

PR #41, opened 2020-04-13
Initial work on mesh support for primitives, added some examples

fraseyboo:
This is some initial work I've done in implementing mesh support for Solid State Detectors, it gets around the limitations caused by using primitives and can be used to accurately describe common features like bulletisation/chamfering more easily. I wrote it initially to simulate a complex & asymmetrical detector for my collaboration.  The current implementation is incredibly inefficient however looking at the operations it should be relatively straightforward to parallelize them. I've also included some examples of detectors that have been experimentally characterised & simulated at Liverpool in case additional validation work is needed. The exact details on CAO & impurities need to be added though.

The 'in' method works for complex geometry that can have convex and concave components and even handle missing and additional triangles present in the mesh (incomplete/overcomplete geometry). I've tested it on pretty standard geometries for detectors and also a 3D model of baby Yoda with multiple concave facets for geometric completeness.

Due to the inefficiencies in the code I'm not expecting this to get approved anytime soon, this is my first taste of Julia so it's more so I can get pointers on proper coding practices. 

ortho

proof1
proof2
visualisation-slices

yoda_test

Figures were made in my own visualiser which I've adapted to handle the rectilinear grids from this program.


lmh91:
Dear @fraseyboo,

as we have done a lot of changes regarding the geometry in SSD I will close this now.
But thanks for looking into this! This looks indeed quite interesting and to add support for loading geometries 
from formats like STL would be really cool. We will keep this in mind and might look into this again.

PR #43, opened 2020-04-27
[WIP] charge cloud size evolution

azsigmon:
Dear all,

Coming back to the diffusion model. The function in this PR does what siggen does for the holes but it can be run also for the electrons and the whole time evolution of the charge cloud size is saved.

The usage would be something like this in the Simulation.jl
`sigma_h = chargecloudsize(
    get_interpolated_drift_field(simulation.electric_field), 
    get_interpolated_drift_field(simulation.hole_drift_field), 
    simulation.charge_drift_model,
    event1.drift_paths[1].h_path, 
    event1.drift_paths[1].timestamps_h, 
    1,
    energy_dep1[1], 
    initial_size = 0.5u"mm", max_growth = 0.05u"mm", temperature = 77u"K"
)`

`sigma_e = chargecloudsize(
    get_interpolated_drift_field(simulation.electric_field), 
    get_interpolated_drift_field(simulation.electron_drift_field), 
    simulation.charge_drift_model,
    event1.drift_paths[1].e_path, 
    event1.drift_paths[1].timestamps_e, 
    -1,
    energy_dep1[1], 
    initial_size = 0.5u"mm", max_growth = 0.05u"mm", temperature = 77u"K"
)`

The output would be something like this for the holes in the example inverted coax detector
holeschargecloudsize
The breaking down at the end is due to drift within the contact, where the signal does not change anymore.

There are in siggen two ways to approximate the mobility which is visible in the function: `dv_dE` or `v_over_E`. The two are used for the different parts from repulsion and diffusion. Both ways are approximations but they make a difference in the result and we don't really know what is the truth. I left the alternative to use only one of the approximations commented out.

The next step would be to convolute the charge cloud size with the signal. I do not fully understand the siggen code but it is only using the final charge cloud size (maximum in the picture) only for the holes and convoluting with the whole signal. This is something we could improve and convolute time dependently for both holes and electrons. How this convolution is done, I don't know yet, so let me know if you already have some ideas.

After this we also have to integrate in the already created structure of `ChargeCloudModels` with all the necessary (initial size) and optional (temperature) inputs.

Thanks,
Anna

### DIFF
--- a/src/ChargeCloudModels/GaussianChargeCloud.jl
+++ b/src/ChargeCloudModels/GaussianChargeCloud.jl
@@ -1,0 +1,82 @@
+function chargecloudsize(
+        electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, 
+        velocity_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, 
+        cdm::AbstractChargeDriftModel{T}, 
+        drift_path::Vector{CartesianPoint{T}}, 
+        timestamps::Vector{T}, 
+        charge::RealQuantity,
+        energy::RealQuantity; 
+        initial_size::RealQuantity = 0.1u"mm", 
+        max_growth::RealQuantity = 0.05u"mm", 
+        temperature::RealQuantity = 77u"K"
+    )::Vector{T} where {T <: SSDFloat}
+
+    max_Δσ = to_internal_units.(internal_length_unit, max_growth)*10^9 #0.05 mm/ns
+    max_ΔσΔt = to_internal_units.(internal_length_unit, 0.1u"mm")
+    itemperature::T = ustrip(uconvert(u"K", temperature))
+    
+    n = length(timestamps)
+    σ::Vector{T} = zeros(T, n)
+
+    σ[1] = to_internal_units.(internal_length_unit, initial_size)
+    previous_vel = norm(get_velocity_vector(velocity_field, drift_path[1]))
+
+    ncharges =  to_internal_units(internal_energy_unit, energy) / to_internal_units(internal_energy_unit, hpge[:E_ionisation])
+    repulsion_factor = ncharges * elementary_charge / (4*π*hpge[:ϵ_r]*ϵ0) * 0.68^3 #charge in 1σ, unitless
+    
+    for i in 2:n
+
+        Δt = timestamps[i] - timestamps[i-1]
+        σ[i] = σ[i-1]
+
+        ####### acceleration part #######
+        vel = norm(get_velocity_vector(velocity_field, drift_path[i]))
+        σ[i] = σ[i-1] * vel / previous_vel
+
+        ####### repulsion part #######
+        Efield = get_velocity_vector(electric_field, drift_path[i])
+
+        v_over_E = vel / norm(Efield)
+        
+        Efield_2 = Efield.*T(1.30) #arbitrary choice of 30% increase of E
+        vel_2 = vel
+        
+        if charge < 0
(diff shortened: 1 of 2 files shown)


lmh91:
Dear Anna,

I tried it out but it did not worked. There are mistakes in it.
I guess you tried it out in the REPL and had some variables defined which are not defined in the Package. I fixed this for myself, but you also need to fix that on your branch.

Could you also provide some example code here?
E.g. how to fold (?) the output of your function with the waveform?

Thanks,
Lukas

azsigmon:
Dear Lukas,

I gave you access to my fork so you can push your fixes. Thanks!

I exactly uploaded it here to figure out together how we could convolute with the signals, because I don't really understand what siggen does and we should be able to do it time dependently.

Anna

lmh91:
Okay, I integrated your code into SSD. 

But something is wrong. I think there is a problem with some units.

E.g. one of your comments in the code is "k*T/e is in siggen 0.67". 
But in SI units this is:
```julia
PC.StefanBoltzmannConstant * 77u"K" / PC.ElementaryCharge
2.725160391255594e13 W C^-1 K^-3 m^-2
```

```julia
using Plots
pyplot()
using SolidStateDetectors
using SolidStateDetectors: get_interpolated_drift_field
using SolidStateDetectors: chargecloudsize
using Unitful

T = Float32

sim = Simulation(SSD_examples[:InvertedCoax]);
set_charge_drift_model!(sim, ADLChargeDriftModel())
simulate!(sim, max_refinements = 5)

evt = Event([CylindricalPoint{T}(0.01,0.0,0.01)], T[1460])
simulate!(evt, sim, Δt = 1u"ns", max_nsteps = 4000)

sigma_h = chargecloudsize( 
    get_interpolated_drift_field(sim.electric_field), 
    get_interpolated_drift_field(sim.hole_drift_field),
    sim.charge_drift_model, 
    evt.drift_paths[1].h_path, 
    evt.drift_paths[1].timestamps_h, 
    1, 
    evt.energies[1], 
    sim.detector.semiconductors[1].material, 
    initial_size = 0.5u"mm", 
    max_growth = 0.05u"mm"
)

plot(evt.waveforms[1], label = "waveform", lw = 1.0)
plot!(ustrip.(evt.waveforms[1].time)[1:end-1], evt.energies[1] .* sigma_h ./ maximum(sigma_h), 
    label = "scaled sigma_h", lw = 1.0)
```

The sigma_h should be in units of meter, right?
For me it returns something like >200m... 

What did you used for `boltzmann_constant`?

azsigmon:
Thanks Lukas!
I did check all the units and used your predefined elementary_charge etc.
The sigma output is in meter as on the plot it goes up to 1 mm that makes sense.

I put the Boltzmann constant by hand but of course converting to SI units
`const boltzmann_constant = 1.38064852e-23 #m^2 kg s^-2 K^-1"`
(forgot to include this next to the elementary_charge in the package)
then I get for this line
`boltzmann_constant / elementary_charge * 77 #m^2 kg s^-2 C^-1`
the output
`0.006635344913784144`
which is a factor hundred off from siggen explained by the different length units

lmh91:
Okay, my bad... I used the `StefanBoltzmannConstant` instead of the `BoltzmannConstant` as `kB`...
Now, reasonable numbers are produced.

azsigmon:
Thank you Lukas!

Next task is the folding with the signal
@oschulz you had some ideas how to convolute, right?

oschulz:
> @oschulz you had some ideas how to convolute, right?

Yes, I have something in the works.

oschulz:
> Yes, I have something in the works.

Sorry, misunderstanding, I meant something else.

PR #45, opened 2020-05-07
Add virtual volumes for arbitrary drift modulation

lmh91:
Add virtual volumes for arbitrary drift modulation e.g. to model dead layers and/or transition layers
(linear or whatever). Could also be used for drift modification in n+ layers. 

Working principle: During charge drift at each step it is evaluated whether the current position is inside a virtual volume. This is done with our CSG. If so, the drift vector is modulated by a function. 
For real dead layers it is implemented.
For other layers, like transition layer the user has to overwrite one function. This is because it is hart to generalize. 

In the json config file this looks like this:
```json
  {
            "type": "virtual_drift_volume",
            "name": "Top Dead Layer",
            "model": "dead",
            "geometry": {
                "type": "tube",
                "r": {
                    "from": 1.6,
                    "to": 29.2
                },
                "phi": {
                    "from": 0,
                    "to": 360
                },
                "z": {
                    "from": 52.0,
                    "to": 52.0001
                }
            }            
        },
        {
            "type": "virtual_drift_volume",
            "name": "Top Transition Layer",
            "model": "arbitrary",
            "geometry": {
                "type": "tube",
                "r": {
                    "from": 1.6,
                    "to": 29.2
                },
                "phi": {
                    "from": 0,
                    "to": 360
                },
                "z": {
                    "from": 51.0,
                    "to": 52.0
                }
            }            
        }
```

The function which has to be provided by the user could look like this:
```julia
function SolidStateDetectors.modulate_driftvector(
    sv::SolidStateDetectors.CartesianVector{T}, 
    cp::SolidStateDetectors.CartesianPoint{T}, 
    tl::SolidStateDetectors.ArbitraryDriftModificationVolume{T})::SolidStateDetectors.CartesianVector{T} where {T <: SolidStateDetectors.SSDFloat}
    t::T = tl.geometry.z_interval.right
    b::T = tl.geometry.z_interval.left
    scale_factor::T = (cp[3] - b) / (t - b)
    return sv * scale_factor
end
```

Here is the effect on the waveforms:
[linear_transition_layer_waveforms.pdf](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/4593453/linear_transition_layer_waveforms.pdf)

One problem right now is that it is not possible right now to add more than one virtual volume of model `arbitrary` because the modulation function provided by the user would be used for both...


lmh91:
One way to solve this problem with the multiple dispatch for the modulation function for serveral virtual volumes would be to use the name of each virtual volume (specified in the json file, e.g. `"name": "Top Transition Layer"`) and type it via `Val{Symbol(<name>)}` and add this as argument to modulation function.

Than the user could specify a modulation function for each "arbitrary" virtual drift volume. 

Or maybe add some ID for each volume.

azsigmon:
Dear Lukas,

Thank you!
So your function is only for dead layers in the z direction (in xy plane), right?

The Geant4 way to implement this would be to add to each grid point a "probability of drift" meaning dead layer has 0 and transition layer is a function of distance from the surface from 0 to 1 and then only those charges start to drift that are allowed.

This would be a faster implementation because you don't have to check in each step but of course cannot have charge trapping.

Anna

lmh91:
Well yes, the drift takes than a little bit longer (not much), but then you wouldn't need to rerun the G4Simulation when you want to 'fit' a dead layer variation / model to obervations.
Or do they actually postprocess the generated events and kind of delete some events in respect to their location. 

Yes, in this case its just a simple deadlayer in z underneath the top plate of the Ponama detector. 
One could easily add a radial / phi  dependence in this function. But its hard to generalize because the transition layer could be on a taper or on the mantle. So the direction would change. Thats why the user has to define the function.

I still think this belongs here and not in Geant4. Because as you said here you could also add partial charge recombination or charge trapping during the drift. E.g. if the charges drift upwards due to some charged passivation layer the probability also gets higher and the drift would also become slower. It just offers more options. Also randomization could be easily added.

I also though about using these volumes in the calculation of the drift field array. This would then just be done once and would not influence the performance. But this would reduce the amount of modification possibilities again I believe.

azsigmon:
> Or do they actually postprocess the generated events and kind of delete some events in respect to their location.

This is typically postprocessing step usually not done in Geant4 (there are also other solutions of course)

> Yes, in this case its just a simple deadlayer in z underneath the top plate of the Ponama detector.

Ponama is p-type, it has a much larger dead layer all over the n+ surface 

> One could easily add a radial / phi dependence in this function. But its hard to generalize because the transition layer could be on a taper or on the mantle. So the direction would change. Thats why the user has to define the function.

Don't we already have a function to ask how far we are from the surface?
For a tapered inverted coax with groove, your solution would mean 12 functions or so?

> I still think this belongs here and not in Geant4. Because as you said here you could also add partial charge recombination or charge trapping during the drift. E.g. if the charges drift upwards due to some charged passivation layer the probability also gets higher and the drift would also become slower. It just offers more options. Also randomization could be easily added.

I agree with you it is just not so clear for me if we need to mix dead layer effects with charge trapping effects.

> I also though about using these volumes in the calculation of the drift field array. This would then just be done once and would not influence the performance. But this would reduce the amount of modification possibilities again I believe.

I think then we cannot have a transition layer where we randomly decide for each event the amount of charge that drifts.

Let's think a little bit how to get this in a more userfriendly way!

schustermartin:
I am also in favor of this approach, since it provides real flexibility, you can in principle define volumes for special charge drift treatment with many purposes, not just to simulate dead layers. This is also what I had in mind to deal with / get a further handle on near-surface charge drift. For that, being able to define multiple volumes with individial handling would be beneficial/necessary.

azsigmon:
> I am also in favor of this approach, since it provides real flexibility, you can in principle define volumes for special charge drift treatment with many purposes, not just to simulate dead layers. This is also what I had in mind to deal with / get a further handle on near-surface charge drift. For that, being able to define multiple volumes with individial handling would be beneficial/necessary.

I am just trying to think about something that legend people can actually use. For p-type detectors they care much less about charge trapping. For them something "simple" has to be defined automatically for the non-segmented geometries with a few input parameters. This would not exclude special treatment of volumes for charge trapping.

oschulz:
> For other layers, like transition layer the user has to overwrite one function. 

Maybe we could provide such a function for typical cases?

lmh91:
I added an ID so one can now add and use several volumes.

lmh91:
> I am just trying to think about something that legend people can actually use. For p-type detectors they care much less about charge trapping. For them something "simple" has to be defined automatically for the non-segmented geometries with a few input parameters. This would not exclude special treatment of volumes for charge trapping.

I fully agree. But I still want to be able to do all this and not be limited to simplicity. 
I guess, like Oli just said, we should provide some function which generates all the needed functions for the LEGEND detectors. Maybe in combination with the SigGen read-in.

azsigmon:
> we should provide some function which generates all the needed functions for the LEGEND detectors. Maybe in combination with the SigGen read-in.

sounds reasonable to me :+1:

lmh91:
So can I merge?

lmh91:
Since this is just an additional feature I will merge now. 
It is not yet documented anyhow and if so we can state that it is experimental.

PR #49, opened 2020-05-25
Fixes to in() functions of cone and tube primitives; Updates/Adding of some plot recipes; Tweaks to the Logo;

schustermartin:
Fixed some edge cases where `in(...) `function was returning false results due to forth and back conversion between Cylindrical and Cartesian Points.
This is done by skipping trigonometric conversions where unnecessary.
Fixed `in(...)` function for the `Cone` Primitive. So far translate vector was not taken into account.

 Fixed and updated drawing of electric field lines. There is now a separate plot recipe `plot_electric_fieldlines(...)`, which adds modularity.

Tweaks to the Logo:
    - Added white margin around the "D" part of the logo to establish consistency.
    - Added keyword to toggle transparent background.
    - Pushed the "etectors" text a tad to the right to improve readability.
    - For the .gifs, the first frame (frame 0) now shows the last frame. (Useful for export to .pdf)
    - Updated Logos


lmh91:
I will close this. 
Please make individual PRs.

PR #50, opened 2020-06-03
Added CZT simulation and fixes in LinearChargeDensityModel

fhagemann:
- Added CZT material properties from [here](http://www.umich.edu/~ners580/ners-bioe_481/lectures/pdfs/SemiConductor-material_prop.pdf) and a config file for a 11x11 pixelated CZT array. 
- Fixed the unit conversion in the `LinearChargeDriftModel` (gradients have the unit `length^{-4}` and not `length^{-3}` and should be converted accordingly)


fhagemann:
Then let us merge!

PR #55, opened 2020-09-18
Added inner_taper_width

ffischer42:
The inner_taper_width is now calculated when a taper_angle is given.


lmh91:
Can you please use the Julia function `deg2rad` (or respectively `rad2deg`) instead of 2pi*x/360.

PR #71, opened 2020-10-16
Hexagonal primitive

leenderthayen:
Hi,

This is just the implementation of a hexagonal primitive, allowing for an inner cutout.
New file HexagonalPrism.jl and test configuration file. 
Plotting, electric potential, field, and drift has been tested.

Leendert

### DIFF
--- a/examples/example_detector_config_files/minimum_hexagon_config.json
+++ b/examples/example_detector_config_files/minimum_hexagon_config.json
@@ -1,7 +1,7 @@
 {
     "name": "Simple Hexagon",
     "units": {
-        "length": "mm",
+        "length": "m",
         "angle": "deg"
     },
     "grid": {
@@ -41,12 +41,12 @@
             "bulk_type": "n",
             "geometry": {
                 "type": "HexagonalPrism",
-                "a": 3,
-                "h": 3,
+                "a": 1,
+                "h": 1,
                 "translate": {
-                  "x": 8,
-                  "y": 5,
-                  "Z": 2
+                  "x": 0,
+                  "y": 0,
+                  "z": 0
                 }
             }
         }
--- a/examples/example_detector_config_files/minimum_hexagon_config.json
+++ b/examples/example_detector_config_files/minimum_hexagon_config.json
@@ -10,24 +10,24 @@
 		"coordinates": "cartesian",
 		"axes": {
             "x": {
-								"from": -5,
-                "to": 5,
+								"from": -2,
+                "to": 2,
                 "boundaries": {
                     "left": "inf",
                     "right": "inf"
                 }
             },
             "y": {
-                "from": -5,
-                "to": 5,
+                "from": -2,
+                "to": 2,
 								"boundaries": {
                     "left": "inf",
                     "right": "inf"
                 }
             },
             "z": {
-                "from": -5,
-                "to": 5,
+                "from": -1,
+                "to": 3,
                 "boundaries": {
                     "left": "inf",
                     "right": "inf"
@@ -46,21 +46,21 @@
 				"name": "linear",
 				"r": {
 					"init": 0.0,
-					"gradient": 0.0
+					"gradient": 1e-12
 				},
 				"phi": {
 					"init": 0.0,
 					"gradient": 0.0
 				},
 				"z": {
-					"init": 1e-7,
-					"gradient": 0.0
+					"init": 1e-12,
+					"gradient": 1e-13
 				}
 			},
 			"geometry": {
 				"type": "HexagonalPrism",
 				"r": {
-					"from": 0.5,
+					"from": 0.0,
 					"to": 1.0
 				},
 				"h": 2.0,
@@ -78,12 +78,12 @@
 			"geometry": {
 				"type": "HexagonalPrism",
 				"r": {
-					"from": 0.5,
+					"from": 0.,
 					"to": 1.0
 				},
-				"h": 1e-6,
+				"h": 1e-2,
 				"translate": {
-					"z": 2.0
+					"z": 2.005
 				}
 			}
 		},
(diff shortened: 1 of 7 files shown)


src/Geometries/VolumePrimitives/HexagonalPrism.jl
```diff
@@ -0,0 +1,148 @@
+struct HexagonalPrism{T} <: AbstractVolumePrimitive{T, 3} ## Only upright hexagons at the moment
+    rInner::T
+    rOuter::T
+    h::T #total height
+    translate::Union{CartesianPoint{T}, Missing}#origin at middle of hexagonal prism
```
lmh91, line 5: Use the type CartesianVector for translate

src/Geometries/VolumePrimitives/HexagonalPrism.jl
```diff
@@ -0,0 +1,148 @@
...
+end
+
+# Auxiliary function for point in a triangle check
+function tri_area(p1::CartesianPoint{T}, p2::CartesianPoint{T}, p3::CartesianPoint{T})::T where {T}
+    return (p1.x - p3.x) * (p2.y - p3.y) - (p2.x - p3.x) * (p1.y - p3.y)
+end
+
+# is a point inside the hexagonal prism?
+function in(pt::CartesianPoint{T}, hp::HexagonalPrism{T})::Bool where {T}
+    pt = ismissing(hp.translate) ? pt : pt .- hp.translate
```
lmh91, line 36: When `hp.translate` is a CartesianVector you do not need the broadcast `.-` but only `-`

src/Geometries/VolumePrimitives/HexagonalPrism.jl
```diff
@@ -0,0 +1,148 @@
...
+    return geom_round.(T[-hp.h/2, hp.h/2])
+end
+# X
+function get_important_points(hp::HexagonalPrism{T}, ::Val{:x})::Vector{T} where {T <: SSDFloat}
+    return geom_round.(T[-hp.rOuter, -sqrt(3)/2*hp.rOuter, -hp.rInner, -sqrt(3)/2*hp.rInner, sqrt(3)/2*hp.rInner, hp.rInner, sqrt(3)/2*hp.rOuter, hp.rOuter])
+end
+# Y
+function get_important_points(hp::HexagonalPrism{T}, ::Val{:y})::Vector{T} where {T <: SSDFloat}
+    return geom_round.(T[-hp.rOuter, -sqrt(3)/2*hp.rOuter, -hp.rInner, -sqrt(3)/2*hp.rInner, sqrt(3)/2*hp.rInner, hp.rInner, sqrt(3)/2*hp.rOuter, hp.rOuter])
+end
```
lmh91, line 131: The translate vector is not taken into account.

lmh91 (changes requested):
Nice work 👍 

I added some comments regarding the translation of the primitive.
I simulated your example detector but the second contact was not recognized. 

It is just an example, but is it intendet that the semiconductor has a hole?

leenderthayen:
Thanks for looking at this so quickly. I'll take a look at the point map and resolve the translations. The hole in the contact was just for testing purposes and does not represent something physical. We were interested in having hexagonal rings around pixels to represent passivated inter-pixel material, hence the cutout. I'll get back to you.

leenderthayen:
The 'in' and get_important_points are now fixed. The upper contact is now being recognized (I think it was because the translation vector was not taken into account in get_important_points for z), and everything should be nicely hexagonal now.

ep_hexagon
hex_wp

examples/example_detector_config_files/minimum_hexagon_config.json
```diff
@@ -0,0 +1,109 @@
...
+		{
+			"type": "contact",
+			"material": "Si",
+			"name": "n+ contact",
+			"potential": 0.0,
+			"channel": 1,
+			"geometry": {
+				"type": "HexagonalPrism",
+				"r": {
+					"from": 0.,
```
lmh91, line 81: Actually, one of the two contacts should have some inner radius > 0, e.g. 0.4. 
For reasons of testing.

lmh91 (changes requested):
I think everything is correct now. 

Yes, the second contact was not recognized as the `get_important_points` methods
did not took the translation into account and no (correct) grid ticks were generated where the second contact was. 

Please just add an finite inner radius to the one contact again.

lmh91 (approved):
Alright, then I will merge. 

Thank you for this nice contribution :)
And please report issues if you encounter some.

src/Geometries/VolumePrimitives/HexagonalPrism.jl
```diff
@@ -0,0 +1,144 @@
...
+
+#and a sample function to paint the primitive on the grid (necessary if the object is small)
+function sample(hp::HexagonalPrism{T}, stepsize::Vector{T})  where T
+    samples = CartesianPoint{T}[]
+    for x in -hp.rOuter : stepsize[1] : hp.rOuter
+        for y in -hp.rOuter : stepsize[2]: hp.rOuter
+            for z in -hp.h/2 : stepsize[3] : hp.h/2
+                p = CartesianPoint{T}(x, y, z)
+                if in(samples, p)
+                    push!(samples, p)
```
fhagemann, line 137: @leenderthayen: I don't understand this two lines of code. 
This `if`-statement is always false (`samples` cannot be in `p`).

If you are trying to only `push!`, if `p` is in `samples`, you could use:
```julia
if in(p,samples)
    push!(samples,p)
end
```

However, as `samples` is empty at the beginning, this will also never happen.

Did you intend to write:
```julia
if !in(p,samples)
    push!(samples,p)
end
```
?

fhagemann (commented):
Please review the `samples` function.

src/Geometries/VolumePrimitives/HexagonalPrism.jl
lmh91: Hmm, actually I believe it should be 
```julia
if p in hp push!(samples, p) end
```
?

lmh91: PR: #86

Issue #59, opened 2020-10-09
Refactor plot recipes - GR Support


lmh91:

fhagemann:
I have already spent some time to refactor the plot recipes to support the GR backend.
However, when adapting the plot recipes, they might not work as intended for PyPlot anymore.
(example: polar plots expect the angle in degree in GR and in radiants in PyPlot)\

If we prioritize fast support of GR, some plots might not be producible with PyPlot.
How should we proceed here?

fhagemann:
This issue is related to #65.

PR #76, opened 2020-10-23
Update plot recipes to fix issues with doc build

fhagemann:
The old plot recipes were removed and replaced by new ones.
So far, there are plot recipes for `SolidStateDetector`, `Contact`, `Tube`, `Cone`, `LineSegment`.

As most of the code for plot recipes for scalar potentials is identical, a generic plot recipe was defined for `ScalarPotential`.
Specific requirements (e.g. title, colorbar limits or colors) are saved in the potential-specific plot recipes.

The standard backend was changed to `GR` for a speed-up on the plot output.
For this, also a new struct to plot (partial) circles, `PartialCircle` was introduced with a corresponding plot recipe.

This set of plot recipes is not complete, but it looks like it can plots all detectors from the `SSD_examples` reasonably.
More work will follow within the next weeks.


fhagemann:
Fixes #59.

oschulz:
Thanks!!

fhagemann:
The docs were built, but the titles and axes labels on the plots seem to be placed (partly) outside the canvas or are hidden (partly) by the heatmap. We might need a work around for this.

PR #80, opened 2020-10-28
Update on plot recipes

fhagemann:
This pull request addresses three encounters in the plot recipes:
 - Minor fix in the `Cone` plot recipe
 - Adding plot recipes for the scalar potentials in Cartesian coordinates which were removed in #76 and are now reintroduced
 - For the scalar potentials: if all entries of the heatmap were identical, GR did not manage to display the plot.
   This issue was [fixed](https://github.com/JuliaPlots/Plots.jl/pull/3097) and will be part of the new Plots v1.7.4 release. The line that manually set the color bar limits is removed.

Please do not merge before the Plots 1.7.4 release!


fhagemann:
We have decided on keeping the manual `clims` in this pull request until the new Plots.jl release.
The manual `clims` can be removed in #90.

PR #81, opened 2020-10-28
Fix radial charge density gradients in cartesian space

### LINKED ISSUE #75: Radial charge density gradients in cartesian space
Hi,

It seems like when specifying a radial charge distribution gradient in a cartesian grid, it takes the 'r' argument and turns it into an x gradient instead.

radial_charge_density_error

### LINKED ISSUE #79: Fix for hexagon primitive plot recipe
Hi,

I noticed a mistake in the hexagon primitive that was added recently and I thought it would be easier to fix through here than a full pull request.

In HexagonalPrism.jl line 76 should be
`for φ in [deg2rad(30),deg2rad(90), deg2rad(150), deg2rad(210), deg2rad(270), deg2rad(330)]`
Currently the plot recipe is rotated 30 degrees relative to the rest of the geometries functions.

For example,
image
versus
image

fhagemann:
This pull request is a follow-up to #80 to fix #75.

A new field to the `LinearChargeDensityModel` was added, `CS`, which saves the coordinate system of the gradient. This should be either `:cartesian` or `:cylindrical`. In the detector config file, the coordinate system is determined from the coordinates that define the charge density.

If only `r` (and `z`) are given, it is assumed to be a `:cylindrical` geometry, whereas if `x` or `y` are passed, a `:cartesian` geometry is assumed. If both `r` and either `x` or `y` are passed, this results in an `ErrorException` which asks to adjust the config file.

This might add a conversion of `CylindricalPoint` to `CartesianPoint` and vice versa. Regarding the fact that we want to refactor the update step of the boundary conditions and pass both the `CylindricalPoint` and `CartesianPoint` to the functions, this might be easily incorporated then.

Do not merge before #80 is merged!

### DIFF
--- a/src/ChargeDensityModels/ChargeDensityModels.jl
+++ b/src/ChargeDensityModels/ChargeDensityModels.jl
@@ -14,15 +14,18 @@ slopes in each dimension.
 struct LinearChargeDensityModel{T <: SSDFloat} <: AbstractChargeDensityModel{T}
     offsets::NTuple{3, T}
     gradients::NTuple{3, T}
+    CS::Symbol
 end
 
 function get_charge_density(lcdm::LinearChargeDensityModel{T}, pt::AbstractCoordinatePoint{T})::T where {T <: SSDFloat}
+    pt::AbstractCoordinatePoint{T} = (lcdm.CS == :cylindrical) ? CylindricalPoint(pt) : CartesianPoint(pt)    
     ρ::T = 0
     for i in eachindex(lcdm.offsets)
         ρ += (lcdm.offsets[i] + pt[i] * lcdm.gradients[i]) #* T(1e16) # * T(1e10) * T(1e6) -> 1/cm^3 -> 1/m^3
     end
     return ρ
 end
+
 
 """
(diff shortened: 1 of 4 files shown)


src/ChargeDensityModels/ChargeDensityModels.jl
```diff
@@ -14,16 +14,19 @@ slopes in each dimension.
 struct LinearChargeDensityModel{T <: SSDFloat} <: AbstractChargeDensityModel{T}
     offsets::NTuple{3, T}
     gradients::NTuple{3, T}
+    CS::Symbol
 end
```
lmh91, line 18: Use a parametric type instead of an additional field:
`struct LinearChargeDensityModel{T <: SSDFloat, CS <: {:cylindrical, :cartesian}}`

fhagemann: Thanks, @lmh91 for the input. I have changed the code accordingly.

fhagemann:
Effective charge density inside the CGD detector, when `r` and `z` are passed:

Cylindrical_LCDM_new

Effective charge density inside the CGD detector when `x`, `y` and `z` are passed:

Cartesian_LCDM_new

lmh91:
I close this because we decided for #88

PR #82, opened 2020-10-28
Hexagon plot recipe fix

rjtayl:
Rotates the hexagon plot recipe by 30 degrees so that it agrees with the other functions in HexagonalPrism.jl


fhagemann (approved):
This looks fine to me.
I saw that the hexagons are only upright so far, but are there any plans on making these hexagonal primitives rotational? That might influence the plot recipes as well.

lmh91:
Well, first we need to add rotation of objects in general. For now this is totally fine. 
Thanks for the fix!

PR #88, opened 2020-11-04
Introduce new ChargeDensityModel that separates Cartesian and cylindrical gradients

### LINKED ISSUE #75: Radial charge density gradients in cartesian space

fhagemann:
This pull request is an alternative to #81 to fix #75.

A new struct `RadialChargeDensityModel` was added which is used when the impurity density changes linearly with `r` and/or `z`. Thus, it is the cylindrical equivalence to `LinearChargeDensityModel` which now only considers `x`, `y` and `z`.
Warnings are added, if `LinearChargeDensityModel` is parsed with `r` or if `RadialChargeDensityModel` is parsed with either `x` or `y`.

This might add a conversion of `CylindricalPoint` to `CartesianPoint` and vice versa. Regarding the fact that we want to refactor the update step of the boundary conditions and pass both the CylindricalPoint and CartesianPoint to the functions, this might be easily incorporated then.

Do not merge before #80 is merged!


lmh91:
Hmm, I am really unsure about the naming... 
I know I kind of suggested `RadialChargeDensityModel` but I would actually go for something like
* `StandardCrystalImpurityDensity`
* `SimpleCrystalImpurityDensity`
* `LinearCzochralskiImpurityDensity`
* `LinearCylindricalImpurityDensity`

as most users actually only use the gradient in `z` which is not mentioned in the name at all.

I would remove `Model` in general to shorten the name.

* `Impurity` vs `Charge`: In principle I think impurity would be better but as we also use those models also for charged up volumes `charge` would be better. The proper way would be to implement both though, what kinda complicate everything 
but is in principle the correct way. Charge density model would then return not a unitless number but the impurity value times the electric charge.

fhagemann:
If only a linear gradient in `z` is assumed, this is still perfectly covered by `LinearChargeDensityModel`, i.e. `"linear"` in the config file.
Thus, `RadialChargeDensityModel`, i.e. `"radial"` in the config file, should only be used if (in addition to the `z` gradient) a radial dependency is passed.

lmh91:
Sure, this would be possible but it would not be well defined. 
And I am not talking here about that we can identify the correct model by the fields inside the config file.

The `LinearChargeDensity` should only take `x`,`y` and `z` in my opinion.

And I would change `RadialChargeDensity` into `CylindricChargeDensity`

`Impurity` vs `Charge` we should put into another PR.

fhagemann:
I agree.  I'll change the name to `CylindricalChargeDensity` (with `al` to match the naming of `CylindricalPoint`) and remove the `Model` ending of all impurity density models.

fhagemann:
Removing `Model` from the `ChargeDensityModels` will result with a conflict between:
- the functions `ChargeDensityModel` in [ChargeDensityModels.jl](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/master/src/ChargeDensityModels/ChargeDensityModels.jl) which parse the charge density models from the config file 
- the struct `ChargeDensity` which saves the effective charge density in [ChargeDensity.jl](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/master/src/PotentialSimulation/ChargeDensity.jl).

fhagemann:
We could rename the struct `ChargeDensity` to `EffectiveChargeDensity` as this is what is actually stored. Any thoughts on this?

lmh91:
Yes, rename it to `EffectiveChargeDensity`

fhagemann:
From my side this is ready to merge.

PR #96, opened 2020-11-12
Some minor bug fixes and some cleaning

schustermartin:
- Fixed some edge cases where in(...) function was returning false results due to forth and back conversion between Cylindrical and Cartesian Points. This is done by skipping trigonometric conversions where unnecessary.
- Fixed in(...) function for the Cone Primitive. So far translate vector was not taken into account.
- Switched to the use of RotationsPackage for VolumePrimitives (not used so far, just how general structure could look like).
- Got rid of some unnecessary lines of code (expressing the same in less lines).

### DIFF
--- a/src/Geometries/VolumePrimitives/Tube.jl
+++ b/src/Geometries/VolumePrimitives/Tube.jl
@@ -30,14 +30,14 @@ function Tube{T}(dict::Dict{Any, Any}, inputunit_dict::Dict{String,Unitful.Units
     else
         @warn "please specify a height of the Tube 'h'"
     end
-    
-    phi_interval =  if haskey(dict, "phi") 
+
+    phi_interval =  if haskey(dict, "phi")
         Interval(geom_round(T(deg2rad(dict["phi"]["from"]))), geom_round(T(deg2rad(dict["phi"]["to"]))))
     else
         Interval(geom_round(T(deg2rad(0))), geom_round(T(deg2rad(360))))
     end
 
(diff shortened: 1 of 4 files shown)


src/Geometries/VolumePrimitives/Tube.jl
```diff
@@ -56,20 +56,14 @@ function Geometry(T::DataType, t::Val{:tube}, dict::Dict{Union{Any,String}, Any}
 end
 
 function in(point::CartesianPoint{T}, tube::Tube{T}) where T
-    ismissing(tube.translate) ? nothing : point -= tube.translate
+    (ismissing(tube.translate) || tube.translate == CartesianVector{T}(0.0,0.0,0.0)) ? nothing : point -= tube.translate
```
fhagemann, line 59: `tube.translate` should never be zero

fhagemann (approved):
Right now, the `translate` vector of `Tube` should never be zero. But we can leave it like this.

src/Geometries/VolumePrimitives/Tube.jl
lmh91: Well, we have to change the general struct of our primitives anyhow when we add Rotations. 
This `if ismissing` should never appear. I guess we will use parametric types to handle this on compilation level.

lmh91: For both, translation and rotation.

lmh91 (approved):

PR #97, opened 2020-11-12
Minor updates on the Geometry Primitives

fhagemann:
- added rotations around the `z`-axis to `HexagonalPrism` 
(right now with a scalar, should be replaced by a rotation matrix at some point)
- added some consistency in the `Tube` and `Cone` code 
- minor fix in `get_important_point` in the `Sphere` primitive

### DIFF
--- a/docs/src/man/primitives.md
+++ b/docs/src/man/primitives.md
@@ -2,7 +2,7 @@
 
 CSG = Constructive Solid Geometries
 
-## Primitve: Box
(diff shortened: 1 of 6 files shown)


lmh91 (approved):
Don't spend to much time on tuning the primitives. 
Also don't add partial rotations in general.

We should talk at some point how to change the primitives in general to enable
fast rotation and translation. I have something in mind already.

PR #100, opened 2020-11-13
Fix display of non-uniform polar plots in GR

fhagemann:
Due to the adaptive grid, the axes of the grid will almost always not be uniform. This leads to non-uniform polar plots when plotting cross sections in `z` on a cylindrical grid. These plots look fine when using the `pyplot` backend. Here is an example:

segBEGe_pyplot

However, when using the `gr` backend, non-uniform plots are displayed as if the axes were uniform. Thus, regions with high grid point densities are stretched and regions with low grid point densities are compressed. The same polar plot looks like this in `gr`:

segBEGe_GR

Thus, a new keyword, `resample = true` is added. It manually resamples the data array for polar plots in a cylindrical coordinate system. The result from this is comparable to the one from `pyplot`:

segBEGe_GR_new

Whenever polar plots are created, an `@info` is displayed that data is resampled and how many points are used for the resampling.
If no resampling is wished, it can be omitted by passing the keyword `resample = true` when calling the `plot()` command.
All of this can be removed, if nonuniform polar plotting is implemented in `GR`.

Minor changes:
    - A `WeightingPotential` can now also be displayed with equipotential lines when called with `contours_equal_potential = true`
    - I have also updated the plot recipes for the `Grid`, as it gave errors when plotting axes with only one tick (two-dimensional configurations).

### DIFF
--- a/src/Grids/Grids.jl
+++ b/src/Grids/Grids.jl
@@ -96,15 +96,18 @@ end
 include("RefineGrid.jl")
 
 @recipe function f(grid::Grid{T, N, S}) where {T, N, S}
-    layout --> N
-    st --> :stephist
+    
+    #only plot grid point densities for axes with more than one tick
+    iaxs = findall(ax -> length(ax) > 1, grid.axes)
+    layout --> length(iaxs) 
+    seriestype --> :stephist
     legend --> false
 
-    for iax in 1:N
+    for (i,iax) in enumerate(iaxs)
         @series begin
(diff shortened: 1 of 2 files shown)


fhagemann:
It looks like this will be introduced with the new GR and Plots release:
https://github.com/jheinen/GR.jl/commit/9b1dd41252bdf188539bebc5262765b024d73bd4

We can delete the workaround then

PR #106, opened 2020-11-18
Toroidalprimitive

hervasa2:
**Added Torus Primitive
Added Rotation support to PartialCircle LinePrimitive. The plot recipe of Torus depends on this.**
We should discuss how to add Rotation support in general. From what I see rotations in the Cone primitive only allow a for a rotation matrix of type RotXYZ, which will rotate in the order Z->X->Y. We need all permutations of this and thus it requires more user input (order choice). We also need to allow a rotation of type AngleAxis (what I used in PartialCircle). 
Examples of the Torus Primitive follow, including usage of the sample function and potential calculations:
**Whole coaxial torus, added to example config files**
torus_whole
torus_whole_plots
**Coaxial torus, but with a partial thick n contact section**
torus_section_whole
torus_section_whole_plots
**A bulletized PPC**
bulletized_PPC
bulletized_PPC_plots

### DIFF
--- a/examples/coaxial_torus_config.json
+++ b/examples/coaxial_torus_config.json
@@ -1,0 +1,118 @@
+{
+    "name": "Coaxial Torus",
+    "units": {
+        "length": "mm",
+        "angle": "deg",
+        "potential": "V",
+        "temperature": "K"
+    },
+    "grid": {
+        "coordinates": "cylindrical",
+        "axes": {
+            "r": {
+                "to": 20.0,
+                "boundaries": "inf"
+            },
+            "phi": {
+                "from": 0,
+                "to":0,
+                "boundaries": "periodic"
+            },
+            "z": {
+                "from": -10,
+                "to": 10.0,
+                "boundaries": {
+                    "left": "inf",
+                    "right": "inf"
+                }
+            }
+        }
+    },
+    "medium": "vacuum",
+    "objects": [
+        {
+            "type": "semiconductor",
+            "material": "HPGe",
+            "bulk_type": "p",
+            "temperature": 77.0,
+            "charge_density_model": {
+                "name": "linear",
+                "r": {
+                    "init": 0,
+                    "gradient": 0
+                },
+                "phi": {
+                    "init": 0.0,
+                    "gradient": 0.0
+                },
+                "z": {
+                    "init": -1e7,
+                    "gradient": 0.0
+                }
+            },
+            "geometry": {
+                "type": "torus",
+                "c": 10.0,
+                "a": {
+                    "from": 2.0,
+                    "to": 5.0
+                },
+                "phi": {
+                    "from": 0.0,
+                    "to": 360.0
+                },
+                "theta": {
+                    "from": 0.0,
+                    "to": 360.0
+                }
+            }
+        },
+
+        {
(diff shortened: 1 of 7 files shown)


examples/coaxial_torus_config.json
```diff
@@ -0,0 +1,118 @@
...
+                    "gradient": 0.0
+                }
+            },
+            "geometry": {
+                "type": "torus",
+                "c": 10.0,
+                "a": {
+                    "from": 2.0,
+                    "to": 5.0
+                },
```
lmh91, line 59: Can you explain me what `c` and `a` is? I expected something like `r1` and `r2`?

hervasa2: `c`(radius from the center of the hole to the center of the torus tube) and `a`(tube radius) are described in lines 23 and 47 respectively. I reviewed some literature to see what is commonly used and ended up picking the notation from [wolfram](https://mathworld.wolfram.com/Torus.html).  I think we should not use `r1` and `r2` because it could get confused with coordinate `r`

schustermartin: Is that not precisely why we should call it something like `r1` and `r2`  because it belongs to the `r` direction of the grid. I would prefer something expressive, `a`, `b`, and `c` are very generic and you have to consult a manual to find out what to put there.

hervasa2: `r_1` I guess would but then `r_2` would not. How about `r_tube` and `r_center`?

lmh91: So in general try to go along with OpenSCAD / OpenCSG / Geant4. 
And I strongly prefer something with `r` at these are radii. 
In the config file we can also allow for different notations. Like also r_center and r_tube.
But please add also a documentation to it on top of the struct, like:
```julia
"""
    # struct Torus{T}

`r_1`: ....
`r_2`: ...
...
"""
struct Torus....
    ....
end
```

I am also suspicious about the `a_interval`?

hervasa2: `a_interval` would be converted to `r_tube_interval` and `c` to `r_center`

lmh91: I don't understand how a Torus has a fixed interval in `r` as this interval depends on `z`.
Or am I missing something?

lmh91: Ah, is it that the torus can have a "hole". So like a tube and not a cylinder?

schustermartin: Yes, I believe that's how it works!
torus

lmh91 (changes requested):
Very nice work!
But please rename `c` and `a` in the config files to something with `r`.
E.g. `r_torus` and `r_tube`. Or just `r1` and `r2`. In the config file we can support several notations. The internal structure we can change later as it does not affect the general user.

hervasa2:
I've renamed the variables everywhere and also added documentation and general rotational matrix type support in PartialCircle. Also added CoaxialTorus to SSD_Examples

schustermartin (approved):
Very nice Job :)

fhagemann (approved):

lmh91 (approved):
:+1:

PR #115, opened 2020-11-28
Refactor csg

fhagemann:
- Added `Cone`, `Box`, `Sphere` and `HexagonalPrism` primitives with the respective constructors and `in()` methods.
- Expanded `CSG_text.jl` to test the new primitives.
- Added `geom_round` to the `ConstructiveSolidGeometry` module.
- First work on the `get_important_points` and `sample` for the `HexagonalPrism` primitive. 
Please have a look at the code before I start implementing it for the other primitives.
HexagonalPrism_sample


src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -0,0 +1,8 @@
+@inline _left(r::T) where {T <: Real} = -r
```
lmh91, line 1: Shouldn't this be 0?

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -0,0 +1,8 @@
...
+@inline _left(r::AbstractInterval) = r.left
+
+@inline _right(r::T) where {T <: Real} = r
+@inline _right(r::AbstractInterval) = r.right
+
+@inline _width(z::T) where {T <: Real} = T(2) * z
```
lmh91, line 7: I believe the compiler is clever enough that you can just right `2z`. 
Right @oschulz ?

src/ConstructiveSolidGeometry/Intervals.jl
lmh91: also, `_left(r::T)` in general is not good. It depends on what `r` is. Like it's different for `x` and `r`

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -22,36 +22,30 @@ end
...
-* `y`: y-component in meter
-* `z`: z-component in meter
-"""
-struct CartesianVector{T} <: AbstractCoordinateVector{T, Cartesian}
-    x::T
-    y::T
-    z::T
-end
+@inline _in_radius(p::CartesianPoint, radius::Real) = norm(p) <= radius
+@inline _in_radius(p::CartesianPoint, radius::AbstractInterval) = norm(p) in radius
```
lmh91, line 35: Naming is not clear. `radius` can be seen different for polar and spherical coordinates. I guess this is for the cylindrical case, and, thus, `norm(p)` does unnecessary calculations

lmh91: I know I defined `_in_r` but that just crossed my mind.

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -22,36 +22,30 @@ end
...
-* `φ`: Polar angle in radiance. φ == 0 <=> Parallel to x-axis of cartesian coordinate system."
+* `φ`: Polar angle in radians. φ == 0 <=> Parallel to x-axis of cartesian coordinate system."
 * `z`: z-coordinate in meter
 """
 struct CylindricalPoint{T} <: AbstractCoordinatePoint{T, Cylindrical}
     r::T
     φ::T
     z::T
-    CylindricalPoint{T}(r::T, φ::T, z::T) where {T} = new(r, mod(φ,2π), z)
+    CylindricalPoint{T}(r::T, φ::T, z::T) where {T} = new(r, mod(φ,T(2π)), z)
```
lmh91, line 48: Here, `T(2\pi)` is indeed necessary

src/ConstructiveSolidGeometry/PointsAndVectors.jl
lmh91: Ah, I see. `_in_r` is for cylindrical and `_in_radius` for spherical coordinates?
We should change this to `_in_r(p, z, ::Type{Cylindrical})` and add `abstract type Spherical`.

src/ConstructiveSolidGeometry/VolumePrimitives/HexagonalPrism.jl
```diff
@@ -0,0 +1,86 @@
...
+end
+HexagonalPrism(rInner, rOuter, zMin, zMax) = HexagonalPrism(; rInner, rOuter, zMin, zMax)
+
+function HexagonalPrism(rOuter::R, height::H) where {R<:Real, H<:Real}
+    T = promote_type(R,H)
+    HexagonalPrism( T, T(rOuter), T(height/2))
+end
+
+in(p::CartesianPoint, hp::HexagonalPrism{<:Any, <:Real, <:Any}) =
+    _in_z(p, hp.z) && _in_x(p, hp.r * sqrt(3)/2) && _in_y(p, hp.r - abs(p.x) /sqrt(3))
```
lmh91, line 27: `sqrt(T(3))/T(2)` is probably better. I am not sure if the compiler is clever enough. 
Did you testes what happens if `T = Float32`. Plain `sqrt(3)` should be Float64.

src/ConstructiveSolidGeometry/VolumePrimitives/HexagonalPrism.jl
```diff
@@ -0,0 +1,86 @@
...
+    return geom_round.(T[-sqrt(3)/2*_right(hp.r), -sqrt(3)/2*_left(hp.r), 0., sqrt(3)/2*_left(hp.r), sqrt(3)/2*_right(hp.r)])
+end
+
+# Y
+function get_important_points(hp::HexagonalPrism{T, <:Any, <:Any}, ::Val{:y})::Vector{T} where {T}
+    return geom_round.(T[-_right(hp.r), -_right(hp.r)/2, -_left(hp.r), -_left(hp.r)/2, 0., _left(hp.r)/2, _left(hp.r), _right(hp.r)/2, _right(hp.r)])
+end
+    
+
+function sample(hp::HexagonalPrism{T, <:Any, <:Any}, stepsize::Vector{<:Real}) where {T}
```
lmh91, line 74: Force us to give the right input types: `stepsize::Vector{T}`

src/ConstructiveSolidGeometry/VolumePrimitives/Sphere.jl
```diff
@@ -0,0 +1,23 @@
...
+    T = promote_type(typeof.((rMin, rMax))...)
+    r = rMin == 0 ? T(rMax) : T(rMin)..T(rMax)
+    Sphere(T, r)
+end
+
+Sphere(rMin, rMax) = Sphere(;rMin, rMax)
+
+function Sphere(r::T) where {T}
+    Sphere(T, r)
+end
```
lmh91, line 20: one line please

src/ConstructiveSolidGeometry/VolumePrimitives/Tube.jl
```diff
@@ -25,8 +26,9 @@ function Tube(radius::R, height::H) where {R,H}
     Tube( T, T(radius), nothing, T(height/2))
 end
 
-in(p::CartesianPoint, t::Tube{<:Any, <:Any, Nothing, <:Any}) = 
+in(p::AbstractCoordinatePoint, t::Tube{<:Any, <:Any, Nothing, <:Any}) =
```
lmh91, line 29: 👍

test/ConstructiveSolidGeometry/CSG_test.jl
```diff
@@ -42,73 +37,205 @@ end
...
                 rot_tube, 
-                rot_streched_tube, 
+                rot_stretched_tube, 
                 translated_tube, 
                 translated_rotated_tube ,
-                translated_rot_streched_tube,
+                translated_rot_stretched_tube,
             ]
         xs = T[]
         ys = T[]
```
lmh91, line 241: Can you please remove my plotting and example lines and put this into a clean test file?

lmh91 (changes requested):
Very very nice :)

Please have a look at my comments.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -0,0 +1,53 @@
...
+
+#Constructors
+function Cone(;rtopMin = 0, rtopMax = 1, rbotMin = 0, rbotMax = 1, φMin = 0, φMax = 2π, zMin = -1/2, zMax = 1/2)
+    T = promote_type(typeof.((rtopMin, rtopMax, rbotMin, rbotMax, φMin, φMax, zMin, zMax))...)
+    rtop, rbot = if rtopMin == rbotMin == 0 
+        T(rtopMax), T(rbotMax)
+    else
+        T(rtopMin)..T(rtopMax), T(rbotMin)..T(rbotMax)
+    end
+    φ = mod(φMax - φMin, 2π) == 0 ? nothing : T(φMin)..T(φMax)
```
fhagemann, line 25: This calculation of `φ` is also not good because `2π` is different for Float32 and Float64.

For instance, `mod(Float32(2π) - Float32(0), 2π) = 1.7484556025237907e-7` which is not zero.
Calling `Cone` with `φMin = Float32(0), φMax = Float32(2π)` will not result in `φ = nothing` like we would like it to.

We could write `mod(T(φMax) - T(φMin), T(2π))`
However, we do not force `T` to be a `Float`, so I guess it can also be an `Int`. This would give an `InexactError`.

Is it acceptable to force T to be a `Float`?

src/ConstructiveSolidGeometry/VolumePrimitives/Tube.jl
```diff
@@ -14,9 +14,10 @@ end
 # Constructors
 function Tube(;rMin = 0, rMax = 1, φMin = 0, φMax = 2π, zMin = -1/2, zMax = 1/2) 
     T = promote_type(typeof.((rMin, rMax, φMin, φMax, zMin, zMax))...)
+    r = rMin == 0 ? T(rMax) : T(rMin)..T(rMax)
     φ = mod(φMax - φMin, 2π) == 0 ? nothing : T(φMin)..T(φMax)
```
fhagemann, line 18: Same applies here for the `Tube`.

src/ConstructiveSolidGeometry/Intervals.jl
fhagemann: I only used `_left` for intervals that were not the cylindrical `r`.
(I can see that using `r` as variable name is confusing... 😅)

src/ConstructiveSolidGeometry/PointsAndVectors.jl
fhagemann: We could also go for the convention to use `ρ` for radii in cylindrical coordinates and `r` for radii in spherical coordinates.
(This is the general convention, see the Wikipedia entries for [cylindrical](https://en.wikipedia.org/wiki/Cylindrical_coordinate_system) and [spherical](https://en.wikipedia.org/wiki/Spherical_coordinate_system) coordinates).
However, `r` has been used throughout the whole SSD code and might cause confusions with the impurity charge density.

src/ConstructiveSolidGeometry/Intervals.jl
fhagemann: If `T` is an `Int32`, `2 * z` gives and `Int64` and `T(2) * z` gives an `Int32`.
For all `Float` it seems to make no difference.

src/ConstructiveSolidGeometry/VolumePrimitives/HexagonalPrism.jl
fhagemann: `sqrt(T(3))/2` is already enough

src/ConstructiveSolidGeometry/Intervals.jl
lmh91: I guessed so, but non the less the name `_left` for a function is really unclear. It should somehow give the reader of the code an idea what it will do. When passing an interval it is clear, but not when passing a number. The computer also can't tell whether the single number passed is an radial coordinate or an linear one. 
Here, I don't the see the point to make the name of the function as short as possible. þ

src/ConstructiveSolidGeometry/Intervals.jl
lmh91: Yes, because `promote_type(Int32, Int64) = Int64` and `promote_type(IntX, FloatY) = FloatY`. 
`f(x) = 2x; @code_native f(2.3)` gives me almost the same as `g(x::T) where {T} = T(2) * x; @code_native g(2.3)` 
Also almost the same for `@code_llvm`. As far as I understand the output, the compiler handles it automatically and we can use `2z` here (and in general in such situations). But I am not entirely sure.

src/ConstructiveSolidGeometry/PointsAndVectors.jl
lmh91: Hmm, I don't know how known this convention is. I would rather go for adding an additional argument. Or two functions with different naming `_in_cyl_r` / `_in_sph_r` or so.

src/ConstructiveSolidGeometry/VolumePrimitives/HexagonalPrism.jl
lmh91: Basically the same as above. Test it by putting it into a function (not just the REPL) and look at the produced code via `@code_llvm` and `@code_native` besides `@code_warntype`.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
lmh91: `mod(T(φMax) - T(φMin), T(2π))` 
I did not specified `<: AbstractFloat` (we should replace SSD just with AbstractFloat) so far as for now we could in general also use than something with units, e.g. `CartesianPoint(2.3u"m"...)`. (Cylindrical coordinates make this a bit tricky though. But certainly doable.). 
But using `Int` does not make really any sense in general so we should not care about that.

src/ConstructiveSolidGeometry/VolumePrimitives/Tube.jl
lmh91: so `== T(0)` ?

src/ConstructiveSolidGeometry/PointsAndVectors.jl
lmh91: Ah, and please use `hypot(p[1], p[2], p[3])` for the spherical radius

src/ConstructiveSolidGeometry/Intervals.jl
fhagemann: Okay, we will go for `_left_linear_interval`.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
fhagemann: I will add `float` to `promote_type(typeof.((rMin, rMax, φMin, φMax, zMin, zMax))...)` to assure that `T` is a Float.
We might need to change this if we implement unit support, but right now, promoting the type of `r` and `φ` would make no sense with units (one is in `m`, the other in `rad`)

fhagemann:
Before merging, I will update the `CSG_test.jl` file to also include sets and test transformations on all primitives.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
fhagemann: This is, for now, fixed by adding `float(...)` before `promote_type`.
We might need to rework this for units in the future, but then also think of a reasonable way of promoting spatial and angular coordinates differently.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -16,21 +16,21 @@ end
...
+    φ = mod(T(φMax) - T(φMin), T(2π)) == 0 ? nothing : T(φMin)..T(φMax)
     z = zMax == -zMin ? T(zMax) : T(zMin)..T(zMax)
     rtop == rbot ? Tube( T, rtop, φ, z) : Cone( T, rtop, rbot, φ, z)
 end
 Cone(rtopMin, rtopMax, rbotMin, rbotMax, φMin, φMax, zMin, zMax) = Cone(;rtopMin, rtopMax, rbotMin, rbotMax, φMin, φMax, zMin, zMax)
 
-function Cone(rtop::R1, rbot::R2, height::H) where {R1, R2, H}
-    T = promote_type(R1, R2, H)
-    Cone( T, T(rtop), T(rbot), nothing, T(height/2))
+function Cone(rtop::R1, rbot::R2, height::H) where {R1<:Real, R2<:Real, H<:Real}
```
lmh91, line 31: Lets restrict it to `AbstractFloat`. `Real` includes e.g. `Int`. 
And if we do restrict the type in general we should do it in the base construct of your primitives.
If we want to support units in the future we can still change/adapt to it. 
For now I would just add constructors which convert units into SI units and strip the units away.

lmh91 (changes requested):
Let restrict our CSG to `AbstractFloat` for now.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
fhagemann: Everywhere? Then also in the functions that are defined in the struct?

fhagemann:
In fact, `Tube` is now a special case of `Cone`. Now that everything is written so generic, we can adapt the `Cone` code in a way such that `Tube` is the case in which `rtop` is `rbot`.

src/ConstructiveSolidGeometry/Transformations.jl
```diff
@@ -1,20 +1,20 @@
...
     inv_s::SVector{3,T}
-    StretchedGeometry(p::P, s::SVector{3,T}) where {T,P} = new{T,P}(p, inv.(s))
+    ScaledGeometry(p::P, s::SVector{3,T}) where {T,P} = new{T,P}(p, inv.(s))
 end
-in(p::CartesianPoint, g::StretchedGeometry) = in(CartesianPoint(g.inv_s .* p), g.p)
-stretch(g::AbstractGeometry, s::SVector{3,T}) where {T} = StretchedGeometry(g, s)
-stretch(g::StretchedGeometry, s::SVector{3,T}) where {T} = StretchedGeometry(g.p, g.inv_s .* inv.(s))
+in(p::CartesianPoint, g::ScaledGeometry) = in(CartesianPoint(g.inv_s .* p), g.p)
+scale(g::AbstractGeometry, s::SVector{3,T}) where {T} = ScaledGeometry(g, s)
+scale(g::ScaledGeometry, s::SVector{3,T}) where {T} = ScaledGeometry(g.p, g.inv_s .* inv.(s))
```
lmh91, line 8: Shouldn't this be `ScaledGeometry(g.p, inv.(g.inv_s) .* s)`? :thinking: 
or better: `ScaledGeometry{T}(g.p, g.inv_s) .* inv.(s))`

lmh91 (changes requested):

src/ConstructiveSolidGeometry/Transformations.jl
fhagemann: Good catch! Calling `scale` on a `ScaledGeometry` returns a wrong `inv_s`!

src/ConstructiveSolidGeometry/Transformations.jl
```diff
@@ -14,7 +14,7 @@ struct RotatedGeometry{T,P<:AbstractGeometry{T},RT} <: AbstractGeometry{T}
 end
 in(p::CartesianPoint, g::RotatedGeometry) = in(g.inv_r * p, g.p)
 rotate(g::AbstractGeometry{T}, r::RotMatrix{3,RT,9}) where {T,RT} = RotatedGeometry(g, r)
-rotate(g::RotatedGeometry{T,<:Any,RT}, r::RotMatrix{3,RT,9}) where {T,RT} = RotatedGeometry{T,typeof(g),RT}(g.p, inv(r) * gp.r)
+rotate(g::RotatedGeometry{T,<:Any,RT}, r::RotMatrix{3,RT,9}) where {T,RT} = RotatedGeometry(g.p, r * inv(g.inv_r))
```
lmh91: I believe you just added one inversion. Now `rotate` will call `inv` two times.
Or is this beneficial rounding/numerical-precesion wise?
Same for `scale(g::ScaledGeometry, s::SVector{3,T})`.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -1,53 +1,79 @@
...
-    rbot::TR
+    r::TR #if not a Tuple, then Cone is a Tube
     φ::TP
     z::TZ
-    
+    #if r is a Tuple, the first entry refers to the r-interval at the bottom, the second one to the r-interval at the top
     function Cone( ::Type{T},
-                   rtop::Union{T, <:AbstractInterval{T}},
-                   rbot::Union{T, <:AbstractInterval{T}},
+                   r::Union{T, <:AbstractInterval{T}, Tuple{T,T}, Tuple{<:AbstractInterval{T},<:AbstractInterval{T}}},
```
lmh91, line 7: Is `Tuple{T, <:AbstractInterval{T}}` not also an option?
But make things probably just to complicated?

fhagemann: I excluded this option because if one `Interval` starts at `0` and the other one doesn't, we still have to determine the `r`-Interval for a given `z` anyhow. Then it is easier to stick to only deal with intervals.

src/ConstructiveSolidGeometry/Transformations.jl
fhagemann: This gave an error when I called it. I just realized this was due to `gp.r` and not `g.r` in the arguments.
I can change it back

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
lmh91: Alright, than just leave it like that.

src/ConstructiveSolidGeometry/Transformations.jl
lmh91: Hmm... I just recognized the following:
```julia
using Rotations, StaticArrays
m1 = RotMatrix(RotX(67) * RotY(45) * RotZ(13))
m2 = RotMatrix(RotX(-1.67) * RotY(5) * RotZ(.45))
m1i = inv(m1)
m2i = inv(m2)
m1i * m2i ≈ inv(m2 * m1) # true
```
So we have to change the order to : 
```julia
rotate(g::RotatedGeometry{T,<:Any,RT}, r::RotMatrix{3,RT,9}) where {T,RT} = RotatedGeometry{T,typeof(g),RT}(g.p,gp.r * inv(r))`
```
Am i right?

However, there are some rounding issues
```julia
m1i * m2i == inv(m2 * m1) # false
```
So I would go for actually inverting it back to `m1` times as I believe it is numerically better. 
But, I am not sure here...

fhagemann: Apparently, calling `RotatedGeometry{...}` and `ScaledGeometry{...}` is not possible and results in errors.
I added another function into the struct that will handle repeated transformations accordingly.
The results are tested using the repeated function calls in `CSG_test.jl`
Note the order of the matrices in `RotatedGeometry` that is correct now.

lmh91: Because `inv(m1i) == m1` and `inv(m2i) == m2`

fhagemann: I just reverted my commit, now it is inverting it back.

lmh91 (approved):
Very nice work Felix! 👍

Issue #110, opened 2020-11-25
Refactor CSG


lmh91:
I started refactoring our CSG. The core is finished. See branch [refactorCSG](https://github.com/JuliaPhysics/SolidStateDetectors.jl/tree/refactorCSG).

This also concerns #62, #63, #84

The CSG is separated in a new submodule `ConstructiveSolidGeometry` inside SSD.

**Rotation**, **translation** and **stretching/scaling** are now implemented. 

For now, I only added one primitive, the `Tube`.

I added a test script under `/test/ConstructiveSolidGeometry/CSG_test.jl`. 
At the moment is more an example script. 

We can start implementing our other primitives and functionality, like read-in, to it or adept to it. 
But when adding something, please **test for performance (type stability)** like it is done in the test script. 
A PR to this branch must have tests.

## Guideline for Primitives
* Our primitives will be parametric types so that we can save computation time for special version of the primitives. 
E.g., in case of a full tube (cylinder) we don't have to check for `rMin` and the polar coordinate.

* Default constructors: All primitives should be constructed around `(0,0,0)`. This makes rotations easier for the user. This is also the way how it is done in `GEANT-4`. E.g., `Tube(radius = 1.0, height = 1.0)` will create a `Tube` with an radius of 1 and going from -0.5 to +0.5 in `z`. Here, we will have to change (and break) our read-in our config files.

## General remarks

* Branch `refactorCSG`: The old CSG is still inside the package. The new one is only included but not used yet. 
`CartesianPoint` and `CylindricalPoint` are not exported anymore to avoid conflicts as there exists now two version (for now). E.g.: `SolidStateDetectors.CartesianPoint` and `SolidStateDetectors.ConstructiveSolidGeometry.CartesianPoint`.

* The conversion of a `CartesianPoint` into a `CylindricalPoint` got changed. 
Previously: `r = sqrt(pt.x * pt.x + pt.y * pt.y`
New: `r = hypot(pt.x, pt.y)` (special algorithm, see Julia documentation of `hypot`)

## ToDo:
**EDIT May 15th, 2021**
The primitives have all been updated. Find a list of what still needs to be done at the end of this issue.

fhagemann:
When calling a `Tube`, it will be stored as a `Cone` now (the struct `Tube` is removed with #115)

fhagemann:
[...]

PR #118, opened 2020-12-03
Plot recipes for new CSG

fhagemann:
I added plot recipes for the wireframe plotting method for the new CSG.

For this a new struct `LineSegments` was added that stores an `Array` of `CartesianPoint`. 
When plotting `LineSegments`, all `CartesianPoint` in the array are plotted as continuous line consisting of line segments.

Each primitive should be provided with a `get_plot_points` functions which gives an `Array` of `LineSegments`.
These points should be in the basic coordinate system of the primitive.

For all transformations, `get_plot_points` was added that transforms the `LineSegments` of the untransformed geometry accordingly.

Find some results here:

Tube
Box
Sphere
HexagonalPrism

I am not sure about where to save the code (all plotting code saved together or `get_plot_points` in the respective folder?)
and I would appreciate feedback (also about the naming, implementation, etc.)!


src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -4,9 +4,9 @@ struct Cone{T,TR,TP,TZ} <: AbstractVolumePrimitive{T}
     z::TZ
     #if r is a Tuple, the first entry refers to the r-interval at the bottom, the second one to the r-interval at the top
     function Cone( ::Type{T},
-                   r::Union{T, <:AbstractInterval{T}, Tuple{T,T}, Tuple{<:AbstractInterval{T},<:AbstractInterval{T}}},
+                   r::Union{T, <:AbstractInterval{T}, Tuple{T,T}, Tuple{I,I}},
                    φ::Union{Nothing, <:AbstractInterval{T}},
-                   z::Union{T, <:AbstractInterval{T}}) where {T}
+                   z::Union{T, <:AbstractInterval{T}}) where {T, I<:AbstractInterval{T}}
```
lmh91, line 9: maybe
```julia
r::Union{T, I, Tuple{T,T}, Tuple{I,I}},
φ::Union{Nothing, I}
z::Union{T, I}) where {T, I<:AbstractInterval{T}}
```
?

src/ConstructiveSolidGeometry/LinePrimitives/LinePrimitives.jl
```diff
@@ -0,0 +1,3 @@
+struct LineSegments{T} <: AbstractLinePrimitive{T}
+    points::Vector{CartesianPoint{T}}
+end
```
lmh91, line 3: Do we really need this?

Shouldn't we rather implement broadcasting methods for `Vector{CartesianPoint{T}}`? :thinking:

src/ConstructiveSolidGeometry/LinePrimitives/LinePrimitives.jl
```diff
@@ -1,3 +1,16 @@
...
+(*)(s::SVector{3,T}, pl::LineSegments{T}) where {T} = pl * s
+
+function (+)(pl::LineSegments{T}, t::CartesianVector{T})::LineSegments{T} where {T}
+    LineSegments(map(p -> p + t, pl.points))
+end
+
+function (*)(r::RotMatrix{3,T,TT}, pl::LineSegments{T})::LineSegments{T} where {T, TT}
+    LineSegments(map(p -> r * p, pl.points))
+end
```
lmh91, line 16: We should implement broadcasting (`.+`-syntax ) methods for our Point and Vector types. 
https://docs.julialang.org/en/v1/manual/arrays/#Broadcasting

lmh91 (commented):
Nice work and plots :)

Please have a look at my comments.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
fhagemann: Are we sure that all intervals are of the same type `(:closed,:closed)`?
This might not allow using `(:closed,:open)` for `φ` for example (not that we are using this right now, but who knows)

src/ConstructiveSolidGeometry/LinePrimitives/LinePrimitives.jl
fhagemann: Good point! Let me check.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
lmh91: ah, year.. You are right.

src/ConstructiveSolidGeometry/LinePrimitives/LinePrimitives.jl
fhagemann: I did not come up with a fast way of implementing the broadcasting methods.

Broadcasting two arrays of different sizes (where one is treated as scalar) does not seem to be that simple as the broadcast method results in an error if the arrays are of different length.

One way to avoid treating a translation vector as `Array` of length 3 would be to pass it as an 1-dimensional `Tuple`:
```julia
@inline (+)(vp::Vector{CartesianPoint{T}}, v::CartesianVector{T}) where {T} = vp .+ (v,)
@inline (+)(vvp::Vector{Vector{CartesianPoint{T}}}, v::CartesianVector{T}) where {T} = vvp .+ (v,)  
vvp + v
```
However, this results in allocations which we do not want.

A loop gives allocation free result if the input vector is overwritten:
```julia
function add!(vvp::Vector{Vector{CartesianPoint{T}}}, v::CartesianVector{T}) where {T}
    for i in eachindex(vvp)
        vp::Vector{CartesianPoint{T}} = vvp[i]
        for j in eachindex(vp)
            vp[j] += v
        end
    end
    #vvp
end
add!(vvp, v)
```

@oschulz do you have a workaround for broadcasting differently sized arrays? 
Something in ArraysOfArrays.jl or UnsafeArrays.jl?

lmh91: Hmm, I also looked a bit into broadcasting but also did not find a good implementation. 

As a work around we can for sure define methods like `translate!`, `rotate!` and `scale!` for `Vector{CartesianPoint}`. 

Personally, I would like to have the `.+`-syntax as well. I would really appreciate input from @oschulz here as I am not sure whether it might not be a wrong usage of that syntax.

oschulz: I don't think we should wer our Point and Vector types as scalars under broadcasting. They are vectors, semantically, and so the default behavior for broadcast should be to broadcast over them. In applications where we do want to treat them as scalars (from a broadcast perspective), we'll just wrap them in a `Ref` or `Tuple`.

oschulz: So the Julianic syntax would be `vp .+ Ref(v)` or `vp .+ (v,)`.

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -21,7 +21,7 @@ end
 
 @inline (*)(p::CartesianPoint{T}, v::SVector{3,T}) where {T} = p .* v
 @inline (*)(vp::Vector{CartesianPoint{T}}, v::SVector{3,T}) where {T} = vp .* (v,)
-@inline (*)(vp::Vector{Vector{CartesianPoint{T}}}, v::SVector{3,T}) where {T} = vp .* (v,)
+@inline (*)(vvp::Vector{Vector{CartesianPoint{T}}}, v::SVector{3,T}) where {T} = vvp .* (v,)
```
lmh91: This, is actually a dangerous notation. Here, I would drop the `*`-method and would call `scale!` or `scale`

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -96,8 +96,6 @@ end
 
 @inline (+)(vp::Vector{CartesianPoint{T}}, v::CartesianVector{T}) where {T} = vp .+ (v,)
 @inline (+)(vvp::Vector{Vector{CartesianPoint{T}}}, v::CartesianVector{T}) where {T} = vvp .+ (v,)
```
lmh91: Is this allocation free?

lmh91 (commented):
So regarding the short `(+)`, `(*)`,... methods for different types (broadcasting discussion):
I would actually drop them for now and only use `rotate!`, `rotate`, `translate!`, ... . 
Then, it is clear what the function will do. Otherwise, it might be unclear in some cases. 
We can later still add them. But for now, I would drop them. 

For the sets, like `GeometryUnion`,.. (&), ... is fine.

src/ConstructiveSolidGeometry/PointsAndVectors.jl
fhagemann: No, this gives one allocation.

fhagemann:
I replaced the Base operations for `CartesianPoint` by `scale(!)`, `rotate(!)` and `translate(!)` for clarity.
Writing the functions as loops instead of broadcasting allows for 0 allocations.
The methods `(*)` and `(+)` to construct transformed geometries are commented out but left in the code.

src/ConstructiveSolidGeometry/Transformations.jl
```diff
@@ -6,8 +6,12 @@ end
...
 scale(g::AbstractGeometry{T}, s::SVector{3,T}) where {T} = ScaledGeometry(g, s)
 scale(g::ScaledGeometry{T}, s::SVector{3,T}) where {T} = ScaledGeometry(g.p, inv.(g.inv_s) .* s)
-(*)(g::AbstractGeometry{T}, s::SVector{3,T}) where {T} = scale(g, s)
-get_plot_points(sg::ScaledGeometry{T}) where {T} = get_plot_points(sg.p) * inv.(sg.inv_s)
+#(*)(g::AbstractGeometry{T}, s::SVector{3,T}) where {T} = scale(g, s)
+function get_plot_points(sg::ScaledGeometry{T}) where {T}
+    pp = get_plot_points(sg.p)
+    scale!(pp, inv.(sg.inv_s))
+    pp
+end
```
lmh91, line 14: I would define the `translate!`, `scale!` and `rotate!` methods such, that they return the vector again. 
Then, you can write here:
```julia
get_plot_points(sg::ScaledGeometry) = scale!(get_plot_points(sg.p), inv.(sg.inv_s))
```
This might also come in handy in other situations.

lmh91 (changes requested):
Almost there :)

src/ConstructiveSolidGeometry/PointsAndVectors.jl
fhagemann: By the way, `translate`, `rotate`, `translate!` and `rotate!` are also methods that are exported by Plots.jl.
When explicitly importing these functions from SSD, it gives a warning in either SSD or Plots.
For this, using `(+)` for translations and `(*)` for rotations would come in quite handy.

lmh91 (approved):
Nice 👍

lmh91:
I will update #110

Issue #84, opened 2020-10-30
CSG: Change internal Primitive Definition: Centered around 0


lmh91:
We should change our definitions of all primitives to be centered around (0,0,0).
Like in GEANT4. 

This also simplifies the rotation of objects (see #62). 

It also comes in handy when we want to add some clustering of objects onto subregions of a grid
in order to parallelize and speed up the initialization of the boundary conditions on the grid for the SOR
which can take some time if there are to many objects and the grid becomes fine. 
This clustering might also be used to speed up the charge drift in general (lesser lookups) (see #85 ).

lmh91:
I added this issues to the v1.0.0 milestone as this might impact the definition of primitives in the config files. 
Although I am not sure about that.

fhagemann:
I was just thinking: the new CSG will probably only affect `Tube` and `Cone` when it comes to reading in config files.
However, the `z`-interval is defined with a field `"h"` right now. 

We could think about getting rid of `"h"` in general and only stick to `"z"` (either pass an interval with `"from"` and `"to"`, or just a number if centered around zero) and provide a conversion function that converts old CSG config files into new ones for the users to update.

lmh91:
I believe you can already do use "from" and "to". We had this at the very beginning. @oschulz wanted the `h`-syntax to be possible... :stuck_out_tongue_closed_eyes: 

We could add such a conversion function. User would probably appreciate it. There we could also introduce the change regarding the impurity-charge-density naming change.

oschulz:
`z` may be misleading especially when the object is centered at zero - is it the full height, or just from center to top. I think `h` is clearer.

fhagemann:
As we might need to change the config file layout anyhow, I worked on some functions to convert geometries to a dictionary that can be saved as config file afterwards. (Still experimental and trying to take into account a [comment from @lmh91](https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/122#issuecomment-739755931)).

```julia
using SolidStateDetectors
const SSD = SolidStateDetectors
const CSG = SSD.ConstructiveSolidGeometry
using Rotations
using StaticArrays
using IntervalSets
using DataStructures: OrderedDict

function construct_dict_from_geometry(g::CSG.AbstractGeometry{T}) where {T}
    d = OrderedDict{Any,Any}()
    add_entry_to_config_dict!(d, g)
    d
end

[...]

PR #122, opened 2020-12-06
Constructors for config file for new CSG

fhagemann:
This pull request implements the construction of CSG objects from config files.
It also allows decomposing volumes and plotting the results.

## Construct CSG geometries from config files

To read in existing config files, I added the functions `Geometry` for all primitives, the `TranslatedGeometry` and all sets.
This allows parsing dictionaries of type `"geometry"` from the config files. As an example,
```julia
using SolidStateDetectors
const SSD = SolidStateDetectors
const CSG = SolidStateDetectors.ConstructiveSolidGeometry
T = Float32

filename = SSD_examples[:InvertedCoax]
dict = SSD.parse_config_file(filename)
input_unit_dict = SSD.construct_units(dict)
g = CSG.Geometry(T, dict["objects"][3]["geometry"], input_unit_dict)

```
will result in an object `g` of type 

```CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Float32,Nothing,Float32}}}(CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}}(CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}}(CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}}(CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.015..0.035, nothing, 0.0f0), Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.035..0.035, nothing, 0.01f0)), TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}(Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}((0.035..0.035, 0.02442..0.02442), nothing, 0.03f0), Float32[0.0, 0.0, 0.02])), TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.005..0.02442, nothing, 0.0f0), Float32[0.0, 0.0, 0.08])), TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.005..0.005, nothing, 0.0275f0), Float32[0.0, 0.0, 0.025])), TranslatedGeometry{Float32,Cone{Float32,Float32,Nothing,Float32}}(Cone{Float32,Float32,Nothing,Float32}(0.005f0, nothing, 0.0f0), Float32[0.0, 0.0, 0.025]))```

We might need to think about reducing what to print when showing the type, as this can get quite long...

## Decomposition of CSG geometries

The object `g` is an `AbstractGeometry{T}` that can be decomposed into positive and negative geometries using
```julia 
positive_geometry, negative_geometry = CSG.get_decomposed_volumes(g) 
```

## Plotting of composed CSG geometries

Decomposing the volumes allows plotting the positive geometries via the updated plot recipe for `AbstractGeometry`,  i.e. `plot(g)`
plotg

## Primitive defined with centre around 0 (#84)

Note that now we run into the issue that the `Tube` and `Cone` primitives with heights defined via `"h"` in the config file are centred around zero, i.e. go from `-h/2` to `h/2` instead of going from `0` to `h`. This is something that we cannot avoid.

To convert old config files to new ones, I have written a small script, that adds `h/2` to the `translate` vector for every `Tube` and `Cone` with a height defined via `h`. It is not yet in the source code because I wanted to discuss where to put it.

```julia
using DataStructures, JSON
function update_config_file(filename::String)
    dict = JSON.parsefile(filename, dicttype = DataStructures.OrderedDict)
    update_dict!(dict)
    open(replace(filename, ".json" => "-new.json"),"w") do f
        JSON.print(f, dict, 4)
    end
end 

function update_dict!(dict::Union{Dict, OrderedDict})
    dict_keys = keys(dict)
    # translate all Tubes and Cones that are defined via "h" and where h is not zero:
    if "type" in dict_keys && dict["type"] in ["tube", "cone"] && "h" in dict_keys && dict["h"] != 0
        # create translate vector with z = 0 if it does not exist yet:
        if !("translate" in dict_keys) dict["translate"] = Dict{String,Any}("z" => 0) end
        # add h/2 to the z-component of the translate vector:
        dict["translate"]["z"] = CSG.geom_round(dict["translate"]["z"] + Float64(dict["h"] / 2))
        # if the resulting z-component is zero, remove it:
        if dict["translate"]["z"] == 0 pop!(dict["translate"]["z"]) end
        # if the whole translate vector happens to be 0, remove it:
        if prod(values(dict["translate"]) .== 0) pop!(dict["translate"]) end
    end

    # update nested dictionaries recursively:
    for k in dict_keys
        if typeof(dict[k]) <: OrderedDict
            dict[k] = update_dict!(dict[k])
        elseif typeof(dict[k]) <: Array
            for i in eachindex(dict[k])
                if typeof(dict[k][i]) <: OrderedDict
                    dict[k][i] = update_dict!(dict[k][i])
                end
            end
        end
    end
    dict
end
```

To update, call the function with the config file name, e.g. 
```julia
update_config_file(SSD_examples[:InvertedCoax])
```

This results in a new config file with the right values for constructing `g`:

```CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Float32,Nothing,Float32}}}(CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}}(CSGUnion{Float32,CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}}(CSGUnion{Float32,CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}},TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}}(CSGUnion{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32},TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.015..0.035, nothing, 0.0f0), TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.035..0.035, nothing, 0.01f0), Float32[0.0, 0.0, 0.01])), TranslatedGeometry{Float32,Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}}(Cone{Float32,Tuple{Interval{:closed,:closed,Float32},Interval{:closed,:closed,Float32}},Nothing,Float32}((0.035..0.035, 0.02442..0.02442), nothing, 0.03f0), Float32[0.0, 0.0, 0.05])), TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.005..0.02442, nothing, 0.0f0), Float32[0.0, 0.0, 0.08])), TranslatedGeometry{Float32,Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}}(Cone{Float32,Interval{:closed,:closed,Float32},Nothing,Float32}(0.005..0.005, nothing, 0.0275f0), Float32[0.0, 0.0, 0.0525])), TranslatedGeometry{Float32,Cone{Float32,Float32,Nothing,Float32}}(Cone{Float32,Float32,Nothing,Float32}(0.005f0, nothing, 0.0f0), Float32[0.0, 0.0, 0.025]))```

newplotg

For the conversion script, we might want to add a new entry (something like `"version" : 1.0`) to the updated config files and throw a warning for all config files that do not have that field that they need to be updated to work with the new CSG. If people update the config files themselves, they can add this entry to avoid the warning or it created automatically when running ```update_config_file```.

### CODE REVIEW

**Review by lmh91 — CHANGES_REQUESTED** [review]
Nice 👍  You are too fast :) I wanted to tell you some general changes for this. 
But you should be able to easily change it. 

* We should change `T::DataType` to `::Type{T}`
* Instead of `Val{:tube}` we should use `::Type{Cone{...}}`, ...
* Instead of `Val{:union}` we should use `::Type{CSGUnion}`, ...
* We should define our unit dict not as `Dict` but as a `NamedTuple`

**Comment by lmh91** [comment]
Regarding the long type of the CSG. 
This might become a problem performance wise if the type gets to large. 
For now I would let it like this. Otherwise we might have to life with some allocations. 

But for the pretty printing have a look here: https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing

File: src/ConstructiveSolidGeometry/Decompose.jl [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+function decompose_volume(vol::TranslatedGeometry, pos, neg, translate, flag = 1)::Nothing
+    translate += vol.t
+    decompose_volume(vol.p, pos, neg, translate, flag)
+    nothing
+end
+
+function decompose_volume(vol::AbstractGeometry, pos, neg, translate, flag = 1)::Nothing
+    flag == 1 ? push!(pos, vol + translate) : push!(neg, vol + translate)
+    nothing
+end
```
lmh91, line 36: Gereral Julia Styleguide: Should be
```julia
function decompose_volume!(pos, neg, vol::CSGIntersection, translate, flag = 1)::Nothing
```
as `pos` and `neg` are changed in the function, thus `!` and they are first. 
`RotatedGeometry` and `ScaledGeometry` will be added right?

For the `flag` I would use `Val{:PositiveVolume}` and `Val{:NegativeVolume}` 
and write to one-line methods like:
```julia
decompose_volume!(vol::AbstractGeometry, pos, neg, translate, ::Type{Val{:PositiveVolume}}) = push!(pos, vol+translate)
```

src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
```diff
@@ -11,6 +11,7 @@ module ConstructiveSolidGeometry
     using RecipesBase
     using Rotations
     using StaticArrays
+    using Unitful
```
lmh91, line 14: I am not sure if we want this here. Sure, In general, general unit support would be really nice inside the CSG but it might makes everything more complicated. 

I have to think about this.

lmh91 (changes requested):
I added some comments :)

lmh91:
Could you add example `yaml`-config files for the individual primitives? 
We can use them for documentation and testing. 
E.g.:
```YAML
type: "tube"
r: 
  from: 0.1
  to: 0.4
phi: # optional, [0,2Pi] is default interval
  from: 0
  to: 6.283185307179586 
h: 0.8 # -> zMin = -0.4 - zMax = 0.4
```
And a load function should return (in that case) just a tube (Cone). 


We should think about adding more possible ways of input, like
```YAML
r: [0.1, 0.4]
```
But such things can be added later.

lmh91:
Okay, I thought about units.

In Unitful.jl there is the function `uparse`. It converts a string like `"1.2m"` into `1.2u"m"`.
I would really like to add this to our read in. Not only regarding the CSG, but also everything else
like temperature and impurity densities. 
This also allows for `phi: [0°, 180°]`

If the value of a variable in the config file is just a number the default unit should be used. In our case SI units.
However, we still allow for setting specifying the default units of the config file as we do already in the unit part. 
But, now we could also overwrite them locally by specifying units. 

In the CSG Submodule we only need 
```julia
const internal_unit_length  = u"m"
const internal_unit_angle   = u"rad"
```
(I would change the order such that via tab-autocomplete one would get with internal_u<\tab>` all possible units.)

fhagemann:
> Could you add example `yaml`-config files for the individual primitives?
> We can use them for documentation and testing.
> E.g.:
> 
> ```yaml
> type: "tube"
> r: 
>   from: 0.1
>   to: 0.4
> phi: # optional, [0,2Pi] is default interval
>   from: 0
>   to: 6.283185307179586 
> h: 0.8 # -> zMin = -0.4 - zMax = 0.4
> ```
> 
> And a load function should return (in that case) just a tube (Cone).

We should think of a better way of implementing that `φ` is full `2π`.
Right now, running your code with `Float32` as type precision results in a `Cone` with `φ = 0.0..6.28319` instead of `φ = nothing`.

lmh91:
Well, not defining phi in the config file would be default for `full`. `360°` should also work. `2π` might also work. (Did not tested it.)
The config snipped was only an dummy example.

fhagemann:
This is the function to read in the YAML config files I added:
```julia
using SolidStateDetectors.ConstructiveSolidGeometry: Geometry
function Geometry(::Type{T}, filename::String, input_units::NamedTuple = (length = internal_unit_length, angle = internal_unit_angle)) where {T}
    @assert isfile(filename) "The given filename '$(filename)' does not lead to a valid file."
    dict = if endswith(filename, ".json")
        JSON.parsefile(filename)
    elseif endswith(filename, ".yaml")
        YAML.load_file(filename)
    else
        @error "Only JSON and YAML formats are supported at the moment."
    end
    Geometry(T, dict, input_units)
end
```

I can add it into the source code if we decide on where. If we put it into the `ConstructiveSolidGeometry` module, we need to add `YAML` and `JSON` as dependency.

src/ConstructiveSolidGeometry/IO.jl
```diff
@@ -1,98 +1,111 @@
...
+#parses dictionary entry for heights to respective AbstractFloat/Interval (divides by 2 if it is a single value)
+@inline _parse_interval_height(::Type{T}, x::Union{Real, String}, unit::Unitful.Units) where {T} = _parse_value(T, x, unit)/2
+@inline _parse_interval_height(::Type{T}, dict::Union{Dict{String,Any}, Dict{Any,Any}}, unit::Unitful.Units) where {T} = _parse_interval(T, dict, unit)
+
+
+# parses dictionary entry for φ-interval that has {"from" ..., "to": ...} to the respective Nothing/Interval
+function _parse_interval_φ(::Type{T}, dict::Union{Dict{String,Any}, Dict{Any,Any}}, unit::Unitful.Units) where {T}
+    φTo::T = _parse_value(T, dict["to"], unit)
+    φFrom::T = _parse_value(T, dict["from"], unit)
+    if abs(rem2pi(φFrom - φTo, RoundNearest)) > 2*eps(T)
```
lmh91, line 42: I would just return the interval.

I have the feeling we can't avoid adjusting the `in` method for `phi`-intervals.

lmh91:
I pushed one commit regarding the test file onto your branch. 
It does not fixes all failed tests with Float32, but most of them.

Can you make the one change you had in mind? Then, I would merge.

src/ConstructiveSolidGeometry/IO.jl
fhagemann: I don't understand your comment regarding this specific part of the code.
The problem here was that we need a reliable method to return `φ = nothing` if `2π ` is a full `2π` interval that will skip the `_in_φ` query when calling the `in` method. It is more about how to increase the absolute tolerance which is now set to `2*eps(T)`.

fhagemann:
> I pushed one commit regarding the test file onto your branch.
> It does not fixes all failed tests with Float32, but most of them.

I implemented that the tests run for `Float64` and `Float32`. 
With `Float32`, there seems to be rounding errors when it comes to multiple translations. This should not be.

fhagemann:
The problems seen in the test script seem to be avoided when creating the rotation matrices first in `Float64` and converting them to `Float32` only when applying them to the `AbstractGeometry` to form a `RotatedGeometry`, i.e.
```julia
T = Float32
tube = Tube(T(0), T(0.5), T(0), T(0), T(0), T(1)) # r from 0.0..0.5, z from 0.0..1.0
rotX = RotX{Float64}(deg2rad(90))
rot_tube_x = RotMatrix{3,T}(rotX) * tube
CartesianPoint{T}(0.3, -0.5, 0.4) in rot_tube_x # = true (how it should be)
```
rather than
```julia
T = Float32
tube = Tube(T(0), T(0.5), T(0), T(0), T(0), T(1)) # r from 0.0..0.5, z from 0.0..1.0
rotX = RotX{T}(deg2rad(90))
rot_tube_x = RotMatrix(rotX) * tube
CartesianPoint{T}(0.3, -0.5, 0.4) in rot_tube_x # = false (not how it should be)
```

Adjusting the test script file to the example shown above will result in no failed tests for the primitives tested in the test script.
However, we need to see how we can avoid such problems happening for arbitrary volumes given by the user.

src/ConstructiveSolidGeometry/IO.jl
fhagemann: This is now solved by not calling `geom_round` in `_parse_value`.

lmh91:
In out read in methods for rotations we can use `Float64` for the rotation. 
This fix the issue for config files. If one does wants to create the geomtry in a Julia script, the user has to take care of this. 
We can add this information to the documentation of the `RotatedGeometry`. 

Otherwise, we would need to add a new function for us which would basically only be a wrapper, e.g. `SSDRotX(deg2rad(90))`,
which uses `Float64`.

What happens if you just use `RotX(deg2rad(90))`?

lmh91:
If you add decomposition, please add it with `Rotation`, `Translation` and `Scaling`.

lmh91:
If you run into trouble with multiple dispatch (ambiguous function calls) we might think about to add some abstract layer types 
like `AbstractTransformedVolumePrimitive` and/or `AbstractTransformedCSG`. 

And/Or, it might also be useful to add a function like `is_transformed_volume_primitive(g::AbstractGeometry)` to check whether `g` is only a Primitive which is only translated/scaled/rotated (and arbitrary combinations). 
(The decomposition does not need to be that fast as it is usually done only once.)

fhagemann:
> What happens if you just use `RotX(deg2rad(90))`?

If you use anything in the argument without specifying the type, it will return `RotX{Float64}` (except for if you specify the argument as float which is not a `Float64`).

lmh91:
> > What happens if you just use `RotX(deg2rad(90))`?
> 
> If you use anything in the argument without specifying the type, it will return `RotX{Float64}` (except for if you specify the argument as float which is not a `Float64`).

So I guess we are just fine, right? We just use `RotX` without type specifier. 
It should be somewhere documented and noted though. 

`RotX(deg2rad(90f0))` probably does not work, right?

fhagemann:
> > > What happens if you just use `RotX(deg2rad(90))`?
> > 
> > 
> > If you use anything in the argument without specifying the type, it will return `RotX{Float64}` (except for if you specify the argument as float which is not a `Float64`).
> 
> So I guess we are just fine, right? We just use `RotX` without type specifier.
> It should be somewhere documented and noted though.
> 
> `RotX(deg2rad(90f0))` probably does not work, right?

Well, it gives a rotation matrix of type `RotX{Float32}` that might again result in rounding issues.

fhagemann:
I implemented the decomposition now also for rotated and scaled geometries.
It should be ready to merge now.

lmh91 (approved):

fhagemann:
I updated #110.

lmh91:
Thanks :+1:  I also added a new point to it (Integration into SSD).

PR #127, opened 2020-12-19
Surfaceprimitives

hervasa2:
Added Surface Primitive Structs
`cone = CSG.Cone(4,7,3,5,0,3π/2,0,5); surf = CSG.get_decomposed_surfaces(cone)`
Decomposes the volume primitive into surface primitives. This is the result:
`6-element Array{SolidStateDetectors.ConstructiveSolidGeometry.AbstractSurfacePrimitive,1}:
SolidStateDetectors.ConstructiveSolidGeometry.CylindricalAnnulus{Float64,Interval{:closed,:closed,Float64},Interval{:closed,:closed,Float64},Float64}(4.0..7.0, 0.0..4.71238898038469, 0.0)
SolidStateDetectors.ConstructiveSolidGeometry.CylindricalAnnulus{Float64,Interval{:closed,:closed,Float64},Interval{:closed,:closed,Float64},Float64}(3.0..5.0, 0.0..4.71238898038469, 5.0)
SolidStateDetectors.ConstructiveSolidGeometry.ConeMantle{Float64,Tuple{Float64,Float64},Interval{:closed,:closed,Float64},Interval{:closed,:closed,Float64}}((7.0, 5.0), 0.0..4.71238898038469, 0.0..5.0)
SolidStateDetectors.ConstructiveSolidGeometry.ConalPlane{Float64,Tuple{Interval{:closed,:closed,Float64},Interval{:closed,:closed,Float64}},Float64,Interval{:closed,:closed,Float64}}((4.0..7.0, 3.0..5.0), 0.0, 0.0..5.0)
SolidStateDetectors.ConstructiveSolidGeometry.ConalPlane{Float64,Tuple{Interval{:closed,:closed,Float64},Interval{:closed,:closed,Float64}},Float64,Interval{:closed,:closed,Float64}}((4.0..7.0, 3.0..5.0), 4.71238898038469, 0.0..5.0)
SolidStateDetectors.ConstructiveSolidGeometry.ConeMantle{Float64,Tuple{Float64,Float64},Interval{:closed,:closed,Float64},Interval{:closed,:closed,Float64}}((4.0, 3.0), 0.0..4.71238898038469, 0.0..5.0)`

When `plot(volume_primitive)` is called, the volume primitive is decomposed and each surface is plotted. For `SSD_style = :wireframe` only unique lines are plotted and for `SSD_style = :surface` all meshes of the decomposed surfaces are plotted (meshes unlike the wireframe lines are already unique as given by `get_decomposed_surfaces`) . 

`plot(cone, SSD_style = :surface, c = RGBA(0, 102/255, 1, 0.4)); plot!(cone, c = :black)`

cone_surf_wireframe

Each surface primitive must have a `mesh(surf_primitive)` and a `get_plot_points(surf_primitive)` function for  `SSD_style = :surface` and `SSD_style = :wireframe` respectively. `get_plot_meshes(volume_primitive)` and `get_plot_points(volume_primitive)` call these respectively.  Meshes are a structs with x, y, and z fields which are matrices. 
Note that `SSD_style = :surface` does not work in gr(). For this I have a function that decomposes a mesh into triangles which GR will understand, but it is currently not in. If we always deconstruct into triangles it will work in gr() and pyplot() but plotting time increases drastically. I have added the Colors dependency so that we can always plot with transparency. The problem is that 3D plots in pyplot() ignore `alpha` (I've posted about this issue on Plots.jl) so a default RGBA value needs to be declared in code. This is still not done. 

`scatter(CSG.sample_surface(cone,100u"mm"), markerstrokewidth = 0, alpha = 0.1)`

cone_sample_surface

Sample currently only accepts one unitful step input. It will sample each dimension evenly with this step. φ  is now sampled by arc length. 

`plot(surf[1], c = :black); scatter!(CSG.sample(surf[1],1000u"mm"), markerstrokewidth = 0, alpha = 1)`

annulus

Each surface primitive already has or will have the following functions: `in(),  sample(), distance_to_surface()` among others. The simplest way to construct a surface primitive is to feed it the volume primitive it belongs to and a "slice" value like `CSG.CylindricalAnnulus(cone, z = 0)` or `CSG.ConalPlane(cone, φ = 3π/2)`. Other constructors are also available. 

`get_decomposed_surfaces()` is already written for cone, but will need to be added to all volume primitives.


lmh91:
I separated plotting code from actual source code, see #128. 
It's already merged into `refactorCSG`.
Please adapt your changes to it. So everything which is only needed for plotting goes to one of the plotting directories. 
The sample method for the primitives should go to the primitives as we also need them for the field calculations.

lmh91 (changes requested):
I separated plotting code from actual source code, see #128.
It's already merged into refactorCSG.
Please adapt your changes to it. So everything which is only needed for plotting goes to one of the plotting directories.
The sample method for the primitives should go to the primitives as we also need them for the field calculations.

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -0,0 +1,48 @@
...
+
+#Constructors
+function Torus(;r_torus = 1, r_tubeMin = 0, r_tubeMax = 1, φMin = 0, φMax = 2π, θMin = 0, θMax = 2π)
+    T = float(promote_type(typeof.((r_torus, r_tubeMin, r_tubeMax, φMin, φMax, θMin, θMax))...))
+    r_tube = r_tubeMin == 0 ? T(r_tubeMax) : T(r_tubeMin)..T(r_tubeMax)
+    φ = mod(T(φMax) - T(φMin), T(2π)) == 0 ? nothing : T(φMin)..T(φMax)
+    θ = mod(T(θMax) - T(θMin), T(2π)) == 0 ? nothing : T(θMin)..T(θMax)
+    Torus( T, T(r_torus), r_tube, φ, θ)
+end
+Torus(r_torus, r_tubeMin, r_tubeMax, φMin, φMax, θMin, θMax) = Torus(;r_torus, r_tubeMin, r_tubeMax, φMin, φMax, θMin, θMax)
```
fhagemann, line 23: Please adjust the constructor for the `Torus` according to the other primitives such that it does not crash with julia-v1.0.

fhagemann (changes requested):

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
hervasa2: Done!

hervasa2:
Separated plotting tools. You will note that in plotting/plotting.jl I replaced the include for VolumePrimitives with SurfacePrimatives. Plotting tools for Volumes are no longer used since each of these is now decomposed into surfaces and then plotted. Once I am done with all the surface primitives, I suggest we remove plotting/VolumePrimitives entirely.

lmh91:
Well, we still have recipes for volumes. They will be indeed simple as they basically only decompose the volume
into surfaces and than call the recipes of the surfaces. 

So regarding the non-plotting changes:
What sample-functions do we have now?
I see `sample_surface` and just `sample`. Maybe we should rename `sample` to `sample_volume` to make things clearer. 

I just tried sampling a cone (with your branch):
```julia
p = SolidStateDetectors.ConstructiveSolidGeometry.Cone(1.0, 1.0, 1.0)
SolidStateDetectors.ConstructiveSolidGeometry.sample_surface(p, 0.1u"m")
```
What I do not like here is that the Cone does not have dimensions and, thus, we should not 
have to give the sampling method a Quantity as we do not know the units of the cone.
(Even we agreed on SI units)
Also, the method samples the cone in phi in 0.1-rad steps what does not correspond to 0.1m.

I suggest that the inner most method should take a tuple of length 3 of Integers. Number of samples
points for each dimension: e.g. `sample_surface(p, (10, 10, 20))`. 
Later we can add more methods which call this basic method.

lmh91 (approved):
Thanks for splitting the plotting parts.

lmh91 (changes requested):
See my comment above regarding the sample methods.

hervasa2:
I will check out the progress in https://github.com/JuliaGeometry/Meshes.jl
Volume primitives have `sample` and `sample_surface` and surface primitives only have `sample`. The idea behind this is that `sample` samples the primitive with the dimensionality of said primitive, therefore it can accept volume and surface primitives as input. Do you like this? Or would you like me to change it to `sample_volume` and `sample_surface` for volume primitives and only `sample_surface` for surface primitives?

I agree that the cone not having units and then sampling with units is not correct. I was thinking that sampling with units would be useful in the context of SSD but it definitely does not make sense just in CSG. 
However, I also have a problem with a tuple of 3 numbers. Since each primitive has a natural system of coordinates, then it would have to transform the tuple into its own system of coordinates. This makes it hard to predict how something will be sampled based on the input. 
However, when you only feed `sample` one number then you can always expect the sampling to be done in the natural system of coordinates of the primitive with that step size. What we could do is drop the units.
You mentioned "Also, the method samples the cone in phi in 0.1-rad steps what does not correspond to 0.1m." It actually does. The phi direction gets sampled in a step of  0.1m/r radians. Therefore the arc from one sample to the next will measure 0.1m. This also avoids oversampling the center. An example of this is in ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl line 83.

lmh91:
Okay, than it is fine like it is now. So, `sample` and `sample_surface`

The inner most method should take number of samples per dimension (in my Opinion).
Note, that dimension does not necessarily mean `x`/`r`/.... (in cases of weird primitives)
And we can/will provide "outer" methods which take one distance and translate this into the corresponding number of samples.
But not as the inner most method. 

This also avoids the problem with the cone. In the case of parsing a distance in meter, e.g. `0.1m`, 
you have to define an radius at which you translate this distance into an angle. 
This we will have to to in the outer methods, but this problem does not occur with giving the number of samples.

hervasa2:
Thank you Lukas. With this 3 integer method one of the dimensions would sometimes be ignored (for surfaces), but that would be ok I guess. Do we care of "oversampling" at small radii since you would have the same number of samples at large and small `r`?

lmh91:
The inner most method of `sample_surface` just takes an tuple of two integers (in general, maybe there are strange 2d primitives
where it is better to parameterize the samping with 3 parameters, same for volumes (maybe 4)).
But this does not matter as the Julia can dispatch as their is a type difference between `NTuple{2,Int}` and `NTuple{3,Int}`.

Yes, oversampling can causes problems. Due to numerically issues. 
E.g., at low radii: In the field calculation the distance between two points becomes important. 
If it is to low, it becomes 0 (numerically). And than the division through those distances yields NaNs/Infs...

hervasa2:
One last question. For sampling in `phi` maybe we could feed it the number of samples we want at max radii instead of number of samples in `phi`. What do you think?

lmh91:
Hmm, isn't that the same? 

We can also add other low-lewel methods were we vary the number of sample points in `phi` along `r`. That might be handy,
depending on what the sample points are going to be used for. 
But to get things going we should just start one simple low level `sample` method in my opinion.

hervasa2:
I meant the number of samples in phi we want at max radii which would set a arc length to sample with. Then you sample with that arc length at smaller radii, thus decreasing your number of phi samples at smaller radii.

lmh91:
Ah, okay. No. At least not for now.

fhagemann:
What is still needed to merge?

hervasa2:
Finishing up the sampling methods today

hervasa2:
Now there are two base sampling methods exist for each surface primitive. I left the one I had but took away the units and added the one requested with a NTuple{3,Int}. Updated sample_surface to work for all volume primitives.
Volume sampling is till needed for volume primitives
Example:
`t = CSG.Torus(4,2,3,0,π,0,2π); plot(t, SSD_style = :surface, c = color)`
Screen Shot 2021-02-04 at 1 59 26 AM
`s = CSG.get_decomposed_surfaces(t)[1]`
Now we sample just one surface primitive with even step 0.3
`samples = CSG.sample_surface(s, 0.3); scatter(samples)`
or
`samples = CSG.sample(s, 0.3); scatter(samples)`
Screen Shot 2021-02-04 at 2 01 18 AM
We can also sample the entire surface of the torus feeding it a single step value or in this case a tuple of 3 ints
`samples = CSG.sample_surface(t, (10,40,40)); scatter(samples)`
Screen Shot 2021-02-04 at 2 25 42 AM

src/ConstructiveSolidGeometry/Sampling.jl
```diff
@@ -0,0 +1,11 @@
...
+    samples = CylindricalPoint{T}[]
+    for surf in get_decomposed_surfaces(c)
+        append!(samples, sample(surf, sampling))
+    end
+    samples
+end
+
+function sample_surface(c::AbstractSurfacePrimitive{T}, sampling::Union{Real, NTuple{3,Int}}) where {T}
+    sample(c, sampling)
+end
```
lmh91, line 11: just do 
```julia
sample_surface(c::AbstractSurfacePrimitive, sampling) = sample(c, sampling)
```
So you also parse other methods we might define later. E.g. if sampling is `1mm`.

src/ConstructiveSolidGeometry/Sampling.jl
```diff
@@ -0,0 +1,11 @@
+function sample_surface(c::AbstractVolumePrimitive{T}, sampling::Union{Real, NTuple{3,Int}}) where {T}
+    samples = CylindricalPoint{T}[]
```
fhagemann, line 2: do we want this to be `CylindricalPoint{T}` by default for any Volume Primitive?
Does that work if you sample a `Box{T}` for example?

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
```diff
@@ -27,11 +27,12 @@ get_z_limits(c::ConalPlane{T}) where {T} = (_left_linear_interval(c.z), _right_l
 in(p::AbstractCoordinatePoint, c::ConalPlane) =
     _in_z(p, c.z) && _eq_φ(p, c.φ) && _in_cyl_r(p, get_r_at_z(c, p.z))
 
-function sample(c::ConalPlane{T}, step::Quantity{<:Real, Unitful.𝐋}) where {T}
+#function sample(c::ConalPlane{T}, step::Quantity{<:Real, Unitful.𝐋}) where {T}
+function sample(c::ConalPlane{T}, step::Real) where {T}
```
lmh91, line 31: That is code duplication.
Just calculate the 3 number of samples per dimension based on `step`
and call the other method.

src/ConstructiveSolidGeometry/SurfacePrimitives/Plane.jl
```diff
@@ -114,6 +113,22 @@ function sample(tri::Plane{T, CartesianPoint{T}, Nothing}, step::Quantity{<:Real
     samples
```
fhagemann, line 113: For instance here, the return value is a  `Vector{CartesianPoint{T}}`. We might run into conversion problems and rounding issues.

src/ConstructiveSolidGeometry/Sampling.jl
lmh91: We actually should not use push here at all. 
We can calculate the number of sample points and can allocated the memory of the corresponding vector.

lmh91: And yes, Cylindrical and Cartesian depends on the primitive.

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
```diff
@@ -41,6 +42,20 @@ function sample(c::ConalPlane{T}, step::Quantity{<:Real, Unitful.𝐋}) where {T
...
 
+function sample(c::ConalPlane{T}, Nsamps::NTuple{3,Int}) where {T}
+    samples = CylindricalPoint{T}[]
+    rbotMin::T, rbotMax::T, rtopMin::T, rtopMax::T = get_r_limits(c)
+    zMin::T, zMax::T = get_z_limits(c)
+    #step = T(ustrip(uconvert(u"m", step)))
+    for z in (Nsamps[3] ≤ 1 ? zMin : range(zMin, zMax, length = Nsamps[3]))
+        r_at_z = get_r_at_z(c, z)
+        for r in (Nsamps[3] ≤ 1 ? _left_radial_interval(r_at_z) : range(_left_radial_interval(r_at_z), _right_radial_interval(r_at_z), length = Nsamps[1]))
+            push!(samples, CylindricalPoint{T}(r,c.φ,z))
```
lmh91, line 53: Do not use `push!`.
Calculate the number of sample points and allocate the vector `samples` with corresponding length.

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
hervasa2: It is not code duplication. In one case you cave a constant step regardless of r width, which does not translate into a constant number of samples. You cannot know the number of samples output by `sample(c::AbstractSurfacePrimitive, step::Real)` without knowing the geometry, whereas `sample(c::AbstractSurfacePrimitive, (a,b,c) )` will always output a*b, a*c, or b*c samples regardless of the geometry.

hervasa2: Also the `NTuple{3,Int}` method is meant to always contain the edges of the primitive whereas the the `step::Real` method is not

lmh91: I count 5 copied lines. 
And here you also only need 2 intergers: `NTuple{2,Int}`
You do now the geometry. 
```julia
sample(c::AbstractSurfacePrimitive, step::Real) = 
   sample(c, (phi_width(c)/step, z_width(c)/step)) # Pseudo code
```
The edges should always be included.

hervasa2: Yes you only need two integers in all of these but then you lose the interpretability of reading `NTuple{2,Int}` as in a certain coordinate system. This would then lead to loss of generality of the sample_surface function since you would have to map `NTuple{3,Int}` to different 'NTuple{2,Int}'
Yes you know the geometry, I was just making the point of how these sample functions differ.
In your pseudocode `z_width(c)` varies with r and therefore not a constant parameter you can just pass

hervasa2: I agree that edges should be included. Therefore the sample(Nsamps) will be the default fast and easy sampling method. However I still see value in having a base sampling function which samples the surface with an even step. There is no loss of performance or code duplication here although yes the code is very similar, they differ in a small but meaningful way.

fhagemann: This goes along with my comment on the `Cone{T}` sampling:

I see David's point. Let us assume that we want to sample a `Cone{T}` that is 0.1m wide at the bottom and 0.05m wide at the top with a `step` of 0.01m. According to the pseudo code, this would call `sample` with 10 samples in `r`. However, these 10 samples will be differently spaced at different height and not fulfill the requirement of all being 0.01m apart.

Cone_spacing

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
lmh91: Partially correct.
We will only only use the sample method with a low number of points to add grid points at the beginning.
Afterwards, we will use the sample (surface) method with a higher number of grid points (depending on the current grid density) to paint the contacts on the grid. 
Because the primitives might get translated and rotated, keep that in mind.
At least, that's how I imagine this right now. We will have to see how things develop. 

That's why I just wanted a simple sample method for the beginning so we can continue and actually find out what will work preferably.

fhagemann: So, your proposal is to have a basic `sample` function based on `NTuple{3,Int}` for the moment and write convenience functions for what might come?

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
lmh91: See my comment below the other comment of @fhagemann. 
So okya, leave the two methods like they are but only some minor changes. 
Please add the keyword `stop` to the `range`-methods to increase readability. 
Also, create the ranges befor the loop and calculate their length and allocate the respective vector of points
such that you do not need to use `push!` anymore. 
Actually, that is how we can get rid of the code duplications. Create one `method` where you parse two ranges/vectors to it.

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
lmh91: Well, I would actually add a method which get `ranges/Abstractvector`  over which the loops run to reduce code duplications.
Both current methods `NTuple` and `step` should than calculate the ranges and call the low-level method. 
See comment above :sweat_smile:

fhagemann: But same problem: the `r` range will be different at different `z`, for the `Cone{T}` at least.

lmh91: It's a linear scaling, one could just add a scaling parameter to the method.
That was my hole point from the beginning! Just start with the simple `NTuple` version.
Thanks to Julia we can add very easily more complex and advanced sampling methods without having
to change other parts of the code. But, this we should do once we see certain issues with the simple `NTuple` version.

hervasa2: Exactly, this happens sometimes. However I could write general 1D sampling methods that we could use wherever possible eg. `sample_z(c::AbstractPrimitive{T}, Nsamps::NTuple{3,Int})`, `sample_z(c::AbstractPrimitive{T}, step::Real)`. These would go in Sampling.jl and call on the eg. `get_z_limits` that we have already written when applicable. We would have to use these with care because something like a `TorusMantle` would not have the `get_z_limits`. Therefore these 1D sampling methods would fail when applied to the incorrect primitive. However, they would be general and only internal and used as much as possible to reduce a lot of code duplications of `get_coordinate_limits`.

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
lmh91: I mark this as resolved and continue to write to the conversation below as it is basically the same discussion.

src/ConstructiveSolidGeometry/SurfacePrimitives/ConalPlane.jl
hervasa2: > It's a linear scaling, one could just add a scaling parameter to the method.
It would need a linear scaling and offset, which I think is overkill to add in the method, we can scale and offset the output when needed. However

> That was my hole point from the beginning! Just start with the simple `NTuple` version.
> Thanks to Julia we can add `later` very easily more complex and advanced sampling methods without having
> to change other parts of the code. But, this we should do once we see certain issues with the simple `NTuple` version.

The step method was the first one I made, so all I did was not delete it.

hervasa2: @lmh91 should I write the 1D sampling methods to create ranges as you suggested?

lmh91: No, but keep it in mind as we might want to add it later. 
I want to keep it as simple as possible (this includes as little as code as possible). 
Once we face issues, we can add more advanced solutions. 

The sampling methods you already have you can leave inside. Just please remove the usage of `push!`. 
And make sure to return the correct point types. (@fhagemann comments above regarding the file `sampling.jl`)

oschulz:
Regarding

```julia
samples = CylindricalPoint{T}[]
rbotMin::T, rbotMax::T, rtopMin::T, rtopMax::T = get_r_limits(c)
zMin::T, zMax::T = get_z_limits(c)
#step = T(ustrip(uconvert(u"m", step)))
for z in (Nsamps[3] ≤ 1 ? zMin : range(zMin, zMax, length = Nsamps[3]))
    r_at_z = get_r_at_z(c, z)
    for r in (Nsamps[3] ≤ 1 ? _left_radial_interval(r_at_z) : range(_left_radial_interval(r_at_z), _right_radial_interval(r_at_z), length = Nsamps[1]))
        push!(samples, CylindricalPoint{T}(r,c.φ,z))
```

Why not just

```julia
samples = [
    CylindricalPoint{T}(r,c.φ,z))
    for z in (Nsamps[3] ≤ 1 ? zMin : range(zMin, zMax, length = Nsamps[3]))
    for r in (Nsamps[3] ≤ 1 ? _left_radial_interval(r_at_z) : range(_left_radial_interval(r_at_z), _right_radial_interval(r_at_z), length = Nsamps[1]))
]
```

lmh91:
Thanks all :)
Now we can start integrating the refactored CSG into SSD.

fhagemann:
I updated #110

Issue #132, opened 2021-01-08
Asymmetric artefacts for symmetric objects in cartesian grid


leenderthayen:
Hello,

When preparing toy examples for some of the previous issues posted today, I noticed that an asymmetry in the point types (and upon later inspection also on the electric potentials and every other quantity) was not a result of an asymmetric x-y geometry (like a hexagon), but also appears in the case of round tubes. As an example, I've attached the point types and electric potential of a simple tube in different planes.

![bt_detector_point_types_230 0V](https://user-images.githubusercontent.com/11664090/104073403-9ab25c00-51db-11eb-8976-0daa92c1ccd2.png)

![bt_detector_electric_potential_450 0V](https://user-images.githubusercontent.com/11664090/104073421-a69e1e00-51db-11eb-90a6-5a8b6463c435.png)

[big_tube.txt](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/5789697/big_tube.txt)

Looking also at edges which are not parallel to x or y, there appear grid artefacts which can be tamed somewhat with increasing grid spacing, but remain visible.

In one case, this became particularly apparent

![grid_artifact_ring](https://user-images.githubusercontent.com/11664090/104073598-1d3b1b80-51dc-11eb-91d7-5f42970d0d6e.png)

This is the central pixel of a hexagonal pixel ring. Every pixel contains roughly the same asymmetry, with a minimum along the x-axis through the center of the pixel and a butterfly structure around it.

Thank you!
Leendert

lmh91:
There was indeed a bug. Please try it again with the current master branch.

leenderthayen:
This seems to have fixed things nicely. Thank you for the fast response!

leenderthayen:
While it looks like things are fixed for spherically symmetric geometries, using for example a hexagonal pixel introducing further asymmetric grid artefacts.

![Test_hex_detector_electric_potential_10 0V](https://user-images.githubusercontent.com/11664090/104237601-4bac3700-5426-11eb-8b96-21be19cf03ee.png)
![Test_hex_detector_electric_field_10 0V](https://user-images.githubusercontent.com/11664090/104237621-536bdb80-5426-11eb-9da3-f672a75515ff.png)

[simple_si_pixel.txt](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/5798322/simple_si_pixel.txt)

lmh91:
Okay, I think there is no problem here. 
[...]

PR #133, opened 2021-01-11
fix initial grid spacing based on world size

lmh91:
This PR fixes a bug in the creation of the initial grid based on the dimensions of the world. 
See #132 (thanks @leenderthayen)

Thanks @fhagemann for identifying the corresponding lines in the source code.

### DIFF
--- a/src/SolidStateDetector/DetectorGeometries.jl
+++ b/src/SolidStateDetector/DetectorGeometries.jl
@@ -495,10 +495,10 @@ function Grid(  detector::SolidStateDetector{T, :cylindrical};
         else
             error("This should not happen... World has no dimension")
         end
-        init_grid_size_1::Int = convert(Int, round(ratio * world_diffs[inds[1]], RoundUp))
-        init_grid_size_2::Int = convert(Int, round(ratio * world_diffs[inds[2]], RoundUp))
-        init_grid_size_3::Int = convert(Int, round(ratio * world_diffs[inds[3]], RoundUp))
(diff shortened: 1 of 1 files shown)


fhagemann:
We might think about adding comparisons to analytic solutions for differently spaced grids to the test script.

lmh91:
> We might think about adding comparisons to analytic solutions for differently spaced grids to the test script.

Feel free to add it ;)

fhagemann:
> > We might think about adding comparisons to analytic solutions for differently spaced grids to the test script.
> 
> Feel free to add it ;)

I just had a quick look (one can just adjust the `init_grid_size` in the test script), and I found that the test fails for small grid dimensions (for the coaxial detector in cartesian coordinate system `init_grid_size = (20,20,2)` does result in a failed test, probably because of the discretisation error being too large). Choosing asymmetric grid dimensions with big enough dimensions, e.g. `init_grid_size = (40,20,2)`, does not fail.

However, the criteria for the test passing is set higher for Cartesian coordinates compared to cylindrical coordinates.
(Capacitance: `rtol = 0.06` vs. `rtol = 0.01`;  Potential: `RMS < 1.1` vs. `RMS < 0.01`)

lmh91:
You have to be careful. The `init_grid_size` were specially chosen due to reasons of the boundary handling. 

Yes, the criteria is higher for a cartesian grid as the geometry has a cylindrical feature. Thus, for a coarse grid, it is not as good as on an cylindrical grid.

PR #135, opened 2021-02-07
Update to Julia v1.6

lmh91:
PR to prepare for Julia v1.6. 
Do not merge till v1.6 is released.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         version:
           - '1.0'
-          - '1.5'
+          - '1'
           - 'nightly'
         os:
           - ubuntu-latest
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1.5'
+          version: '1'
       - name: Cache artifacts
         uses: actions/cache@v2
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         version:
           - '1.0'
-          - '1.5'
+          - '1.6'
```
oschulz, line 23: We should set this to '1', that will automatically use the latest Julia release.

.github/workflows/ci.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1.5'
+          version: '1.6'
```
oschulz, line 73: We should set this to '1', that will automatically use the latest Julia release.

oschulz (changes requested):

oschulz:
Nice - like this, we don't have to wait for the v1.6 release to merge it.

lmh91:
That's true. 

Should we actually drop v1.0 already @oschulz?
v1.6 should become the next LTS version, right? 
Or do we wait till v1.7?

oschulz:
Let's set Julia v1.3 as the minimum right now (and also test on v1.3), many package have done that. If v1.6 seems likely to become the new LTS, but this will only be confirmed during the v1.7 release process, from what I understand.

lmh91:
Okay, than I will leave v1.0 inside for now.

PR #138, opened 2021-02-17
Surfaceprimitives

hervasa2:
I have added the distance_to_surface function for all the conal surfaces (ConeMantle, CylindricalAnnulus, ConalPlane) and made small corrections to their dependency in SurfacePrimitives/Plane.jl. An example follows
`cone = CSG.Cone(0, 1, 1, 3, π/6, 3π/2, -3, 3)`
`plot(cone, SSD_style = :surface, color = color, camera = (30,30))`
`plot!(cone, SSD_style = :wireframe, color = :black)`
distance_to_cone
We can scan the cone along z to explore the distance to surface of the cone for each point in the world
`D[i,j] = CSG.distance_to_surface(CSG.CylindricalPoint{T}(r,φ,z),cone)`
`heatmap(Φ,R,D, c = cgrad(:inferno, scale =:exp), clims = (0,3), projection = :polar, title = string("z = ",z))`
zcuts
Lest take a closer look at z = 2. I have also used the `in` function as a further check
`point = CSG.CylindricalPoint{T}(r,φ,2)`
 `D[i,j] = point in a ? CSG.distance_to_surface(point,cone) : NaN`
`heatmap(Φ,R,D, c = cgrad(:inferno, scale =:exp), projection = :polar, title = "z = 2")`
z2cut
We can also scan the cone along φ 
`D[j,i] = CSG.distance_to_surface(CSG.CylindricalPoint{T}(r,φ*π/180,z),cone) `
`heatmap(R,Z,D, c = cgrad(:inferno, scale =:exp), clims = (0,3), title = string("φ = ", φ, "°"), aspect_ratio = 1)`
phicuts
Lest take a closer look at φ = 90°
`point = CSG.CylindricalPoint{T}(r,π/2,z)`
`D[j,i] = point in a ? CSG.distance_to_surface(point,cone) : NaN`
`heatmap(R,Z,D, c = cgrad(:inferno, scale =:exp), title = string("φ = 90°"), aspect_ratio = 1)`
phi90cut

PR #139, opened 2021-02-18
Surfaceprimitives

hervasa2:
Distance to surface for all surface primitives. See #138


lmh91:
I will close this for now as I first want to get the refactored CSG and charge drift fully working. 
Than, we can have a look at this again.

Issue #67, opened 2020-10-09
Parse crystal orientation to drift model


lmh91:
Right now, the drift model assumes that the crystal axis <001> is parallel to the z-axis. 

This issue is kinda related to #62.

PR #144, opened 2021-04-20
Updates to ADLChargeDriftModel

fhagemann:
The source code with the implementation of the `ADLChargeDriftModel` was changed such that now also silicon can be simulated in a proper way, taking into account the conduction band nature which is different from germanium.

Silicon has its conduction band minima in the <100> directions, such that the whole formulas to calculate the electron drift velocity have to be adjusted. I summarized the calculations to determine the formulas [here](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/6346805/Drift_models_Silicon.pdf).

### Changes in the source code
- A new abstract type was introduced, `AbstractDriftMaterial`, which describes materials in which we drift charges in SSD.
Right now, only germanium (`HPGe`) and silicon (`Si`) are implemented. The material is also saved as a field in `ADLChargeDriftModel`.
- An example config file was added for silicon at 300K.
- Tests were added to compare the resulting charge drift velocities to the ones defined by the parametrisation in the function `Vl`.

### Literature values for silicon
- Values for the effective electron masses were taken from [Dexter _et al._ (1956)](https://journals.aps.org/pr/abstract/10.1103/PhysRev.104.637).
- The parameters b and c, describing the warped structure of the valence band for holes, are also taken from [Dexter _et al._ (1956)](https://journals.aps.org/pr/abstract/10.1103/PhysRev.104.637).
- The fit functions for `Λ(k0)`, `Ω(k0)` and `k0(vrel)` were determined by fitting numerically computed values to polynomials.
- As an example config file at a temperature of 300K, I used the values tabulated in Tab. 1 of [Becker _et al._ (2011)](https://doi.org/10.1016/j.sse.2010.10.009).

### Other things to be addressed
- Silicon might not be best described by `ADLChargeDriftModel`.
- In the hole drift model, the approximations in `Λ(k0)`, `Ω(k0)` and `k0(vrel)` lead to errors in the result. Especially comparing the hole drift velocity in <111> direction results in a relative error of already 0.5% from the parameterisation. We might want to find a better fit function that allows reducing deviations in the hole drift model.


fhagemann:
Maybe some more background information why the test for the hole drift velocity in <111> direction needs more tolerance than all others: The functions `Λ(k0)` and `k0(vrel)` are approximated with polynomials. Note that `vrel = v111h/v100h` and as the hole velocity along <111> direction should be smaller than the one along <100>, `vrel` should always be smaller than 1.

Plugging in values and determining `Λ(vrel)` analytically, the result is `Λ(vrel) = 0.75 * (1 - vrel)`. This is not perfectly reproduced when using `Λ(k0)` and `k0(vrel)` to calculate `Λ(k0(vrel))`:

Lambda_k0

The differenced between the analytical `Λ(vrel)` and the approximated  `Λ(k0(vrel))` will automatically result in an error of the resulting hole drift velocity. Most of the time `vrel` will be around 0.8 or 0.9, where the fit function fits the best. However, deviations
become quite noticeable for cases with almost no anisotropy, i.e. `vrel = 1` which corresponds to `v100h = v111h`.
It might be more desirable to use the analytic expression for `Λ(vrel)` and maybe find an analytical expression for `Ω(vrel)`, to reduce avoidable errors in cases where the anisotropy is small (low electric fields).

oschulz:
Very nice!

fhagemann:
I also compared numerical values for `Ω(vrel)` (I did not find any way to easily find a numerical solution like in the case for `Λ(vrel)`) to results when using the current polynomial fit functions from SSD, `Ω(k0)` and `k0(vrel)`, to calculate `Ω(k0(vrel))`:

Omega_k0

Also, here `Ω(vrel)` should be zero at `vrel = 1`. Nonzero values for `Ω(vrel = 1)` will result in a hole drift not parallel to the field, even if the hole velocity is isotropic, i.e. `v100h  = v110h = v111h`, which should **NOT** be the case. 

In order to skip the detour over `k0`, we could determine a polynomial fit for `Ω(vrel)` in orders of `(1-vrel)` directly from the numerical values and force `Ω(vrel = 1)` to be zero.
Results for a fourth-order polynomial with no offset:
```
Ge: Ω(vrel) =  -0.29711*(1-vrel) -1.12082*(1-vrel)² +3.83929*(1-vrel)³ -4.80825*(1-vrel)⁴
Si: Ω(vrel) =  -0.30565*(1-vrel) -1.19650*(1-vrel)² +4.69001*(1-vrel)³ -7.00635*(1-vrel)⁴
```

Omega_k0_new

fhagemann:
To be honest, I would be in favour of changing the current implementation of the hole drift . 
Instead of calculating `Λ(vrel)` and `Ω(vrel)` via a detour over `k0` (which is a dimensionless quantity representing the magnitude of the mean wave vector of heavy holes when applying an electric field), I would suggest to use:
- the analytical expression `Λ(vrel) = 0.75 * (1 - vrel)` to model the longitudinal anisotropy, and
- polynomial fits to the numerical values for  `Ω(vrel)` in orders of `(1-vrel)` and force `Ω(vrel = 1) = 0`.

Running the tests, the errors in the hole drift implementation along <111> is not 0.5% anymore but of the order of rounding errors.

What do you think?

lmh91:
Very nice!

I would prefer the faster computation. 

The accuracy does not really matter I guess, 
since there are larger uncertainties at different quantities which do influence the drift. 
On which level does these changes influence the drift paths and waveforms? 

We could/should put all this than into the documentation.

fhagemann:
My changes would concern the calculation of `Λ(vrel)` and `Ω(vrel)`, which define the hole drift velocity in the `ADLChargeDriftModel`.

`Λ(vrel)` defines the longitudinal anisotropy of the hole drift.
With this change, the hole drift velocity along <100> will remain unchanged, the drift velocity along <111> and <110> should be somewhat closer to the real value.

`Ω(vrel)` defines the transverse anisotropy of the hole drift.
With this change, if the given input velocities for hole drifts along <100> and <111> are identical, the movement along **ALL** directions should have the same magnitude, independent of the direction (achieved by forcing `Ω(vrel = 1) = 0`). This was **NOT** the case before. It might be that this caused some trouble with "wiggling" hole movement at low electric fields, where `v100h` and `v111h` are very close to each other.
 
I will create some plots to show differences in drift paths and waveforms emerging from these changes.

fhagemann:
I ran some simulation over an infinitely long coaxial detector with a radial impurity gradient: [simple_Coax.json](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/6357653/simple_Coax.txt)

Spawning some charges close to the centre and letting them drift outside looks almost identical.
(Differences in the plot with zoomed-in z-axis are in the order of type precision)

#### Old functions for `Λ(vrel)` and `Ω(vrel)`

old_h_drift


#### New functions for `Λ(vrel)` and `Ω(vrel)`
new_h_drift

#### Changes in waveforms

For a point on the <110> axis, drifting outside:
pulses

fhagemann:
However, I see significant changes for complicated geometries.
The standard inverted coaxial detector from our config file with the standard `ADLChargeDriftModel` parameters for germanium:

#### Old functions for `Λ(vrel)` and `Ω(vrel)`
old_h_drift_IVC



#### New functions for `Λ(vrel)` and `Ω(vrel)`
new_h_drift_IVC



#### Changes in waveforms

For a point initially starting on the <110> axis, drifting towards the core contact (left full pulse, right zoomed in to the end):
pulses_ivc

fhagemann:
Also, the current implementation of the hole drift model gives very unphysical results, if the hole drift velocity in <111> direction is higher than the hole drift velocity in <100> direction (at any electric field strength or temperature).

This happens for one of our example ADL config files: `drift_velocity_config_alternative.json`.

Find a plot where the parameterisation of `v100h` and `v111h` used as input to the simulation are shown as dashed lines, and the actual calculated `v100h`, `v110h` and `v111h` that are the output of the `ADLChargeDriftModel` are shown as solid lines.

As said, the velocity along <100> is not influenced, but the velocity along <110> and <111> is described way more accurate.

#### Old functions for `Λ(vrel)` and `Ω(vrel)`
old_drift_output

#### New functions for `Λ(vrel)` and `Ω(vrel)`
new_drift_output

schustermartin:
Very nice work!! I believe this is a straight upgrade. After a quick look and our discussion (@fhagemann), I think this is good to go. So maybe you can add the new calculations for `Λ(vrel)` and `Ω(vrel)` to this pull request before we merge?

fhagemann:
There you go.
We might want to add a warning when the hole drift velocity is higher along <111> than along <100> for any given electric field strength. According to the hole drift model, that should never be the case.

fhagemann:
I spent some time reducing allocations and ensure type stability in the source code of the `ADLChargeDriftModel`.
Here are some benchmarks for the determination of electron and hole drift velocities:

#### Current implementation on master branch

benchmark_old

#### Implementation in this PR before speed-up commit (652244f):

benchmark_new_before_speedup

#### Implementation in this PR after speed-up commit (652244f):

benchmark_new_after_speedup

fhagemann:
For now, I will **NOT** add a warning if the hole drift velocity along <111> is greater than along <100>
because the new implementation will not give completely unphysical results.

When writing the documentation we can still think about introducing warnings for such cases.

If the tests all pass, this PR is ready to be merged.

fhagemann:
I could also implement arbitrary orientation of the crystal axes (#67).
However, we need consistent syntax in the config files for this.

src/ChargeDriftModels/ADL/ADL.jl
```diff
@@ -150,53 +174,73 @@ function ADLChargeDriftModel(configfilename::Union{Missing, AbstractString} = mi
...
+@inline get_AE_and_RE(cdm::ADLChargeDriftModel{T}, V100e::T, V111e::T) where {T} = _get_AE_and_RE(cdm, cdm.material, V100e, V111e)
 
+function _get_AE_and_RE(cdm::ADLChargeDriftModel{T}, ::Type{HPGe}, V100e::T, V111e::T)::Tuple{T,T} where{T <: SSDFloat}
+    mli::T = 1. / cdm.masses.ml
+    mti::T = 1. / cdm.masses.mt
+    B::T = 1/3 * sqrt(8*mti + mli)
 
+    Gamma0::T = sqrt(2/3*mti + 1/3*mli)
+    Gamma1::T = (-4*sqrt(mli) -4/3 * B)/(sqrt(mli) - B)^2
+    Gamma2::T = Gamma1*(sqrt(mli) + 3*B)/(-4)
```
lmh91, line 198: This can be precalculated, right?
It does not involve `v110e` or `v111e`.

src/ChargeDriftModels/ADL/ADL.jl
```diff
@@ -150,53 +174,73 @@ function ADLChargeDriftModel(configfilename::Union{Missing, AbstractString} = mi
...
 
-@fastmath function getVe(fv::SVector{3, T}, cdm::ADLChargeDriftModel{T}, Emag_threshold::T = T(1e-5))::SVector{3, T} where {T <: SSDFloat}
+function _get_AE_and_RE(cdm::ADLChargeDriftModel{T}, ::Type{Si}, V100e::T, V111e::T)::Tuple{T,T} where{T <: SSDFloat}
+    
+    mli::T = 1. / cdm.masses.ml
+    mti::T = 1. / cdm.masses.mt
+
+    Gamma0::T = sqrt(2/3*mti + 1/3*mli)
+    Gamma1::T = -1.5 * (sqrt(mti) + 2*sqrt(mli)) / (sqrt(mli) - sqrt(mti))^2
+    Gamma2::T = 0.5 * (sqrt(mli) + 2*sqrt(mti)) * (2*sqrt(mli) + sqrt(mti)) / (sqrt(mli) - sqrt(mti))^2
```
lmh91, line 213: same here

lmh91 (commented):
Since we are going to not precalculate and interpolate the drift fields soon, 
we should try to get this as fast as possible. 
Please have a look which calculations can be precalculated. 

Otherwise it looks fine to me.

lmh91:
Hmm, this should not be. You save some calculations including multiple divisions and `sqrt`'s.
There is one allocation. That's proabably causing it. Is your last comment with the new commit?

fhagemann:
My latest commit (80e94a7) speeds up the code by precalculating values.

Electron drift model:
- The effective masses (used to be stored in `MassParameters{T}` and the parameters, `Γ0`, `Γ1` and ` Γ2`, used for the calculation of the electron drift velocity are now stored in `ADLParameters{T}`. As only the inverted effective masses are needed, the inverted values for `ml` and `mt` are saved. The parameter `Γ0` is also saved inverted compared to the `Γ0` defined in the PDF at the beginning of this PR. This will go into the documentation.
- The sum over all `oneOverSqrtEγE` and `1/N` is precalculated before the loop in `getVe`.

Hole drift model:
- All the sine and cosine values are determined with `sincos` before the actual calculation of the velocity components.
- The rotation matrices are replaced by matrices from the Rotations package.


Other changes:
- The `material` for the `ADLChargeDriftModel` is now saved as parametric type (`M`).

PR #145, opened 2021-04-22
Corrections to Torus and Cone and respective surfaces + Negative Angle Support

hervasa2:
1. Removed unique call in get_decomposed_surfaces of Torus and Cone
- proper checks are done so that no surfaces are repeated
- when possible reduced the number of surfaces produced
- can now handle negative angles properly (see 3)
2. Added z field to Torus (and TorusMantle and TorroidalAnnulus)
- so that a translated torus in z remains a torus and does not become a translated geometry
3. Corresponding adjustments for Torus z-field: plotting, constructors for surfaces on Torus, PointsAndVectors
4. Minor changes to mesh plotting of surface primitives
5. Added general support for negative angles
- changes to PointsAndVectors which need additional functions (in Intervals.jl)
- checks if negative angles are in arbitrary intervals (in Intervals.jl)
- unions of periodic intervals (in Intervals.jl)


src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+        return true
+    elseif isnothing(union_angular_intervals(α, β))
+        return false
+    elseif (mod(α.left, T(2π)) == mod(β.right, T(2π)) && !_in_angular_interval_open(α.right, β)) ||
+           (mod(α.right, T(2π)) == mod(β.left, T(2π)) && !_in_angular_interval_open(α.left, β))
+        return false
+    else
+        return true
+    end
+end
```
fhagemann, line 94: I feel like this function could be shorter.
```julia
function example(a,b,c)
    if a 
         return true
    elseif b && c
         return false
    else
         return true
    end
end
```
can, for example, be simplified to `example(a,b,c) = a && !(b && c)`

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -99,37 +99,44 @@ get_φ_limits(c::Cone{T, <:Any, <:AbstractInterval, <:Any}) where {T} = (c.φ.le
 
 get_z_limits(c::Cone{T}) where {T} = (_left_linear_interval(c.z), _right_linear_interval(c.z))
 
-function _get_decomposed_surfaces_cone(c::Cone{T}, rbotMin, rbotMax, rtopMin, rtopMax, zMin, zMax) where {T}
+function _is_cone_collapsed(rbotMin::T, rbotMax::T, rtopMin::T, rtopMax::T, zMin::T, zMax::T) where {T}
```
fhagemann, line 102: Can well call this something like `_is_2D_cone`? I find `collapsed` a strange wording for this.. Or is that 
the scientific term I don't know about?

Can you briefly state the purpose of this function, just to be sure that I understand it right?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -1,89 +1,130 @@
-struct Torus{T,TR,TB,TP,TT} <: AbstractVolumePrimitive{T}
+struct Torus{T,TR,TB,TP,TT,TZ} <: AbstractVolumePrimitive{T}
```
fhagemann, line 1: I'm not sure, whether we need that `TZ`, because `z` is **ALWAYS** of type `T`

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -1,89 +1,130 @@
...
+
+function Torus(r_torus::R1, r_tube::R2, z::TZ) where {R1<:Real, R2<:Real, TZ<:Real}
+    T = float(promote_type(R1, R2, TZ))
+    Torus( T, T(r_torus), T(r_tube), nothing, nothing, T(z))
+end
+
+function RoundChamfer(r_torus::R1, r_tube::R2, z::TZ) where {R1<:Real, R2<:Real, TZ<:Real}
+    T = float(promote_type(R1, R2, TZ))
+    Torus( T, T(r_torus), T(r_tube), nothing, T(0)..T(π/2), T(z))
 end
```
fhagemann, line 36: The function `RoundChamfer` is never called nor is it exported. What is this?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -1,89 +1,130 @@
...
+get_r_tube_limits(t::Torus{T}) where {T} = (_left_radial_interval(t.r_tube),_right_radial_interval(t.r_tube))
 
 get_φ_limits(t::Torus{T, <:Any, <:Any, Nothing, <:Any}) where {T} = (T(0), T(2π), true)
 get_φ_limits(t::Torus{T, <:Any, <:Any, <:AbstractInterval, <:Any}) where {T} = (t.φ.left, t.φ.right, false)
 
 get_θ_limits(t::Torus{T, <:Any, <:Any, <:Any, Nothing}) where {T} = (T(0), T(2π), true)
 get_θ_limits(t::Torus{T, <:Any, <:Any, <:Any, <:AbstractInterval}) where {T} = (t.θ.left, t.θ.right, false)
 
-function _get_decomposed_surfaces(t::Torus{T}) where {T}
+function _is_torus_collapsed(t::Torus{T}) where {T}
```
fhagemann, line 58: Same as with the cone: do we want to use `collapsed` as wording here?

fhagemann (commented):
Very nice!

Could you add tests to the `CSG_tests.jl` for:
- `Torus` volume primitive
- existing surface primitives
- checks for angular intervals to see whether values outside the `0..2π` is error-free (especially in extreme cases close to `0` or `2π` or multiple of `2π`)

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+
+@inline get_angular_interval(::Type{T}, α::AbstractInterval{T}) where {T} = α
+@inline get_angular_interval(::Type{T}, α::Nothing) where {T} = T(0)..T(2π)
+
+@inline _in_angular_interval_closed(α::Real, α_int::Nothing) = true
+@inline _in_angular_interval_closed(α::Real, α_int::Nothing, tol::Real) = true
+@inline _in_angular_interval_closed(α::Real, α_int::AbstractInterval{T}) where {T} = mod(α - α_int.left, T(2π)) ≤ α_int.right - α_int.left
+@inline _in_angular_interval_closed(α::Real, α_int::AbstractInterval{T}, tol::Real) where {T} = mod(α - (α_int.left-tol), T(2π)) ≤ (α_int.right+tol) - (α_int.left-tol)
+
+@inline _in_angular_interval_open(α::T, α_int::Nothing) where {T<:Real} = 0 < mod(α, T(2π)) < 2π
```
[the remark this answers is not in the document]
hervasa2: `_in_angular_interval_open` is not used for now. But it will be needed for surface interactions. Cutting angular intervals is tricky because with just one cut you cannot guarantee you will end up with 2 intervals. With full `2π` intervals you need two cuts which are neither at 0 or `2π` to get something that is actually cut.

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+    end
+end
+
+function is_intersection_an_interval(α::AbstractInterval{T}, β::AbstractInterval{T}) where {T}
+    if (mod(α.right - α.left, T(2π)) == 0 || mod(β.right - β.left, T(2π)) == 0)
+        return true
+    elseif isnothing(union_angular_intervals(α, β))
+        return false
+    elseif (mod(α.left, T(2π)) == mod(β.right, T(2π)) && !_in_angular_interval_open(α.right, β)) ||
+           (mod(α.right, T(2π)) == mod(β.left, T(2π)) && !_in_angular_interval_open(α.left, β))
```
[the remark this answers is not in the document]
hervasa2: `mod2pi(2π)  != 0` and `mod2pi(T(2π))  != 0` 
From jlhub: "This function computes a floating point representation of the modulus after division by numerically exact 2pi, and is therefore not exactly the same as mod(x,2pi), which would compute the modulus of x relative to division by the floating-point number 2pi"

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+@inline get_angular_interval(::Type{T}, α::AbstractInterval{T}) where {T} = α
+@inline get_angular_interval(::Type{T}, α::Nothing) where {T} = T(0)..T(2π)
+
+@inline _in_angular_interval_closed(α::Real, α_int::Nothing) = true
+@inline _in_angular_interval_closed(α::Real, α_int::Nothing, tol::Real) = true
+@inline _in_angular_interval_closed(α::Real, α_int::AbstractInterval{T}) where {T} = mod(α - α_int.left, T(2π)) ≤ α_int.right - α_int.left
+@inline _in_angular_interval_closed(α::Real, α_int::AbstractInterval{T}, tol::Real) where {T} = mod(α - (α_int.left-tol), T(2π)) ≤ (α_int.right+tol) - (α_int.left-tol)
+
+@inline _in_angular_interval_open(α::T, α_int::Nothing) where {T<:Real} = 0 < mod(α, T(2π)) < 2π
+@inline _in_angular_interval_open(α::Real, α_int::AbstractInterval{T}) where {T} = 0 < mod(α - α_int.left, T(2π)) < α_int.right - α_int.left
```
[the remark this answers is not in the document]
hervasa2: It passes all my checks

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+function _is_edge_of_angular_interval_union(val::T, α::AbstractInterval, β::AbstractInterval) where {T<:AbstractFloat}
+    tol = 10*geom_atol_zero(T)
+    (_in_angular_interval_union(val+tol, α, β) && !_in_angular_interval_union(val-tol, α, β)) ||
+    (_in_angular_interval_union(val-tol, α, β) && !_in_angular_interval_union(val+tol, α, β))
+ end
+
+function union_angular_intervals(α::AbstractInterval{T}, β::AbstractInterval{T}) where {T} #if no intersection will return nothing
+    tol = 10*geom_atol_zero(T)
+    if α == β
+        return α
```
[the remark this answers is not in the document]
hervasa2: Our intervals for geometries are always closed, so I haven't considered that. SHould I leave it as is or restrict it to closed intervals?

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+        if edges < 3 && _is_edge_of_angular_interval_union(β.right, α, β) && mod(β.right, T(2π)) ≠ mod(α.left, T(2π)) && mod(β.right, T(2π)) ≠ mod(α.right, T(2π))
+            edges == 1 ? θ2 = β.right : nothing
+            edges = edges + 1
+        end
+        if edges == 2
+            θMin, θMax = minmax(mod(θ1, T(2π)), mod(θ2, T(2π)))
+            return _in_angular_interval_union(θMin + tol, α, β) ? θMin..θMax : (θMax-T(2π))..θMin
+        elseif edges == 0
+            return T(0)..T(2π)
+        end
```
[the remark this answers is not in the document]
hervasa2: I should have put some comments. If there is a contiguous union it returns the resulting interval (even if it is `T(0)..T(2π)`). When there is no contiguous union it returns nothing (edges different than 2 or 0). In cases like this for non periodic intervals IntervalSets.jl returns an error.

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+    end
+end
+
+function is_intersection_an_interval(α::AbstractInterval{T}, β::AbstractInterval{T}) where {T}
+    if (mod(α.right - α.left, T(2π)) == 0 || mod(β.right - β.left, T(2π)) == 0)
```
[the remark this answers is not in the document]
hervasa2: I checked and these should be fine.

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -20,9 +20,9 @@ end
 @inline _in_cyl_r(p::CartesianPoint, r::Real) = hypot(p.x, p.y) <= r
 @inline _in_cyl_r(p::CartesianPoint, r::AbstractInterval) = hypot(p.x, p.y) in r
 
-@inline _eq_φ(p::CartesianPoint{T}, φ::Real) where {T} = mod(atan(p.y, p.x), T(2π)) == T(φ)
+@inline _eq_φ(p::CartesianPoint{T}, φ::Real) where {T} = isapprox(mod(atan(p.y, p.x), T(2π)), T(φ), atol = geom_atol_zero(T))
```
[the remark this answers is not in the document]
hervasa2: Changed to `_isapprox_...` to mimic julia's isapprox

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -38,12 +38,12 @@ end
...
+@inline _eq_torr_r_tube(p::CartesianPoint{T}, r_torus::Real, r_tube::Real, z_torus::Real) where {T} = isapprox(hypot(hypot(p.x, p.y) - r_torus, p.z - z_torus), T(r_tube), atol = geom_atol_zero(T))
 
-@inline _in_torr_r_tube(p::CartesianPoint, r_torus::Real, r_tube::Real) = hypot(hypot(p.x, p.y) - r_torus, p.z) <= r_tube
-@inline _in_torr_r_tube(p::CartesianPoint, r_torus::Real, r_tube::AbstractInterval) = hypot(hypot(p.x, p.y) - r_torus, p.z) in r_tube
+@inline _in_torr_r_tube(p::CartesianPoint, r_torus::Real, r_tube::Real, z_torus::Real) = hypot(hypot(p.x, p.y) - r_torus, p.z - z_torus) <= r_tube
+@inline _in_torr_r_tube(p::CartesianPoint, r_torus::Real, r_tube::AbstractInterval, z_torus::Real) = hypot(hypot(p.x, p.y) - r_torus, p.z - z_torus) in r_tube
 
-@inline _in_torr_θ(p::CartesianPoint{T}, r_torus::Real, θ::AbstractInterval) where {T} = mod(atan(p.z, hypot(p.x, p.y) - r_torus), T(2π)) in θ
+@inline _in_torr_θ(p::CartesianPoint{T}, r_torus::Real, θ::AbstractInterval, z_torus::Real) where {T} = _in_angular_interval_closed(mod(atan(p.z - z_torus, hypot(p.x, p.y) - r_torus), T(2π)), θ)
```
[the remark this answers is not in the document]
hervasa2: Similar to your `_in_φ` if `θ = nothing` then the functions is not called

src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -98,11 +98,17 @@ end
...
+@inline _eq_torr_r_tube(p::CylindricalPoint{T}, r_torus::Real, r_tube::Real, z_torus::Real) where {T} = isapprox(hypot(p.r - r_torus, p.z - z_torus), T(r_tube), atol = geom_atol_zero(T))
+@inline _in_torr_r_tube(p::CylindricalPoint, r_torus::Real, r_tube::Real, z_torus::Real) = hypot(p.r - r_torus, p.z - z_torus) <= r_tube
+@inline _in_torr_r_tube(p::CylindricalPoint, r_torus::Real, r_tube::AbstractInterval, z_torus::Real) = hypot(p.r - r_torus, p.z - z_torus) in r_tube
+
+@inline _in_torr_θ(p::CylindricalPoint{T}, r_torus::Real, θ::AbstractInterval, z_torus::Real) where {T} = _in_angular_interval_closed(mod(atan(p.z - z_torus, p.r - r_torus), T(2π)), θ)
+
+function Δ_φ(φ1::T, φ2::T)::T where {T}
+    δφ = mod(φ2 - φ1, T(2π))
+    min(δφ, T(2π) - δφ)
+end
```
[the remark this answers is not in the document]
hervasa2: Will be used by `distance_to_surface`

src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
```diff
@@ -21,20 +21,20 @@ function ConeMantle(t::Torus{T}; θ = π/2) where {T}
...
+        zMax = geom_round(t.z + r_tubeMax*sθ)
     elseif θ > T(π) && θ < T(2π)
-        rtop = t.r_torus + r_tubeMin*cθ
-        rbot = t.r_torus + r_tubeMax*cθ
-        zMax = r_tubeMin*sθ
-        zMin = r_tubeMax*sθ
+        rtop = geom_round(t.r_torus + r_tubeMin*cθ)
+        rbot = geom_round(t.r_torus + r_tubeMax*cθ)
+        zMax = geom_round(t.z + r_tubeMin*sθ)
+        zMin = geom_round(t.z + r_tubeMax*sθ)
```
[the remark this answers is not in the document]
hervasa2: Because of the cosines and sines. They have given me trouble with surface interactions without the geom_round

src/ConstructiveSolidGeometry/Intervals.jl
hervasa2: This is more human readable. Also I believe in your example the case of `else return true` is not covered by `a && !(b && c)`

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
hervasa2: Its to check if one of the dimensions of the cone is zero. For example if the cone volume collapsed into a cone mantle or cylindrical annulus. The term collapsed is commonly used for these 3D to 2D transformations

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
hervasa2: This also happens in many other places. Like for TR. And then for the same fields in surface primitives. Should we keep these parametric types as short as possible or tie the order and number of parameters to the number of fields in the struct?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
hervasa2: It's the mechanical term for a rounded corner. I was using it as a shortcut. Should we delete?

src/ConstructiveSolidGeometry/Intervals.jl
fhagemann: True! Then, let's leave it as is

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
fhagemann: That's a good question! @lmh91 what's your opinion on this?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
fhagemann: Let us leave it. But if it is a user-convenient function, we should export it.

src/ConstructiveSolidGeometry/Intervals.jl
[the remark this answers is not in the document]
fhagemann: For clarification, we could call this function `union_closed_angular_intervals`, such that if we might run into cases where we use open intervals, we know that we might need to adjust some of this.

src/ConstructiveSolidGeometry/Intervals.jl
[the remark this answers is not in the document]
fhagemann: This is breaking a bit what we have for the volume primitives:
For the volume primitives, an "interval" of `nothing` means full `2π`.
Here, `nothing` describes an empty interval. 
Can we avoid this inconsistency somehow?

Also, I only see this method is only called once in your code, checking whether the output is `nothing` (in `is_intersection_an_interval`), where you could also just call `isempty(α ∩ β)`. Or am I missing something?

Is this function important for the future because I do not see the need for it now?

src/ConstructiveSolidGeometry/Intervals.jl
```diff
@@ -15,4 +15,80 @@
...
+        return true
+    elseif isnothing(union_angular_intervals(α, β))
```
fhagemann, line 86: replace with `isempty(α ∩ β)` to avoid unnecessary calculations?

hervasa2: `isempty(α ∩ β)` does not work for periodic intervals

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
lmh91: If it is really **always** of type `T` get rid of it. 

Parametric types are nice but should be kept as small as possible.

test/ConstructiveSolidGeometry/CSG_test.jl
```diff
@@ -22,13 +25,123 @@ using SolidStateDetectors.ConstructiveSolidGeometry:
...
+                end
+                @testset "Negative Interval" begin
+                    neg_interval = -T(π/2)..T(π/2)
+                    @test _in_angular_interval_closed(T(0), neg_interval)
+                    @test _in_angular_interval_closed(T(2π), neg_interval)
+                    @test !_in_angular_interval_closed(T(π), neg_interval)
+                    @test !_in_angular_interval_closed(T(3π), neg_interval)
+                    @test !_in_angular_interval_closed(T(-π), neg_interval)
+                    @test _in_angular_interval_closed(T(-π/2), neg_interval)
+                    #@test _in_angular_interval_closed(T(-3π/2), neg_interval)
```
fhagemann, line 67: I guess these are the cases where it fails for `Float32`?

test/ConstructiveSolidGeometry/CSG_test.jl
```diff
@@ -22,13 +25,123 @@ using SolidStateDetectors.ConstructiveSolidGeometry:
...
+                @testset "CylindricalAnnulus" begin
+                    cannulus = CylindricalAnnulus(T(2.0), T(0.0))
+                    @test CartesianPoint{T}(0, 0, 0) in cannulus
+                    @test CartesianPoint{T}(2, 0, 0) in cannulus
+                    @test CartesianPoint{T}(0, -2, 0) in cannulus
+                    @test !(CartesianPoint{T}(2, 2, 0) in cannulus)
+                    @test !(CartesianPoint{T}(0, 0, 0.1) in cannulus)
+                end
+                @testset "ToroidalAnnulus" begin
+                    tannulus = ToroidalAnnulus()
```
fhagemann, line 102: Do we want `ToroidalAnnulus()` to have default values when nothing is passed?

Project.toml
```diff
@@ -5,7 +5,6 @@ version = "0.5.1"
 [deps]
 ArraysOfArrays = "65a8f2f4-9b39-5baf-92e2-a9cc46fdf018"
 Clustering = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
-Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
```
fhagemann, line 8: Very good!

fhagemann: You will need to also remove `using Colors` from `src/SolidStateDetectors.jl`

test/ConstructiveSolidGeometry/CSG_test.jl
hervasa2: Exactly!

test/ConstructiveSolidGeometry/CSG_test.jl
hervasa2: Ive made it so that all surface primitives have default values just like volume primitives. These are basically just convenience functions for fast testing

test/ConstructiveSolidGeometry/CSG_test.jl
fhagemann: But these are all values smaller than `0` or bigger than `2π`. I would say we can live with this for now and have a look at these cases once they might become crucial!

test/ConstructiveSolidGeometry/CSG_test.jl
fhagemann: Can you please still add the numbers (either in the function call or as a comment thereafter), so that one can easily see what the dimensions of the tested objects are without having to look for the default values in the source code?

fhagemann (approved):
We might want to change the primitive definition such that the initial primitives are **ALWAYS** centred around 0 (#84) and then translated in whatever direction. We could, then, remove:
- the `z` field for `Torus`
- the possibility that `z` of a `Cone` is an `AbstractInterval`.

But if that is needed, it will be part of a follow-up PR.
For now, this is good! :heart:

lmh91:
I though about this a while already and I am actually in favor of removing this `z` field in the primitives. 
I know it is useful for certain cases (especially LEGEND related),
but it is just against the general concept of how we apply the transformations.

In case we find out that we really need that I would also rather add not only the `z` component but 
the total origin (a vector) and also the direction (also a vector) so we can apply any transformation on them. 

But this we can change/add later as Felix said. 

Feel free to merge @fhagemann

PR #146, opened 2021-04-30
CSG: Redefinition of HexagonalPrism by RegularPrism

fhagemann:
`HexagonalPrism{T}` will be renamed `RegularPrism{T}` which allows for all possible polygons as a base.
There is an alias `HexagonalPrism{T} = RegularPrism{6,T}` which allows parsing from the config file exactly like before.

With this comes a breaking change:
Two of the corners of the `HexagonalPrism` used to lay on the y-axis (`φ = π/2`). Now, the first point of the polygon base will be aligned with (`φ = 0`). The old and new implementation will differ by `π/2`, which for the `HexagonalPrism` is equivalent to `π/6`.
I adjusted the tests for `HexagonalPrism`.

We could define `RegularPrism` to have one of its corners aligned with the y-axis. However, this would result in an addition of `π/2` every time `in()` is called. Any comments on this?

@hervasa2 will take care of the respective `SurfacePrimitive`: `RegularPolygon` and `RegularPrismMantle`.


fhagemann:
This PR provides:
- the definition of `RegularPrism`
- aliases `TriangularPrism`, `QuadraticPrism`, `PentagonalPrism` and `HexagonalPrism`
- constructors, equal to the one for `HexagonalPrism`
- `in` methods for `CylindricalPoint`
- `ìn` method for `CartesianPoint` with a special analytical treatment of `HexagonalPrism`
- `sample` function for grid creation
- plot recipe in `plotting/VolumePrimitives` (although it is not read in)
- updated tests due to redefinition of the locations of the corners of `HexagonalPrism`

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
```diff
@@ -0,0 +1,90 @@
...
+function sample(rp::RegularPrism{N,T}, Nsamps::NTuple{3,Int} = (2,N,2))::Vector{CylindricalPoint{T}} where {N,T}
+    rMin::T, rMax::T = get_r_limits(rp)
+    zMin::T, zMax::T = get_z_limits(rp)
+    samples = [
+        CylindricalPoint{T}(r,φ,z)
+        for z in (Nsamps[3] ≤ 1 ? zMin : range(zMin, zMax, length = Nsamps[3]))
+        for φ in 0:2π/N:2π-π/N #ignore the N given in Nsamps and sample with N
+        for r in (Nsamps[1] ≤ 1 ? _right_radial_interval(rp.r) : range(rMin, rMax, length = Nsamps[1]))
+    ]
+end
```
lmh91, line 90: This looks to me like sampling a cylinder, or am I wrong?
I know we check the sample points afterwards if they are inside but this just to substract differences of primitives.

fhagemann: You are right. However, this is intended:
`sample` functions for `VolumePrimitives` are used to get the important points for grid creation.
In fact, sampling a `RegularPrism` looks like sampling a `Tube`, just that we have a **fixed** step size of `2π/N`, i.e. all the corners of the `RegularPrism` are sampled.


HexagonalPrism
Tube


For contact painting, we will use the `sample` methods for the respective `SurfacePrimitives` which are not yet implemented.

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
```diff
@@ -0,0 +1,90 @@
...
+get_z_limits(rp::RegularPrism) = (_left_linear_interval(rp.z), _right_linear_interval(rp.z))
+
+function sample(rp::RegularPrism{N,T}, Nsamps::NTuple{3,Int} = (2,N,2))::Vector{CylindricalPoint{T}} where {N,T}
+    rMin::T, rMax::T = get_r_limits(rp)
+    zMin::T, zMax::T = get_z_limits(rp)
+    samples = [
+        CylindricalPoint{T}(r,φ,z)
+        for z in (Nsamps[3] ≤ 1 ? zMin : range(zMin, zMax, length = Nsamps[3]))
+        for φ in 0:2π/N:2π-π/N #ignore the N given in Nsamps and sample with N
+        for r in (Nsamps[1] ≤ 1 ? _right_radial_interval(rp.r) : range(rMin, rMax, length = Nsamps[1]))
```
hervasa2, line 88: Isn't `_right_radial_interval` just `rMax`?

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
hervasa2: However this will break and sample a cylinder with `Nsamps = (2,x,2)` with `x > N`. I know sample is used to get the important points of the primitive, but it should also be able to sample the actual volume

hervasa2 (changes requested):
Very Nice! This adds so many primitives at once 🕺 . Please see note about sampling

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
```diff
@@ -0,0 +1,90 @@
...
+RegularPrism(N::Integer, rInner, rOuter, zMin, zMax) = RegularPrism(N; rInner = rInner, rOuter = rOuter, zMin = zMin, zMax = zMax)
+
+function RegularPrism(N::Integer, rOuter::R, height::H) where {R<:Real, H<:Real}
+    T = float(promote_type(R,H))
+    RegularPrism( N, T, T(rOuter), T(height/2))
+end
+
+
+@inline in(p::CylindricalPoint, hp::RegularPrism{N, T, <:Real}) where {N,T} = begin
+    _in_z(p, hp.z) && abs(p.r * cos(T(π/N) - mod(p.φ, T(2π/N))) / cos(T(π/N))) <= hp.r
```
hervasa2, line 48: `abs` not needed

hervasa2 (changes requested):

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
hervasa2: I suggest using `Nsamps::NTuple{3,Int} = (2,N+1,2)` and `for φ in (Nsamps[2] ≤ 1 ? 0 : range(0, 2π, length = Nsamps[2]))` with `CylindricalPoint{T}(r*cos(π/N)/cos(π/N - mod(φ, 2π/N)),φ,z)`

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
```diff
@@ -0,0 +1,90 @@
...
+                             z::Union{T, <:AbstractInterval{T}}) where {T}
+        @assert typeof(N) <: Integer && N >= 3 "Failed constructing RegularPrism with $(N) points in the polygon base (needs at least 3 points)."
+        new{N, T,typeof(r),typeof(z)}(r,z)
+    end
+end
+
+
+# Convenience functions
+const TriangularPrism{T,TR,TZ} = RegularPrism{3,T,TR,TZ}
+const QuadraticPrism{T,TR,TZ}  = RegularPrism{4,T,TR,TZ}
```
hervasa2, line 17: I know this naming is consistent with the rest but I suggest we change it to SquarePrism to match the name more commonly used. I am calling `RegularPolygon{4,T}` a `Square{T}`

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
fhagemann: I like this! Sampling with your code works now as intended:
irregular_sampling

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
fhagemann: We should somehow include a warning if `p.r` is negative, though (probably when creating the `RegularPrism` object). 
Otherwise, this would always be `true`..

hervasa2 (approved):

PR #147, opened 2021-04-30
Sphere Mantle and plotting

hervasa2:
Added SphereMantle Surface primitive, thus completing the Sphere Volume Primitive.
`get_decomposed_surfaces` is now implemented for the Sphere Volume Primitive
Includes basic functions and plotting
Minor general plotting updates
Screen Shot 2021-04-30 at 7 23 35 PM


src/ConstructiveSolidGeometry/PointsAndVectors.jl
```diff
@@ -35,6 +35,8 @@ end
 @inline _in_z(p::CartesianPoint, z::Real) = abs(p.z) <= z
 @inline _in_z(p::CartesianPoint, z::AbstractInterval) = p.z in z
 
+@inline _isapprox_sph_r(p::CartesianPoint{T}, radius::Real) where {T} = isapprox(hypot(p.x, p.y, p.z), radius, atol = geom_atol_zero(T))
```
fhagemann, line 38: Can you check whether `norm(p)` is faster than `hypot`?

src/ConstructiveSolidGeometry/SurfacePrimitives/SphereMantle.jl
```diff
@@ -0,0 +1,23 @@
...
+function sample(s::SphereMantle{T}, step::Real)::Vector{CylindricalPoint{T}} where {T}
+    samples = [
+        CylindricalPoint{T}(sqrt(s.r^2 - z^2),φ,z)
+        for z in in -s.r:step:s.r
+        for φ in 0:step/s.r:2π
+    ]
+end
+=#
+
+function sample(s::SphereMantle{T}, Nsamps::NTuple{3,Int})::Vector{CylindricalPoint{T}} where {T}
```
fhagemann, line 17: We would also need `sample` methods that work with `CartesianTuple` and `CylindricalTuple`.

fhagemann (commented):
Do we know why one test is failing?

src/ConstructiveSolidGeometry/SurfacePrimitives/SphereMantle.jl
hervasa2: I left them for you since you are familiar with what you need for the grid. Could you tackle these please? These sample methods are also missing from the Toroidal Surface Primitives

src/ConstructiveSolidGeometry/PointsAndVectors.jl
hervasa2: Just did, difference is negligible

src/ConstructiveSolidGeometry/SurfacePrimitives/SphereMantle.jl
fhagemann: Yes, I can take care of this.

fhagemann:
I addded the `sample` functions for the `SphereMantle`:

#### CartesianTicksTuple

SphereMantle_Cartesian

#### CylindricalTicksTuple

SphereMantle_Cylindrical

fhagemann:
What I figured out while implementing the `sample` method for the `SphereMantle` and running the tests for our example config file with a spherical detector:
- The painting algorithm cannot (yet) handle `CSGDifference` in the painting algorithm, so its paints the whole `SphereMantle` as contact
- If no `grid` is passed in the config file (as is the case in `public_spherical_detector_config.json`), then the world dimensions are determined from the objects of a `SolidStateDetector`. The function `get_world_limits_from_objects`, however, still depends on `get_important_points` and needs an update.

fhagemann:
Sampling the surface primitives of a `Torus`  works now for `CylindricalTicksTuple`:
Torus_Cylindrical

Config files with `Torus` can now be read in.
All of this allows testing simulations based on `Torus` primitives in a cylindrical coordinate system in `runtests.jl`.

fhagemann:
Sampling the surfaces of a `Torus` with `CartesianTicksTuple` should now work, even though the code might need some glow-up.

#### Sampling with `CylindricalTicksTuple`
Torus_Cylindrical

#### Sampling with `CartesianTicksTuple`
Torus_Cartesian

src/ConstructiveSolidGeometry/VolumePrimitives/Sphere.jl
```diff
@@ -31,13 +30,19 @@ end
 
 get_r_limits(s::Sphere{T}) where {T} = (_left_radial_interval(s.r), _right_radial_interval(s.r))
 
+get_decomposed_surfaces(s::Sphere{T, T}) where {T} = AbstractSurfacePrimitive[SphereMantle{T}(s.r)]
+
+function get_decomposed_surfaces(s::Sphere{T, <:AbstractInterval{T}}) where {T}
+    rMin::T, rMax::T = get_r_limits(s)
+    AbstractSurfacePrimitive[SphereMantle{T}(rMin), SphereMantle{T}(rMax)]
```
lmh91, line 37: Can this be a tuple?
At least remove the `AbstractSurfacePrimitive` in front of the vector.

lmh91: Same at line 33

lmh91 (changes requested):
Looks fine but the one minor thing with the array of an abstract type

src/ConstructiveSolidGeometry/VolumePrimitives/Sphere.jl
hervasa2: This is for consistency. `get_decomposed_surfaces` always returns an `AbstractSurfacePrimitive` vector for all volume primitives. The Sphere is just special in the since that the only type of decomposed surfaces it has are `SphereMantle`

lmh91: This creates type instabilities.

I know we merge these arrays so they become arrays of multiple types (what we have to address at some point),
but here we should definitely not return an array of abstract element type here.

fhagemann: We save decomposed surfaces to `Semiconductor`, `Contact`, and `Passive`. Does it make sense to store them as `Tuple` instead of as `Vector` (to save the type information)?

hervasa2: `get_decomposed_surfaces` is never type stable. I think eventually we will need a better fix, but for now this is what I have been working with

hervasa2: I suggest we leave this PR as is and resolve the issue as a whole in a separate PR. It would be a change to all `get_decomposed_surfaces` of every Volume Primitive

fhagemann: That sounds reasonable.

lmh91 (approved):

src/ConstructiveSolidGeometry/VolumePrimitives/Sphere.jl
lmh91: Well, the method `get_decomposed_surfaces(s::Sphere{T, <:AbstractInterval{T}})` clearly always returns two `SphereMantle{T}` so this is always type stable. 

But if this also the case for the other primitives right now we can leave it like it is for now and clean this up later together with the other primitives. Just removing the `AbstractSurfacePrimitive` in front of the `[` should not break anything anyhow.

hervasa2: This happens in other volume primitives too, not like here where this is always the case but in specific scenarios.

PR #149, opened 2021-05-04
Hotfix for signal generation

lmh91:
For some reason the generated drifts could have two equal time stamps
what causes later an issues in the signal generation.

This hotfix just ends the drift if that happens.

On the current v0.6 branch the charge drift is currently refactored anyhow.

### DIFF
--- a/src/ChargeDrift/ChargeDrift.jl
+++ b/src/ChargeDrift/ChargeDrift.jl
@@ -143,7 +143,11 @@ function _drift_charge!(
                         next_pos -= small_projected_vector
                         i += 1
                     end
-                    if i == 1000 && verbose @warn("Handling of charge at floating boundary did not work as intended. Start Position (Cart): $startpos") end
+                    if i == 1000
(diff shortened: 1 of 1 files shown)


fhagemann:
Let me have a look at what causes this problem.

fhagemann:
The duplicate entries result from the drift code implementation for floating boundaries in ChargeDrift.jl:
```julia
while i < 1000 && !(next_pos in det)
    next_pos -= small_projected_vector
    i += 1
end
if i == 1000 && verbose @warn("Handling of charge at floating boundary did not work as intended. Start Position (Cart): $startpos") end
drift_path[istep] = next_pos
drifttime += Δt * (1 - i * T(0.001))
timestamps[istep] = drifttime
if geom_round.(next_pos - current_pos) == null_step
    done = true
end
```
If `i == 1000`, then the drifttime is increased by `0`, resulting in a double entry of the same timestamp in `timestamps`.

fhagemann:
Closed by accident. Sorry..

fhagemann (approved):

lmh91:
Thanks @fhagemann  for looking more into it!

PR #153, opened 2021-05-05
Surface sampling for RegularPrism surface primitives

fhagemann:
This PR changes two things:
- sampling of surface primitives of `RegularPrism` with `CartesianTicksTuple` (a bit tricky, but works nicely now)
- if `CylindricalPoint` is only translated along `z`: skip the conversion to `CartesianPoint`.

Also:
We might want to redefine the CSG implementation of `RegularPrism` such that one of the centre of one of the edges is on the x-axis, rather than one of the corners. That way, `HexagonalPrism` would have the same definition as in the old SSD and `SquarePrism` would be aligned with `x` and `y` instead of being rotated by 45°. We might, then, get a `Box` for free.


fhagemann:
Some obligatory GIFs, so you can see what is sampled (shown in gray is the "Grid"):

#### CylindricalTicksTuple
RegularPrism_Cylindrical


#### CartesianTicksTuple
RegularPrism_Cartesian

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -85,6 +85,14 @@ end
     pt_car in point_types || pt_car in det
 end
 
+function project_to_plane(v⃗::AbstractArray, n⃗::AbstractArray) #Vector to be projected, #normal vector of plane
```
fhagemann, line 88: This function used to be in the old `src/Geometries` folder and is needed for the old drift code implementation.
However, this will not be needed in the future, when the drift code is refactored.

src/ConstructiveSolidGeometry/SurfacePrimitives/RegularPolygon.jl
```diff
@@ -62,6 +62,33 @@ end
...
+    for i in eachindex(corners)[1:end-1]
+        s1,c1 = r .* corners[i]
+        s2,c2 = r .* corners[i+1]
+        if ! isapprox(s1, s2, atol = geom_atol_zero(T))
+            t = (y-s1)/(s2-s1)
+            if t in 0..1 push!(x, c1 + t * (c2-c1)) end
+        end
+    end
+    x
+end
```
fhagemann, line 89: Is there an elegant way to write this in Array syntax like we do it for sampling primitives?

src/ConstructiveSolidGeometry/SurfacePrimitives/RegularPolygon.jl
```diff
@@ -71,10 +98,29 @@ function sample(rp::RegularPolygon{N,T}, Nsamps::NTuple{3,Int} = (2,N+1,2))::Vec
...
         CylindricalPoint{T}(r,φ,rp.z)
         for φ in _get_ticks(g.φ, T(0), T(2π))
         for r in _get_ticks(g.r, get_r_at_φ(rp,φ)...)
     ]
 end
+
+function sample(rp::RegularPolygon{N,T}, g::CartesianTicksTuple{T})::Vector{CartesianPoint{T}} where {N,T}
+    corners = SVector{N+1,Tuple{T,T}}(sincos.(T(2π/N) .* (0:N)))
+    rMin::T, rMax::T = get_r_limits(rp)
+    samples = vcat(
```
fhagemann, line 113: Will `vcat` lead to poor performance (and if so, how can we avoid it)?
I checked: sampling a `RegularPolygon` or `RegularPrismMantle` takes approximately as long as sampling a `ConeMantle`

fhagemann:
This PR would (based on functionality) be complete. See my comments for possible performance questions.

The question that remains: should we redefine the `RegularPrism` according to the initial comment of this PR or leave it as is?

fhagemann:
And now is also the time to get rid of the old implementation of geometries :tada:

fhagemann:
There was some undersampling of the inner edges of `RegularPolygon` with a cutout (thanks to @hervasa2 for pointing that out)

#### Cylindrical sampling of `RegularPolygon` with cutout
Cutout_Cylindrical

#### Cartesian sampling of `RegularPolygon` with cutout
Cutout_Cartesian

fhagemann:
An explanation of how a `RegularPrismMantle` or the edges of a `RegularPolygon` are sampled with `CartesianTicksTuple`:

In the `x`-`y`-plane, all corner points of the `RegularPolygon{N}` with radius `R` are given by `(Rcos(2πk/N), Rsin(2πk/N))`.
Thus, when determining the intersection of a given `y` with the edges of the `RegularPolygon`, we can define intervals in `y` given by two neighbouring corner points (shown by arrows in the picture below).

If the given `y` value falls into one of the intervals (shown in red), the respective edge and the horizontal `y` line intersect in a point which can be determined by linear interpolation. The two `x` values from the intersection points determine the minimum and maximum value for which points are generated according to the `x` values in the `CartesianTicksTuple`.

Sampling

src/ConstructiveSolidGeometry/SurfacePrimitives/RegularPolygon.jl
hervasa2: From my checks, it looks like it's the same speed as `append!`

hervasa2 (approved):
Awesome! Thanks for the explanation, it really helped, cartesian ticks is quite tricky

PR #155, opened 2021-05-08
Update config file format

fhagemann:
This pull request will fix #136 by updating the config file format.

### Remove `type` keyword from primitive dictionaries (9dc9abe)

For example, a `Tube` used to be defined like this:
```yaml
type: "tube"
r:
  from: 1.0
  to: 2.0
phi:
  from: 0.0
  to: 360.0
h: 2.0
```
and is now this:
```yaml
tube:
  r:
    from: 1.0
    to: 2.0
  phi:
    from: 0.0
    to: 360.0
  h: 2.0
```

### Remove `parts` keyword from `AbstractConstructiveGeometry` (9dc9abe)

For example, a `CSGUnion` used to be defined like this:
```yaml
type: "union"
  parts:
    - type: "tube"
      r:
        from: 15.0
        to: 35.0
      phi:
        from: 0.0
        to: 360.0
      h: 0.0
    - type: "tube"
      r:
        from: 35.0
        to: 35.0
      phi:
        from: 0.0
        to: 360.0
      h: 20.0
      translate:
        z: 10.0
```
and is now this:
```yaml
union:
  - tube:
      r:
        from: 15.0
        to: 35.0
      phi:
        from: 0.0
        to: 360.0
      h: 0.0
  - translate:
      tube:
        r:
          from: 35.0
          to: 35.0
        phi:
          from: 0.0
          to: 360.0
        h: 20.0
      z: 10.0
```

### Introduce parsing of `ScaledGeometry` and `RotatedGeometry` (9dc9abe)

The parsing of `ScaledGeometry` is quite simple: define with `scale` and write the value for the Cartesian coordinate (`x`, `y` or `z`) that should be scaled by the factor given
```yaml
scale:
  tube:
    r: 
     from: 0.1
     to: 0.5
    phi:
     from: 0.0
     to: 360.0
    h: 1.0
  x: 2.0
```

`RotatedGeometry` can be defined either by passing a Matrix `M` as a flat array or an axis and an angle to turn around.
Example: `XZ: [60.0,30.0]` will be parsed as `RotXZ(60.0,30.0)`.
Throws an error if `M` is not a `RotationMatrix`.
```yaml
rotate:
  tube:
    r: 
     from: 0.1
     to: 0.5
    phi:
     from: 0.0
     to: 360.0
    h: 1.0
  M: [1,0,0,0,0,-1,0,1,0] # or Z: 90.0 => M = RotZ(90.0)
```

[...]

PR #156, opened 2021-05-13
Rectangle Surface Primitive and Integration with Box

hervasa2:
Definition and plotting for Rectangle Surface Primitive. Rectangles are restricted to have normal vectors on `x`, `y`, or `z`

- Functions: `in()`, `sample()`
- Pending: `distance_to_surface()`. To be added along with that of prism surfaces. 
- Pending: `sample()` with ticks

Box now has `get_decomposed_surfaces()`. It gets decomposed into two of each: `Rectangle{T, Val{:x}}`, `Rectangle{T, Val{:y}}`, `Rectangle{T, Val{:z}}` unless one of the dimensions is zero.

Screen Shot 2021-05-13 at 5 33 58 PM
Screen Shot 2021-05-13 at 5 33 42 PM
Screen Shot 2021-05-13 at 5 33 24 PM
Screen Shot 2021-05-13 at 5 32 27 PM

### DIFF
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
@@ -84,6 +84,91 @@ function sample(r::RectangleZ{T}, Nsamps::NTuple{3,Int} = (2,2,2)) where {T}
     ]
 end
 
+function get_r(r::Rectangle{T, Val{:x}}, φ::T) where {T}
+    atol = geom_atol_zero(T)
+    x::T = r.loc
+    yMin::T, yMax::T = get_l_limits(r)
+    sφ::T, cφ::T = sincos(φ)
+    if x * cφ < 0 return () end
+    tanφ::T = tan(φ)
+    yMin - atol ≤ x * tanφ ≤ yMax + atol ? (x / cφ,) : ()
+end
+
+function get_r(r::Rectangle{T, Val{:y}}, φ::T) where {T}
+    atol = geom_atol_zero(T)
+    y::T = r.loc
+    xMin::T, xMax::T = get_l_limits(r)
+    sφ::T, cφ::T = sincos(φ)
+    if y * sφ < 0 return () end
+    cotφ::T = cot(φ)
+    xMin - atol ≤ y * cotφ ≤ xMax + atol ? (y / sφ,) : ()
+end
+
+function sample(r::Union{Rectangle{T, Val{:x}}, Rectangle{T, Val{:y}}}, g::CylindricalTicksTuple{T}) where {T}
+    samples = [
+        CylindricalPoint{T}(R, φ, z)
+        for φ in g.φ
+        for R in get_r(r,φ)
+        for z in _get_ticks(g.z, get_w_limits(r)...)
+    ]
+end
+
+function get_r_ticks(r::Rectangle{T, Val{:z}}, g::CylindricalTicksTuple{T}, φ::T)::Vector{T} where {T}
+    xMin::T, xMax::T = get_l_limits(r)
+    yMin::T, yMax::T = get_w_limits(r)
+    sφ::T, cφ::T = sincos(φ)
+    tanφ::T = tan(φ)
+    cotφ::T = inv(tanφ)
+    R = sort!(filter!(R -> R > 0,
+        [
+            (xMin ≤ yMin * cotφ ≤ xMax ? yMin / sφ : T(NaN)),
+            (xMin ≤ yMax * cotφ ≤ xMax ? yMax / sφ : T(NaN)),
+            (yMin ≤ xMin * tanφ ≤ yMax ? xMin / cφ : T(NaN)),
+            (yMin ≤ xMax * tanφ ≤ yMax ? xMax / cφ : T(NaN))
+        ]
+    ))
+    ticks = if length(R) == 2
+        _get_ticks(g.r, R...)
+    elseif length(R) == 1
+        _get_ticks(g.r, T(0), R[1])
+    else
+        []
+    end
+end
+
+function sample(r::Rectangle{T, Val{:z}}, g::CylindricalTicksTuple{T}) where {T}
+    samples = [
+        CylindricalPoint{T}(R, φ, z)
+        for z in _get_ticks(g.z, r.loc, r.loc)
+        for φ in g.φ
+        for R in get_r_ticks(r, g, φ)
+    ]
+end
+
+function sample(r::Rectangle{T, Val{:x}}, g::CartesianTicksTuple{T}) where {T}
+    samples = [
+        CartesianPoint{T}(r.loc,y,z)
+        for y in _get_ticks(g.y, get_l_limits(r)...)
+        for z in _get_ticks(g.z, get_w_limits(r)...)
+    ]
+end
+
+function sample(r::Rectangle{T, Val{:y}}, g::CartesianTicksTuple{T}) where {T}
+    samples = [
+        CartesianPoint{T}(x,r.loc,z)
+        for x in _get_ticks(g.x, get_l_limits(r)...)
+        for z in _get_ticks(g.z, get_w_limits(r)...)
+    ]
+end
+
+function sample(r::Rectangle{T, Val{:z}}, g::CartesianTicksTuple{T}) where {T}
+    samples = [
+        CartesianPoint{T}(x,y,r.loc)
+        for x in _get_ticks(g.x, get_l_limits(r)...)
+        for y in _get_ticks(g.y, get_w_limits(r)...)
+    ]
+end
+
 function get_vertices(r::RectangleX{T}) where {T}
     lMin::T, lMax::T = get_l_limits(r)
     wMin::T, wMax::T = get_w_limits(r)
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
@@ -48,7 +48,7 @@ end
 end
 
 @inline in(p::AbstractCoordinatePoint, r::RectangleZ) = begin
-    _in_x(p, r.l) && _in_y(p, r.w) && _eq_z(p, r.loc)
+    _in_x(p, r.l) && _in_y(p, r.w) && _isapprox_z(p, r.loc)
 end    
 
 get_l_limits(r::Rectangle) = (_left_linear_interval(r.l), _right_linear_interval(r.l))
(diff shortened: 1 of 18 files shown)


src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
```diff
@@ -0,0 +1,104 @@
+struct Rectangle{T,TN,TL,TW} <: AbstractSurfacePrimitive{T}
+    normal::TN
+    l::TL
+    w::TW
```
fhagemann, line 4: `l` will be the coordinate that comes alphabetically before `w`,
i.e. `l ≡ x, w ≡ y`, `l ≡ x, w ≡ z` or `l ≡ y, w ≡ z`?

src/ConstructiveSolidGeometry/plotting/SurfacePrimitives/Rectangle.jl
```diff
@@ -0,0 +1,9 @@
+function get_plot_points(r::Rectangle{T}; n = 30) where {T <: AbstractFloat}
+    v = get_vertices(r)
+    [Vector{CartesianPoint{T}}([v[1], v[2]]), Vector{CartesianPoint{T}}([v[2], v[4]]), Vector{CartesianPoint{T}}([v[3], v[4]]), Vector{CartesianPoint{T}}([v[3], v[1]])]
+end
+
+function mesh(r::Rectangle{T}; n = 30) where {T <: AbstractFloat}
+    vertices = [get_vertices(r)...]
+    Mesh(reshape(map(p -> p.x, vertices), (2,2)),reshape(map(p -> p.y, vertices), (2,2)),reshape(map(p -> p.z, vertices), (2,2)))
```
fhagemann, line 8: Is it faster here to replace `map` with `broadcast`?

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
```diff
@@ -0,0 +1,104 @@
+struct Rectangle{T,TN,TL,TW} <: AbstractSurfacePrimitive{T}
```
fhagemann, line 1: How about changing the order to `Rectangle{TN, T, TL, TW}`.
This would allow to easily introduce aliases like `RectangleX{T,TL,TW} = Rectangle{Val(:x),T,TL,TW}` which might be more convenient for the user (see `RegularPrism`).

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
```diff
@@ -0,0 +1,104 @@
...
+    _in_y(p, r.l) && _in_z(p, r.w) && _isapprox_x(p, r.loc)
+end
+
+@inline in(p::AbstractCoordinatePoint, r::Rectangle{<:Any, Val{:y}}) = begin
+    _in_x(p, r.l) && _in_z(p, r.w) && _isapprox_y(p, r.loc)
+end
+
+@inline in(p::AbstractCoordinatePoint, r::Rectangle{<:Any, Val{:z}}) = begin
+    _in_x(p, r.l) && _in_y(p, r.w) && _eq_z(p, r.loc)
+end
```
fhagemann, line 44: By introducing aliases, this could then be 
`@inline in(p::AbstractCoordinatePoint, r::RectangleZ)`

fhagemann (commented):
Nice! :+1: 

Some comments:
- I find the surface primitives `Square` from `RegularPrism` and `Rectangle` from `Box`.
  People might think that a square `Rectangle` might be a `Square`, so we should discuss whether we want to rename `Square` then.
- For the `Rectangle` constructors: what do you think about introducing aliases?
  (e.g. `RectangleX = Rectangle{<:Any, Val(:x)}` because I do not think it is the user-friendliest option to use `Val(:x)` as input for `Rectangle`. Also, it would allow getting rid of most of the `Val(:x)` in the code.

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
fhagemann: Also, we do not need to save `TN` as a field to the struct, but just leave it as a parametric type, right?

fhagemann:
I added the `sample` functions with `CartesianTicksTuple` and `CylindricalTicksTuple`.

The sampling with `CartesianTicksTuple` is extremely straightforward.
The `Rectangles` are sampled at their edges and all the grid ticks in between:
Sample_Cartesian
Sample_offset_Cartesian


The sampling with `CylindricalTicksTuple` is a bit more tricky.
For each `φ`, the intersections with the four edges of the `Rectangle` at positive radii are determined. This can be zero (no intersection), one (the `Rectangle` includes the origin of the coordinate system) or two (the `Rectangle` has an offset and does not include the origin of the coordinate system). Then, all ticks in `r` between the intersection points are sampled with the grid spacing as step size:
Sample_Cylindrical
Sample_offset_Cylindrical

test/comparison_to_analytic_solutions.jl
```diff
@@ -24,7 +24,7 @@ e = SolidStateDetectors.elementary_charge * u"C"
         @test isapprox(C_ssd, C_true, rtol = 0.001) 
     end
 end
-=#
+
```
fhagemann, line 27: I have no idea why this test fails. It passes locally on my machine...

fhagemann: All of the tests calculate `C_ssd` to be 6.3529pF, while the true value `C_true` is 14.1667pF.
On my laptop, `C_ssd` has another value closer to `C_true` than the ones calculated here...
Box_test

lmh91: Hmm, strange. 
It also run rough on my system in an empty Julia environment...

fhagemann: It looks like both simulations are using different initial grids and grid sizes (?)
and the grid is refined in `y` and `z` on GitHub, while it is not locally...

Locally:

Grid_size_local

GitHub:

Grid_size_GitHub

fhagemann: On this branch, the config file format is not yet updated (#155).
Could it be that merging it to the `v0.6` branch will result in a wrong parsing of the config file somehow?

fhagemann: This is indeed the source of the problem. I will have a look into it and try to fix it.

fhagemann: Let's see whether merging #158 will resolve this problem.

fhagemann: If I merge the `JuliaPhysics:v0.6` branch into `hervasa2:v0.6`, the tests do not fail anymore.
I don't know while the tests here still fail. But the problem should have been fixed by the PR 158.

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
```diff
@@ -43,6 +43,26 @@ get_x_limits(b::Box{T}) where {T} = (_left_linear_interval(b.x), _right_linear_i
...
+    zMin::T, zMax::T = get_z_limits(b)
+    tol = geom_atol_zero(T)
+    if isapprox(xMin, xMax, atol = tol)
+        return AbstractSurfacePrimitive[Rectangle(b, Val(:x), xMin)]
+    elseif isapprox(yMin, yMax, atol = tol)
+        return AbstractSurfacePrimitive[Rectangle(b, Val(:y), yMin)]
+    elseif isapprox(zMin, zMax, atol = tol)
+        return AbstractSurfacePrimitive[Rectangle(b, Val(:z), zMin)]
+    else
+        return AbstractSurfacePrimitive[
```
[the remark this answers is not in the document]
hervasa2: We've had this conversation before. This also happens with the Sphere and in some other specific cases. The `AbstractSurfacePrimitive` is there for consistency. This is still a pending issue that deserves its own PR

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
hervasa2: Yes!
- RectangleX: `l ≡ y, w ≡ z`
- RectangleY: `l ≡ x, w ≡ z`
- RectangleZ: `l ≡ x, w ≡ y`

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
```diff
@@ -0,0 +1,197 @@
...
+    _in_y(p, r.l) && _in_z(p, r.w) && _isapprox_x(p, r.loc)
+end
+
+@inline in(p::AbstractCoordinatePoint, r::RectangleY) = begin
+    _in_x(p, r.l) && _in_z(p, r.w) && _isapprox_y(p, r.loc)
+end
+
+@inline in(p::AbstractCoordinatePoint, r::RectangleZ) = begin
+    _in_x(p, r.l) && _in_y(p, r.w) && _eq_z(p, r.loc)
+end
```
fhagemann, line 52: Is there a reason why for `x` and `y` we use `_is_approx` while for `z` we use `_eq`?

test/ConstructiveSolidGeometry/CSG_test.jl
```diff
@@ -177,6 +177,17 @@ using SolidStateDetectors.ConstructiveSolidGeometry:
                     @test !(CartesianPoint{T}(1, 0, -0.1) in prism)
                     @test !(CartesianPoint{T}(1, 0, 1.1) in prism)
                 end
+                @testset "Rectangle" begin
+                    rectangle = RectangleZ(T(1.0), T(2.0), T(3.0)) # x from -0.5..0.5, y from -1.0..1.0, z = 3
```
fhagemann, line 181: If we use `_is_approx` for `RectangleX` and `RectangleY` and `_eq_z` for `RectangleZ`, I would also suggest testing both here. We might also think about incorporating `CylindricalPoints` in general in the tests.

fhagemann:
I would like the sampling of `Rectangle` to also sample intersections of `r` ticks with the edges and not only the intersection of the `φ` ticks. This would result in more samples on the blue lines in this figure:

> Sample_offset_Cylindrical

Let me work on this before we merge.

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
hervasa2: I did not consider rotations. I will add `_is_approx_z`

fhagemann:
Sampling with `CylindricalTicksTuple` now also includes sample points at `r`-ticks on the edges:

### Sampling only at `r` ticks 
sample_at_r_ticks

### Combined sampling
Sample_offset_Cylindrical_updated

The corners of a `Rectangle` are not explicitly sampled. However, when creating the `Grid`, the "important points" of the `Rectangle` will be added such that the corners will "naturally" be sampled at a specific `r` and `φ` tick.

PR #159, opened 2021-05-17
Distance to Surface for Remaining Surface Primitives

hervasa2:
`distance_to_surface()` added to RegularPolygon, RegularPrismMantle, and Rectangle
All are type stable, allocation free and have a worst case btime of ~50ns
`rp = CSG.RegularPolygon(7,0.8,1.3,0)`
Screen Shot 2021-05-17 at 8 30 51 PM
distance_to_regular_polygon

`rpm = CSG.RegularPrismMantle(3,1.3,1)`
Screen Shot 2021-05-17 at 8 31 54 PM
distance_to_regular_prism_mantle
distance_to_regular_prism_mantle2

`rec = CSG.RectangleX(0.5, 1, -1.2, 0.6, 0)`
Screen Shot 2021-05-17 at 8 33 53 PM
distance_to_rectangle
distance_to_rectangle2

`rec = CSG.RectangleZ(0.5, 1, -1.2, 0.6, 0.5)`
Screen Shot 2021-05-17 at 8 35 23 PM
distance_to_rectangle3

### DIFF
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
@@ -238,7 +238,7 @@ function distance_to_surface(point::AbstractCoordinatePoint{T}, r::Rectangle{<:A
     l, w, d = _get_rectangle_coordinates(point, r)
     lMin::T, lMax::T = get_l_limits(r)
     wMin::T, wMax::T = get_w_limits(r)
-    if l in lMin..lMax && w in wMin..wMax
+    if lMin ≤ l ≤ lMax && wMin ≤ w ≤ wMax
         return abs(d - r.loc)
     else 
         if lMin ≤ l ≤ lMax
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
@@ -227,3 +227,31 @@ function get_vertices(r::RectangleZ{T}) where {T}
     CartesianPoint{T}(lMin, wMax, r.loc),
     CartesianPoint{T}(lMax, wMax, r.loc))
 end
+
+_get_rectangle_coordinates(point::CartesianPoint, r::RectangleX) = (point.y, point.z, point.x)
+_get_rectangle_coordinates(point::CartesianPoint, r::RectangleY) = (point.x, point.z, point.y)
+_get_rectangle_coordinates(point::CartesianPoint, r::RectangleZ) = (point.x, point.y, point.z)
+
+_get_rectangle_coordinates(point::CylindricalPoint, r::Rectangle) = _get_rectangle_coordinates(CartesianPoint(point), r)
+
+function distance_to_surface(point::AbstractCoordinatePoint{T}, r::Rectangle{<:Any, T})::T where {T}
+    l, w, d = _get_rectangle_coordinates(point, r)
+    lMin::T, lMax::T = get_l_limits(r)
+    wMin::T, wMax::T = get_w_limits(r)
+    if l in lMin..lMax && w in wMin..wMax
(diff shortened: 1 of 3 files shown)


src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
```diff
@@ -226,4 +226,32 @@ function get_vertices(r::RectangleZ{T}) where {T}
...
+_get_rectangle_coordinates(point::CartesianPoint, r::RectangleY) = (point.x, point.z, point.y)
+_get_rectangle_coordinates(point::CartesianPoint, r::RectangleZ) = (point.x, point.y, point.z)
+
+_get_rectangle_coordinates(point::CylindricalPoint, r::Rectangle) = _get_rectangle_coordinates(CartesianPoint(point), r)
+
+function distance_to_surface(point::AbstractCoordinatePoint{T}, r::Rectangle{<:Any, T})::T where {T}
+    l, w, d = _get_rectangle_coordinates(point, r)
+    lMin::T, lMax::T = get_l_limits(r)
+    wMin::T, wMax::T = get_w_limits(r)
+    if l in lMin..lMax && w in wMin..wMax
```
fhagemann, line 241: Can we use existing internal functions (similar to `_in_planar_u`)?
If `r.l` and `r.w` are `Intervals`, we would determine the edges and create a "new" interval from them.

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
```diff
@@ -226,4 +226,32 @@ function get_vertices(r::RectangleZ{T}) where {T}
...
+        if lMin ≤ l ≤ lMax
+            y = w < wMin ? wMin : wMax
+            return hypot(w - y, d - r.loc)
+        elseif wMin ≤ w ≤ wMax
+            x = l < lMin ? lMin : lMax
+            return hypot(l - x, d - r.loc)
+        else
+            lnear = l < lMin ? lMin : lMax
+            wnear = w < wMin ? wMin : wMax
+            return hypot(l - lnear, w - wnear, d - r.loc)
```
fhagemann, line 253: How about `lnear = min(max(l, lMin), lMax)` and `wnear = min(max(w, wMin), wMax)` ?
This should cover all cases as `lnear = min(max(l, lMin), lMax)` is:
- `lMin` if `l < lMin`
- `l` if `lMin < l < lMax`
- `lMax` if `l > lMax`

The result in all cases would still be `hypot(l - lnear, w - wnear, d - r.loc)`?
as `l - lnear` will be `0` in the case `lMin < l < lMax`.

Lines 241 - 254 could then become:
```julia
return hypot(l - min(max(l, lMin), lMax), w - min(max(w, wMin), wMax), d - r.loc)
```

Not tested for performance.

fhagemann:
> `rp = CSG.RegularPolygon(7,0.8,1.3,0)`
> Screen Shot 2021-05-17 at 8 30 51 PM
> distance_to_regular_polygon

Is this the same object? The cutout of the `Polygon` looks bigger in the GIF than in the figure...

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
hervasa2: This for sure would work and its shorter. But it I explicitly wrote it the way it is to minimize number of operations

src/ConstructiveSolidGeometry/SurfacePrimitives/Rectangle.jl
hervasa2: 3 options. 

- `_in_planar_u` would work but then you have to create PlanarPoints
- use multiple dispatch. This would be a lot of unnecessary code
- leave as is. Creates new intervals. 

All have drawbacks.

fhagemann (approved):
All good!

PR #160, opened 2021-05-17
Add tests for CSG sampling

fhagemann:
In order to extend the CSG sampling with `CSGUnion`, `CSGIntersection` and `CSGDifference` using `in()`,
tests are provided to see that all sampled points are "in" the primitive they were sampled from.

For most primitives, the sampled points are "in" the geometry they were sampled from. The only (mostly rounding) errors occur for:
- [x] `ConeMantle` with `CartesianTicksTuple`
- [x] `RegularPolygon` with `CartesianTicksTuple`
- [x] `RegularPrismMantle` with `CartesianTicksTuple`
- [x] `RectangleZ` with `CylindricalTicksTuple`

which, for now, are excluded from the tests. I will check if the `sample` / `in` methods can be changed such that these tests pass.

### DIFF
--- a/src/ConstructiveSolidGeometry/PointsAndVectors.jl
+++ b/src/ConstructiveSolidGeometry/PointsAndVectors.jl
@@ -108,18 +108,18 @@ end
 
 @inline _isapprox_x(p::CylindricalPoint{T}, x::Real) where {T} = isapprox(p.r * cos(p.φ), x, atol = geom_atol_zero(T))
 
-@inline _in_x(p::CylindricalPoint, x::Real) = abs(p.r * cos(p.φ)) <= x
-@inline _in_x(p::CylindricalPoint, x::AbstractInterval) = p.r * cos(p.φ) in x
+@inline _in_x(p::CylindricalPoint{T}, x::Real) where {T} = abs(p.r * cos(p.φ)) <= x + geom_atol_zero(T)
+@inline _in_x(p::CylindricalPoint, x::AbstractInterval) = p.r * cos(p.φ) in x || _isapprox_x(p, x.left) || _isapprox_x(p, x.right)
 
 @inline _isapprox_y(p::CylindricalPoint{T}, y::Real) where {T} = isapprox(p.r * sin(p.φ), y, atol = geom_atol_zero(T))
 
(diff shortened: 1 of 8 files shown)


fhagemann:
I added some tolerances (`geom_atol_zero(T)`) when checking if values/points are inside a given range/interval, which now seems to fix the problems in Julia v1.6. However, the tests in Julia v1.0 still fail (apparently, `hypot` was calculated differently before..)
How should we proceed here?

Also: if we extend all intervals and ranges by `geom_atol_zero(T)` (which is `1e-6` for `Float32`), we might run into problems for very thin objects. Effectively, all objects are expanded by `1e-6`, which equals to 1µm, if necessary. That might also cause trouble.

schustermartin:
Would it work if, for a given comparison, we compare the value itself as well as `prevfloat()` and `nextfloat()`, and if either is true, then it passes? That should also work for very thin objects and as I remember for most cases its a matter of the very next or previous float when checks fail where they shouldn't.

fhagemann:
I can try this out. Thanks for the input

fhagemann:
Simply using `prevfloat` and `nextfloat` does not seem to fix it because there are multiple operations.

Example: To check if a `CartesianPoint` is in a `RegularPolygon{N}`, SSD does the following:
- Converts the point into a `CylindricalPoint`
`@inline in(p::CartesianPoint, rp::RegularPolygon) = in(CylindricalPoint(p), rp)`
- Reduces the problem in `φ` to the interval from `0` to `2π/N`
  `mod(p.φ, T(2π/N))`
- And checks whether the radius expected at that reduced `φ` is in the radius defined in the `RegularPolygon`:
  `p.r * cos(T(π/N) - mod(p.φ, T(2π/N))) / cos(T(π/N)) <= rp.r` (or `... in rp.r` if it is an interval)

Thus, `p.r * cos(T(π/N) - mod(p.φ, T(2π/N))) / cos(T(π/N))` might be several `prevfloat` or `nextfloat` away from `rp.r`.
There is a possibility of going multiple `nextfloat`, i.e. `nextfloat(rp.r, 3)` gives the third float after `rp.r` but then we would need to define tolerances for each case and I am not sure how to test them all.

fhagemann:
Closing this in favour of #163, where sampling with `TicksTuple` is no longer needed.

Issue #136, opened 2021-02-12
Update config file format


fhagemann:
Based on the clarification of impurity and charge densities (#65), the redefinition of primitives (#84), the implementation of transformations (#62) and the refactoring of the CSG (#110), we need to update the config file format.

For this, we should decide on a long-term supported format, such that the user will have to update the config file format once (ideally with the help of conversion functions that SSD will provide).

lmh91:
At the end, so after the changes you mentioned, we should also have a look regarding the simulation of multiple detectors. 
This not only concerns the config file but maybe also `mutable struct Simulation`. 
Just to keep this in mind.

fhagemann:
Also, parameters of the charge drift model should be part of the general config file (preferably with splitted config files), rather than being stored separately in a folder that no one will find.

fhagemann:
I gathered all comments in PR regarding the new structure of config files to gather them here:

### PR #122 in a [comment from @lmh91](https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/122#issuecomment-739755931) about the general definition of `VolumePrimitives`
Instead of 
```yaml
type: "tube"
r: 
  from: 0.1
  to: 0.5
phi: 
  from: 0.0
  to: 360.0
h: 1.0
```
define the whole thing as `tube` instead of having a `type: "tube"`` inside the definition:
```yaml
tube:
  r: 
   from: 0.1
   to: 0.5
  phi:
   from: 0.0
   to: 360.0
  h: 1.0
```

### PR #84 in a [comment from @fhagemann](https://github.com/JuliaPhysics/SolidStateDetectors.jl/issues/84#issuecomment-758534887) about the general definition of `CSGTransformations`

Instead of making transformations part of the objects, i.e.
```yaml
type: "tube"
r: 
  from: 0.1
  to: 0.5
h: 1.0
translate:
  z: 1.0
```
give the object the name of the transformation and pass it the values needed for its definition, i.e.
```yaml
translate:
  z: 1.0  # defines the translation
  tube:   # primitive that is translated
    r: 
      from: 0.1
      to: 0.5
    h: 1.0
```
### PR #84 in a [comment from @lmh91](https://github.com/JuliaPhysics/SolidStateDetectors.jl/issues/84#issuecomment-758569474) on the general definition of `CSGRotations`

Rotations should be either defined by angles around specific axes or as a general rotation matrix:
```yaml
[...]

PR #161, opened 2021-05-23
Extend ADLChargeDriftModel to arbitrary crystal orientations

### LINKED ISSUE #67: Parse crystal orientation to drift model

fhagemann:
This PR closes #67.

I have:
- merged the changes of the `master` branch into `v0.6`  (7a9bac2), especially to merge the changes from #144.
- replaced the number `phi110` in `ADLChargeDriftModel` by a rotation matrix `crystal_orientation` (8881997) that describes the rotation of the coordinate system of the crystal. The initial crystal orientation (no rotation) is chosen such that the `<100>`, `<010>` and `<001>` axes are aligned with the `x`, `y` and `z` axes.
- added tests for `ADLChargeDriftModel` (a73503a) for which the `<100>`, `<011>` and `<111>` axes are in the `xy`-plane.

Find some plots attached in the `z` cross-section of a true coaxial detector with the following layout for drift paths and axes:
Legend

Initially, `phi110` was part of the `ADLChargeDriftModel` and described the angle between `<110>` and `x` axes.
This way of parsing the crystal orientation will still be possible:

### Coaxial detector, "raw", corresponds to `phi110 = 45°` (was like this before)
Coax_raw

### Coaxial detector, `phi110 = -15°` (was like this before)
Coax_phi110_-15

### Coaxial detector, `X = 45°` (new; this is the case that is added to the tests)
Coax_X45


src/ChargeDriftModels/ADL/ADL.jl
```diff
@@ -116,18 +151,40 @@ function ADLChargeDriftModel(configfilename::Union{Missing, AbstractString} = mi
...
-    ismissing(phi110) ? phi110 = T(config["phi110"]) : phi110 = T(phi110)  #if you give the angle of the 110 axis it will be used, otherwise read from config file
+    crystal_orientation::SMatrix{3,3,T,9} = if ismissing(phi110) 
+        if "phi110" in keys(config)
+            RotZ{T}(- π/4 - config["phi110"] )
+        else
+            transpose(parse_rotation_matrix(T, config, u"rad")) # replace u"rad" with the units in the config file
+        end
+    else
+        RotZ{T}(- π/4 - phi110)
+    end
```
fhagemann, line 185: This is where `crystal_orientation`  gets parsed.
`"X": x` in the config file will rotate the crystal axes by `x` radians around the `X` axis.

Right now, the value `x` is **always** parsed in units of radians (see L. 181).
This should be changed when updating the config file format (#136) and making the `charge_drift_model` a part of a `Semiconductor`. Then, the user-defined units in the config file should be chosen.

src/ChargeDriftModels/ADL/ADL.jl
```diff
@@ -116,18 +151,40 @@ function ADLChargeDriftModel(configfilename::Union{Missing, AbstractString} = mi
...
+        mt = T(material_properties[Symbol(material)].mt)
+        ADLParameters{T}(ml, mt, material)
     end
 
-    ismissing(phi110) ? phi110 = T(config["phi110"]) : phi110 = T(phi110)  #if you give the angle of the 110 axis it will be used, otherwise read from config file
+    crystal_orientation::SMatrix{3,3,T,9} = if ismissing(phi110) 
+        if "phi110" in keys(config)
+            RotZ{T}(- π/4 - config["phi110"] )
+        else
+            transpose(parse_rotation_matrix(T, config, u"rad")) # replace u"rad" with the units in the config file
```
fhagemann, line 181: The rotation matrix is transposed here. This is because the rotation of the coordinate system is a passive rotation.

In active rotations, an object is rotated around an axis. In passive rotations, the coordinate system is rotated such that all stationary objects are effectively rotated in the opposite direction with respect to the coordinate system.

However, transposing the matrix results in the fact that if the coordinate system should be first rotated by an angle `x` around the `X` axis and subsequently by an angle `z` around the `Z` axis, the corresponding implementation would have to look like this:
```yaml
"ZX": [z, x],
```
which is not intuitive. For this, we would need proper documentation.

lmh91 (approved):
Looks fine to me! :+1:

PR #162, opened 2021-05-28
ADLChargeDriftModel config files: Relocation and incorporation into detector config files

fhagemann:
In this PR, the config files for the `ADLChargeDriftModel` are moved to `examples/example_config_files` and can be read in using `"include"` in our splitted config file syntax.

## Changes

The changes in this PR are:
- Renaming `example_detector_config_files` to `example_config_files` (0e15ac9), as now also `ADLChargeDriftModel` config files will be stored here (dbecc33) and read in using the splitted config file syntax (5b26a5d).
- Adding the key `"model"` to the `ADLChargeDriftModel` config file (d69776a), so that the config file parser knows which charge drift model to use.
- Removing the field `charge_drift_model` from the `Simulation` struct and adding it to `Semiconductor` (42c3d4b).
- Instead of allowing multiple `Semiconductors` in a `SolidStateDetector`, only one `Semiconductor` per `SolidStateDetector` is allowed (3a50e1c). Also, the config files are adjusted to only expect **exactly one** `Semiconductor` (3a50e1c).

## Points to discuss
- [x] Where exactly we want to store information in the config files, especially `material`, `phi110` and `temperature`.
- [x] Decide on a long-term stable config file format for the next release.

## Follow up PRs
- Allow multiple `SolidStateDetectors` per `Simulation` (however still only one `Semiconductor` per `SolidStateDetector`).
- Proper creation of the world with `get_world_limits_from_objects` (the current implementation still uses `get_important_points` which does not exist anymore...)


fhagemann:
I would need some feedback on this:

In the current status of this PR, detector config files look like this:
```yaml
name: "Detector Name"
units: ... # units to parse config file entries
grid: ... # grid dimensions and boundary conditions
medium: ... # medium surrounding the detector

objects:
  semiconductor: # only one Semiconductor is allowed, so "semiconductor" has exactly one entry
    material: "HPGe"
    impurity_density: ... # impurity density
    charge_drift_model: ... # charge drift model
    geometry: ... # geometry
  contacts: # multiple Contacts are allowed, so this is a Vector{Contact}
    - material: "HPGe"
      channel: 1
      potential: 5000.0
      geometry: ... # geometry
    - material: "HPGe"
      channel: 2
      potential: 0.0
      geometry: ... # geometry
```

Multiple detectors could be defined by making a `Vector` of `detectors`, consisting of a `name` and `objects`, each:
```yaml
name:  Do we even need a name for the whole simulation?
units: ... # units to parse config file entries
grid: ... # grid dimensions and boundary conditions
medium: ... # medium surrounding the detector

detectors: # This would be an Array of entries, each corresponding to one detector
  - name: "Detector Name 1" # first detector
    objects:
      semiconductor:
        material: "HPGe"
        impurity_density: ... # impurity density
        charge_drift_model: ... # charge drift model
        geometry: ... # geometry
      contacts:
        - material: "HPGe"
          channel: 1
          potential: 5000.0
          geometry: ... # geometry
        - material: "HPGe"
          channel: 2
          potential: 0.0
          geometry: ... # geometry
 - name: "Detector Name 2" # second detector
   objects:
      semiconductor:
        material: "HPGe"
        impurity_density: ... # impurity density
        charge_drift_model: ... # charge drift model
        geometry: ... # geometry
      contacts:
        - material: "HPGe"
          channel: 1
          potential: 5000.0
          geometry: ... # geometry
        - material: "HPGe"
          channel: 2
          potential: 0.0
          geometry: ... # geometry
```
This leaves some open questions:
- Do we want to give detectors more attributes other than `name` and `objects`? Something like `detno`?
- Where do we save information about `temperature`, `material`, and `crystal_orientation`? Right now, information is split over objects and some are saved in the detector config files while others are saved in the config files for the `ADLChargeDriftModel`. My proposal would be:
  - Keep `material` as it is, as part of `Semiconductor` and `Contact`. Right now, it is also saved in `charge_drift_model`, but I think it can be removed and one could check for the material of the `Semiconductor`, when setting the charge drift model.
  - Right now, the `temperature` is saved in the charge drift model, as it is only needed there. However, I would propose to add it to either `detector` or `semiconductor` because there are charge drift models that do not need it at all. If charge drift models are added afterwards, the information might be missing/overwritten. By default, I would propose 78K, which is the temperature at which the charge drift parameters in germanium were determined.
  - Add `crystal_orientation` to the `Semiconductor` which the same argument as `temperature`: even if it is only used by the charge drift model, it is still a semiconductor property.

lmh91:
I thought about allowing multiple semiconductors per detector. (I know I said that we only need one.) 
But would it be possible to add this feature later one?
I mean, that it can either be as it is right now, just a single one, or one can define a vector of it? 
This could be added in the future without violating the old (current) syntax, right? Just a conceptional question. 

I would leave the `name` field for the whole simulation. In addition of the names of individual detectors. 
I don't think we need something like `detno`. The user will most likely use the name field for this. 

I would add the `temperature` directly to the detector, actually maybe even the whole simulation. 
The idea is that this would be passed recursive to all subobjects. And one could overwrite it (later, not now)
by adding it to also subobjects. (If we want/need different temperatures at different places).

`crystal_orientation` I would add to the (each) semiconductor.

`Material` should be also in semiconductor and contacts. 

Each detector should have his own charge drift model. 

While writing this I got the impression that recursive parsing of properties would be quite nice in general.
E.g. If a detector gets the `material: HPGe`, than all subobjects, semiconductors, contacts & passives, would automatically be set to HPGe. But, each one could be overwritten individually. 
Also for other field, not just material. 
This might reduce the config file size a bit. What do you think?

Do we want to add a version-number field? We talked about this at some point I believe.

fhagemann:
I agree with most of the points, especially the recursive parsing, except for `crystal_orientation`, which for me is a semiconductor property, so I would put it there.

To sum up what still needs to be done:
- Make `semiconductors` a Vector (like it was before) and, for now, allow only one entry when parsing the config file in v0.6? 
  This would allow defining more than one in future releases without having to change the syntax.
- General stuff concerning all of the objects will be defined in the `detector` (`temperature`, `material`) and recursively passed to the respective objects.
- Add a version number field (with no version number field meaning that it is still in the old format of v0.5.2 or older?).

This would result in the following example config file:
```yaml
name: "Simulation Name"
version: "0.6"
units: ... # units to parse config file entries
grid: ... # grid dimensions and boundary conditions
medium: ... # medium surrounding the detector

detectors: # This would be an Array of entries, each corresponding to one detector
  - name: "Detector Name 1" # first detector
    temperature: 78
    material: "HPGe"
    objects:
      semiconductors: # will now be an Array, but in the source code check that only one entry is present (?)
        - impurity_density: ... # impurity density
          crystal_orientation: ...
          charge_drift_model: ... # charge drift model
          geometry: ... # geometry
          # material and temperature will be taken from detector
      contacts:
        - material: "Si" # overwrites "HPGe" from detector with "Si"
          channel: 1
          potential: 5000.0
          geometry: ... # geometry
        - material: "HPGe"
          channel: 2
          potential: 0.0
          geometry: ... # geometry
 - name: "Detector Name 2" # second detector
   temperature: 78
   objects:
      semiconductors:
        - impurity_density: ... # impurity density
          crystal_orientation: ...
          charge_drift_model: ...
          geometry: ... # geometry
          temperature: 80 # can also be defined here
          material: "HPGe" # can also be defined here
      contacts:
        - material: "HPGe"
          channel: 1
          potential: 5000.0
          geometry: ... # geometry
        - material: "HPGe"
          channel: 2
          potential: 0.0
          geometry: ... # geometry
```

I just realized: do we even need `objects:`? We could just define a detector as:
```yaml
detectors: # This would be an Array of entries, each corresponding to one detector
  - name: "Detector Name 1" # first detector
    temperature: 78
    material: "HPGe"
    semiconductors: # will now be an Array, but in the source code check that only one entry is present (?)
      - impurity_density: ... # impurity density
        crystal_orientation: ...
        charge_drift_model: ... # charge drift model
        geometry: ... # geometry
        # material and temperature will be taken from detector
    contacts:
      - material: "Si" # overwrites "HPGe" from detector with "Si"
        channel: 1
        potential: 5000.0
        geometry: ... # geometry
      - material: "HPGe"
        channel: 2
        potential: 0.0
        geometry: ... # geometry
```

What do you think?

fhagemann:
I agree also with your point, that we should agree on the **format of the config file** which should be easily extendable in future releases but without breaking the existing syntax.  What happens in the source code of SSD and where the information will be stored in the different `structs` during the simulation; that is another question. As long as it works for now.

fhagemann:
What about deleting the `semiconductor` field completely and adding all bulk properties to `detector`?
This would reduce the number of entries and simplify the implementation of allocation-free `in(point, detector)`.

```yaml
name: "Simulation Name"
SSD_version: "0.6"
units: ... # units to parse config file entries
grid: ... # grid dimensions and boundary conditions
medium: ... # medium surrounding the detector
temperature: ... # with the option of overwriting it at lower level

detectors: # This would be an Array of entries, each corresponding to one detector
  - name: "Detector Name 1" # first detector
   
    # properties of the bulk
    temperature: 78 # optional: would overwrite the world temperature
    material: "HPGe"
    impurity_density: ... # impurity density
    crystal_orientation: ...
    charge_drift_model: ... # charge drift model
    geometry: ... # geometry

    # additional objects
    contacts:
      - material: "Si" # overwrites "HPGe" from detector with "Si"
        channel: 1
        potential: 5000.0
        geometry: ... # geometry
      - material: "HPGe"
        channel: 2
        potential: 0.0
        geometry: ... # geometry
   passivations: # optional
     - material: "aGe"
       charge_density: ... #optional
       geometry: ... # geometry
   virtual_volumes: # optional
     - model: ... # how the drift will be manipulated
       id: 1 
       geometry: ... # geometry
```

oschulz:
Groups could be realized like this, with the above scheme:

```yaml
groups:
   - name: "group1"
     detectors:
     - name: "det1"
       #...
     - name: "det2"
       # ...
   groups:
   - name: "module1"
     detectors:
     - name: "det1"
       # ...
```

fhagemann:
Let's stick to this:

```yaml
name: "Simulation Name"
SSD_version: "0.6"
units: ... # units to parse config file entries
grid: ... # grid dimensions and boundary conditions
medium: ... # medium surrounding the detector
temperature: ... # with the option of overwriting it at lower level

detectors: # This would be an Array of entries, each corresponding to one detector
  - name: "Detector Name 1" # first detector
    bulk: # properties of the bulk (each detector should consist of only one bulk)
      temperature: 78 # optional: would overwrite the world temperature
      material: "HPGe"
      impurity_density: ... # impurity density
      crystal_orientation: ...
      charge_drift_model: ... # charge drift model
      geometry: ... # geometry
    contacts:
      - material: "Ge(Li)"
        channel: 1
        potential: 5000.0
        geometry: ... # geometry
      - material: "Ge(B)"
        channel: 2
        potential: 0.0
        geometry: ... # geometry
   passivations: # optional
     - material: "aGe"
       charge_density: ... #optional
       geometry: ... # geometry
   virtual_volumes: # optional
     - model: ... # how the drift will be manipulated
       id: 1 
       geometry: ... # geometry
   translate: ... # optional: a translate vector that translates ALL detector objects
   rotate: ... # optional: a ŕotation matrix that rotates ALL detector objects

surroundings: # all objects in the vicinity of the detector(s)
  - name: "Cryostat"
    material: "Al"
    potential: 0
    geometry: ... # geometry
```

The config file will change according to the following points:
- Bulk properties will be saved in a field called `bulk`.
- The detectors should have optional fields (`translate` and `rotate`) to translate/rotate **ALL** detector objects at once
  (Such that the translation does not need to be added to all primitives if the whole detector needs to be moved)


Definitions of translations and rotations should be definable in a nested way or as a field of a primitive. 
  Example for a `Tube` translated `1.0` along `z`:
 ```yaml
translate:
  tube:
    r: ...
    z: ...
  z: 1.0
``` 
should be parsed the same as
 ```yaml
tube:
  r: ...
  z: ...
  translate:
    z: 1.0
```
If both, `rotate` and `translate` are defined within a primitive, we will first rotate and then translate.

fhagemann:
Find an overview of where the information is saved after the last commits below:

|  <div style="width:290px">SSD v0.5</div>    |  <div style="width:290px">SSD v0.6</div>       |
| ------------- | ------------- | 
| `Simulation{T}`<ul><li>detector</li><li>ρ</li><li>ρ_fix</li><li>ϵ</li><li>point_types</li><li>electric_potential</li><li>weighting_potentials</li><li>electric_field</li><li>**charge_drift_model** (42c3d4b)</li><li>electron_drift_field</li><li>hole_drift_field</li><li></li><li></li><li></li></ul> | `Simulation{T,CS}` <ul><li>**config_dict** (75c4af8)</li><li>**input_units** (75c4af8)</li><li>**medium** (75c4af8)</li><li>detector (--> detectors)</li><li>**world** (bf97b91)</li><li>q_eff_imp</li><li>q_eff_fix</li><li>ϵ_r</li><li>point_types</li><li>electric_potential</li><li>weighting_potentials</li><li>electric_field</li><li>electron_drift_field</li><li>hole_drift_field</li></ul>|
| `SolidStateDetector{T,CS}`<ul><li>name</li><li>**input_units** (75c4af8)</li><li>**world** (bf97b91)</li><li>**config_dict** (75c4af8)</li><li>**medium** (75c4af8)</li><li>_semiconductors_ (d29208c)</li><li>contacts</li><li>passives</li><li>virtual_drift_volumes</li></ul> | `SolidStateDetector{T}` (cc3f964) <ul><li>name</li><li>_semiconductor_ (d29208c)</li><li>contacts</li><li>passives</li><li>virtual_drift_volumes</li><li></li><li></li><li></li><li></li></ul> |
| `Semiconductor{T}` <ul><li>name</li><li>id</li><li>temperature</li><li>material</li><li>charge_density_model</li><li>geometry</li><li>geometry_positive</li><li>geometry_negative</li><li></li><li></li><li></li></ul> | `Semiconductor{T}` <ul><li>name</li><li>id</li><li>temperature</li><li>material</li><li>impurity_density_model</li><li>**charge_drift_model** (42c3d4b)</li><li>geometry</li><li>geometry_positive</li><li>geometry_negative</li><li>decomposed_surfaces</li></ul> |

fhagemann:
The detectors now have optional fields (translate and rotate) to translate/rotate ALL detector objects at once (8269292).
Note that sampling points of rotated objects  with `CartesianTicksTuple` or `CylindricalTicksTuple` still needs some updates, so it will not perfectly work yet.

fhagemann:
> Definitions of translations and rotations should be definable in a nested way or as a field of a primitive.
> Example for a `Tube` translated `1.0` along `z`:
> 
> ```yaml
> translate:
>  tube:
>    r: ...
>    z: ...
>  z: 1.0
> ```
> 
> should be parsed the same as
> 
> ```yaml
> tube:
>  r: ...
>  z: ...
>  translate:
>    z: 1.0
> ```
> 
> If both, `rotate` and `translate` are defined within a primitive, we will first rotate and then translate.

This should work now (5fab6ac)

fhagemann:
I would like to keep this PR as is and implement the simulation of multiple detectors in a single config file in a follow-up PR.

fhagemann:
We are good to go!

lmh91 (approved):
Thanks! :)

oschulz (commented):
Good from my side!

PR #165, opened 2021-06-30
Update Documentation

lmh91:


fhagemann:
If we want to drop support for Julia 1.0, we couĺd also incorporate lines 30-33 back:
https://github.com/lmh91/SolidStateDetectors.jl/blob/docs/src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl#L30

which will replace `SolidStateDetectors.ConstructiveSolidGeometry.Cartesian` with only `Cartesian`, whenever it is printed.
The same goes for `Cylindrical`. This resulted in errors in Julia 1.0

fhagemann:
Also, is this PR mostly about the tutorial or also about docstrings?

lmh91:
It is only about the tutorial in the documentation. 
Not it runs through again. Some changes were needed since `SolidStateDetector` is not mutable anymore. 

But the plot recipes will need some further adjustments / improvements. 
Docstrings can always be improved/added.

lmh91:
Regarding the Julia 1.0 comment: 
I would do the drop of v1.0 in a dedicated PR.

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -20,11 +20,23 @@ function Geometry(::Type{T}, ::Type{Ellipsoid}, dict::AbstractDict, input_units:
...
-        r = r, 
-        φ = φ, 
-        θ = θ
-    )
+    e = if r isa Tuple{T,T}
+        Ellipsoid{T,ClosedPrimitive,T,typeof(φ),typeof(θ)}(
+            r = r[2], 
+            φ = φ, 
+            θ = θ
+        ) - Ellipsoid{T,ClosedPrimitive,T,typeof(φ),typeof(θ)}(
```
fhagemann, line 28: Shouldn't this be an `OpenPrimitive`? Or do we want the inner radius excluded?
Also, treating a `Sphere` with an inner radius `r1 != 0` results in failing tests.

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -1,37 +1,24 @@
...
-    linecolor --> :black
+
+@recipe function f(csg::AbstractConstructiveGeometry)
+    ps = primitives(csg)
     @series begin
-        linestyle --> :solid
-        csguniontype --> :difference
-        csg.a
+        label --> "CSG"
+        linestyle := isClosedPrimitive(ps[1]) ? :solid : :dot
```
fhagemann, line 13: So this means, that `OpenPrimitives` are always the negative geometry of a `CSGDifference`?

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
lmh91: Ah correct, that should be an OpenPrimitive. 
Yes, the IO is not perfect yet. Also for other primitives. 
That is something I want to talk about but not tackle in this PR.

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
lmh91: No. Differences of Differences switches Open an Closed and so on...
There is no ideal solution for plotting this. This is just a simple way to have a look at the CSG.
But for sure nothing for a publication.

PR #166, opened 2021-07-02
Add config file formats for `Box`

lmh91:
Add some ways to define a box primitive:
```yaml
box:
   widths: [2,2,2] # or 

   # halfwidths: [1, 1, 1] # or

   # hX: 1
   # hY: 1
   # hZ: 1 # or 

   # x: 1 
   # y: 
   #   from: -1
   #   to: 1
   # z: 1

  origin: [0, 0, 0] # optional
  rotate: 
    Z: 0 # optional
```
Origin can also be specified like
```yaml
origin: 
  x: 0
  y: 0
  z: 0 # or

origin: 
  x: 1 # -> origin at (1, 0, 0)
```
Also add functions `get_origin` and `get_rotation` which read in the optional fields `origin` and `rotate`.
These will also be used in the read-in of the other primitives (next PR's).

The rotation is applied before the translation to the origin.


fhagemann:
Thanks for using commit messages that well describe what you are doing ;)
Silence

PR #168, opened 2021-07-05
Add `origin` and `rotation` to IO of Primitives

lmh91:


fhagemann:
For the config files: right now we are using `origin` and `translate` interchangably. Can we find a consistent way of naming this?

lmh91:
I would use `origin` only inside the definition of primitives. `origin` is the correct term here in my opinion. 
Outside of it I would suggest to use `translate`, because this is an `transformation`/`operation` performed on the respective geometry. 

For rotations we use `rotation` in both cases right now, what is kinda inconsistent. 
I would suggest to use `rotate` outside and `rotation` only inside of the primitive definitions.

We could allow both though as I don't see any conflict potential and it might be more user friendly.

fhagemann:
> We could allow both though as I don't see any conflict potential and it might be more user friendly.

I would prefer this because it might confuse people that definitions of translations or rotations are different depending on whether they are inside or outside the primitive definition.

lmh91:
> > We could allow both though as I don't see any conflict potential and it might be more user friendly.
> 
> I would prefer this because it might confuse people that definitions of translations or rotations are different depending on whether they are inside or outside the primitive definition.

Okay, now it is possible to use `origin`, `translate` or `translation` for the `origin` of a primitive and 
`rotation` or `rotate` for its rotation. 
Outside of the definition of a primitive only `translate` and `rotate are accepted. 
Is that okay?

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
```diff
@@ -42,7 +42,7 @@ function Geometry(::Type{T}, ::Type{Box}, dict::AbstractDict, input_units::Named
         _parse_value(T, dict["hY"], length_unit), 
         _parse_value(T, dict["hZ"], length_unit)
     elseif haskey(dict, "x") && haskey(dict, "y") && haskey(dict, "z")
-        @warn "Deprication warning: Detected old primitive definition for `Box`. 
+        @warn "Deprecation warning: Detected old primitive definition for `Box`. 
             Please update your configuration file to the new format 
             via `widhts`, `haldfwidht` or `hX`, `hY` and `hZ`
```
fhagemann, line 47: I guess this should be `width` and `halfwidth` ?

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -17,24 +17,41 @@ const FullSphere{T,CO} = Ellipsoid{T,CO,T,Nothing,Nothing}
...
+    
     r = parse_r_of_primitive(T, dict, input_units.length)
-    φ = nothing
-    θ = nothing
+    φ = parse_φ_of_primitive(T, dict, angle_unit)
+    if !(φ === nothing)
+        error("Partial Ellipsoid (`φ = φ`) is not yet supported.")
+    end
+    θ = parse_θ_of_primitive(T, dict, angle_unit)
+    if !(φ === nothing)
```
fhagemann, line 31: θ === nothing

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
```diff
@@ -19,19 +19,34 @@ const HexagonalPrism{T,CO,TR}  = RegularPrism{T,CO,6,TR}
...
-    hZ = typeof(z) <: Real ? z : width(z)/2
-    origin = if typeof(z) <: Real
-        CartesianPoint{T}(zero(T), zero(T), zero(T))
-    else
-        CartesianPoint{T}(zero(T), zero(T), mean(z))
+    hZ = if haskey(dict, "h")
+        _parse_value(T, dict["h"], length_unit) / 2
+    elseif haskey(dict, "z")
+        z = parse_height_of_primitive(T, dict, length_unit)
+        hZ = typeof(z) <: Real ? z : (z[2] - z[1])/2
```
fhagemann, line 31: For consistency: `z isa Real` like the line below

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -18,18 +18,39 @@ const FullTorus{T,CO} = Torus{T,CO,T,Nothing,Nothing}
...
+    rotation = get_rotation(T, dict, angle_unit)
+
     r_torus = _parse_value(T, dict["r_torus"], length_unit)
     r_tube = _parse_radial_interval(T, dict["r_tube"], length_unit)
     φ = parse_φ_of_primitive(T, dict, angle_unit)
+    if !(φ === nothing)
+        error("Partial Ellipsoid (`φ = φ`) is not yet supported.")
+    end
     θ = parse_θ_of_primitive(T, dict, angle_unit)
+    if !(φ === nothing)
```
fhagemann, line 31: θ === nothing

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -18,18 +18,39 @@ const FullTorus{T,CO} = Torus{T,CO,T,Nothing,Nothing}
...
     length_unit = input_units.length
     angle_unit = input_units.angle
+    origin = get_origin(T, dict, length_unit)
+    rotation = get_rotation(T, dict, angle_unit)
+
     r_torus = _parse_value(T, dict["r_torus"], length_unit)
     r_tube = _parse_radial_interval(T, dict["r_tube"], length_unit)
     φ = parse_φ_of_primitive(T, dict, angle_unit)
+    if !(φ === nothing)
+        error("Partial Ellipsoid (`φ = φ`) is not yet supported.")
```
fhagemann, line 28: `Torus not supported`

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -18,18 +18,39 @@ const FullTorus{T,CO} = Torus{T,CO,T,Nothing,Nothing}
...
+    end
     θ = parse_θ_of_primitive(T, dict, angle_unit)
+    if !(φ === nothing)
+        error("Partial Ellipsoid (`θ = θ`) is not yet supported.")
```
fhagemann, line 32: `Torus not supported`

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
lmh91: Thanks, I fixed it.

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
lmh91: Ah.. yes...

src/ConstructiveSolidGeometry/VolumePrimitives/RegularPrism.jl
lmh91: Ok.

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
lmh91: I fixed it.

fhagemann (approved):

PR #169, opened 2021-07-05
Require Julia v1.6

lmh91:
Julia v1.0 starts making trouble from time to time. 

Julia v1.7 is already in the beta phase and I believe 1.6 should become the new LTS version of Julia once v1.7 is released. 

So, I would suggest to require Julia v1.6 from now on. 
Since we will have this v0.6-branch on our master branch for some time for testing,
Julia v1.7 will probably be released once we want to make a new release of SSD.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -56,7 +56,7 @@ Tables = "0.2, 1"
 TypedTables = "1.2"
 Unitful = "0.18, 1"
 YAML = "0.3, 0.4"
-julia = "1"
+julia = "1.6"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -56,7 +56,7 @@ Tables = "0.2, 1"
 TypedTables = "1.2"
 Unitful = "0.18, 1"
 YAML = "0.3, 0.4"
-julia = "1"
+julia = "1.6"
```
fhagemann, line 59: Is this the minimum required? Or should it be `"1.6,1"`?

lmh91: "1.6" translates to [1.6, 2[

oschulz:
Sounds like a good idea. It's not entirely clear yet if v1.6 or v1.7 will become LTS (though v1.6 is quire likely), but I don't think we'll have any users that need to use a Julia older than v1.6.

lmh91:
The failed test is solved in #168 so I merge this.

PR #170, opened 2021-07-06
Plotting electric field lines with new CSG

fhagemann:
This PR deals with plotting enhancements, especially concerning `ElectricField` and `ScalarPotential`.

Some minor changes:
- The keyword `full_det` when plotting `φ` crosssections has now been implemented to all scalar potentials (f541ef7) except for `DielectricDistribution`, as this is mapped on a separate grid. 
- `show_normal` default was set to `false` (dec570a). I don't see why anyone would like to see this except for developing/debugging...
- I added a function initially written by @schustermartin to map electric potential on different grids (7be8b01).

### The most major update is reintroducing electric field line plotting with the new CSG (ade28ad).

As the new CSG rethought the whole process of sampling, the new implementation for generating electric field lines is as follows:
- Determine the plane in which the plot will be generated (or define the default plane `x` in `Cartesian` and `φ` in `Cylindrical`).
 Note that this can be any plane, it does **NOT** need to be a symmetry plane (even though everything else makes little sense)
- Determine the contacts to spawn charges on (usually all contacts but the core).
- Divide the plotting plane into a set of "horizontal" and "vertical" lines (`CSG.Line`), e.g. the `xz` plane is represented by a set of equidistant lines along `x` and `z`.
- Sample the contacts by determining the `intersection` of the lines with the `surfaces` of the contacts => `spawn_positions`
- Move these sample points by a given `offset` along the `normal` vector of the surface into the bulk of the detector.
- Drift the charges according to the electric field (similar to `ElectricChargeDriftModel`).
- Plot the projection of the charge drift paths in the plotting plane. 
  If the charge drift paths leave that plane they will still be plotted in that plane.

They are plotting using the old syntax, e.g.
```julia
plot(simulation.electric_field, y = 0)        # for the heatmap in the background
plot_electric_fieldlines!(simulation, y = 0)  # for the electric field lines
```
intersection_ivc_efield

### Some GIFs

Sampling of contacts:
sample

Resulting projection of electric field lines:
electric_field

Slices through the detector volume (even though everything except for `x=0` does not really make sense I guess...)
slices

Works also for rotated detectors:
tilted

Even though the electric field lines look a bit strange there (?)
titled_field_lines

I might have a look into this, which brings me to..

### To be done
- [x] Excessive testing (especially rotated detectors)
- [x] Implement field line plotting in the cylindrical `z` plane in polar projection
- [x] Reimplement electric field line plotting as part of the tutorial
- [x] Implement `normal` for `TorusMantle` as it is needed for the electric field line generation


src/ConstructiveSolidGeometry/IO.jl
```diff
@@ -17,7 +17,11 @@ const CSG_dict = Dict{String, Any}(
 
 function get_geometry_key(::Type{T}, dict::AbstractDict, input_units::NamedTuple) where {T}
     g = collect(filter(k -> k in keys(CSG_dict), keys(dict)))
-    @assert !(length(g) > 1) "Too many geometry entries in dictionary: $(length(g))."
+    filter!(k -> !(
+        (k == "translate" && !any(broadcast(key -> key in keys(CSG_dict), keys(dict["translate"])))) ||
+        (k == "rotate" && !any(broadcast(key -> key in keys(CSG_dict), keys(dict["rotate"]))))
+    )  , g) # exclude translate and rotate that have no primitives inside
+    @assert !(length(g) > 1) "Too many ($(length(g))) geometry entries in dictionary: $(g)."
```
fhagemann, line 24: This might need to be adjusted to changes in #168.

fhagemann:
> Even though the electric field lines look a bit strange there (?)
> titled_field_lines

Ok, the strange behaviour seen in the tilted detector is because the impurity density still increases in `z` and is not rotated together with the detector geometry. We should keep this in mind.

src/ConstructiveSolidGeometry/IO.jl
lmh91: I did not put it into #168 to avoid merge conflicts.

src/PlotRecipes/ElectricField.jl
```diff
@@ -38,43 +38,97 @@ end
...
+        )
+    #=
+    elseif dim_symbol == :z
+        vcat(
+            [Line(CartesianPoint{T}(x,0,v), CartesianVector{T}(0,1,0)) for x in xrange],
+            [Line(CartesianPoint{T}(0,y,v), CartesianVector{T}(1,0,0)) for y in yrange]
+        )
+    =#
+    else # dim_symbol == :r
+        throw(ArgumentError("Plotting field lines at constant 'r' makes no sense."))
```
lmh91: Depends on the geometry. There might be cases. Please just state something like "not yet implemented"

src/PlotRecipes/ElectricField.jl
```diff
@@ -38,43 +38,97 @@ end
...
+            [Line(CartesianPoint{T}(r*cos(v),r*sin(v),0), CartesianVector{T}(0,0,1)) for r in rrange],
+            [Line(CartesianPoint{T}(0,0,z), CartesianVector{T}(cos(v),sin(v),0)) for z in zrange]
+        )
+    #=
+    elseif dim_symbol == :z
+        vcat(
+            [Line(CartesianPoint{T}(x,0,v), CartesianVector{T}(0,1,0)) for x in xrange],
+            [Line(CartesianPoint{T}(0,y,v), CartesianVector{T}(1,0,0)) for y in yrange]
+        )
+    =#
```
lmh91: Can we just take the same as for the Cartesian grid?
We just take -r_max:+r_max as ranges for `y` and `x`.

fhagemann: This changed with my latest commit (d559dac).

lmh91: Ah, alright :+1:

src/PlotRecipes/ElectricField.jl
fhagemann: I just implemented it (6ca509e) but I don't see why anyone would want to use it (especially because the electric field lines are projected to the `r` plane then...)

src/ConstructiveSolidGeometry/IO.jl
fhagemann: Still, you added "translation", "origin", etc.
We have to see that we get this straight one after another.

lmh91: Yes, but only inside primitives. But lets see.

fhagemann:
Well, the electric field lines look decent for our example detectors:

CoaxialTorus
CGD
Hexagon
Spherical
InvertedCoaxInCryostat
BEGe
CGD_CylGrid

lmh91:
Very nice! :+1:

PR #163, opened 2021-06-08
Refactor Volume Primitives and Improvements in Painting

lmh91:
So, as we discussed we want to store the transformation in the primitives. 
This has several advantages, but requires some work.

I removed all Primitives again: Volumes, surfaces and also line primitives (sorry guys).
I implemented the `Box` primitive, which now has its current position and rotation stored in it. 
Transformation applied to the primitive will update those field. 
I added also all surface primitives and line primitives which I need for it.

I added a config file for a box-detector, which is defined around (0,0,0) and is rotated. 
It works quite well: 
rotated_box_det
rotated_box_det_pts

I disabled all tests but the one for the Box detector.

I also made the structs of `Semiconductor`, `SolidStateDetector`, `Contact` and `Passive` immutable,
which improved the application of the boundary conditions quite a bit. 

In addition, the contacts are not sampled anymore but directly painted onto the grid which also improved
the application of the boundary conditions as no vectors of points of unknown length are generated anymore. 

My suggestion how to proceed:
* Make the box detector working on a cylindrical grid
* Add one by one the other primitives

PR #171, opened 2021-07-07
Enhance initialization of `Grid`

lmh91:
This PR addresses the initialization of the Grid. 

In order to have a fast convergence towards equilibrium it is good when the widths of the voxels
are not too different compared to the widths of the neighbor voxels.  

In extreme cases this can lead to the effect of shadowing (blocking the spreading of the electric potential in the SOR).
See comment: https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/163#issuecomment-866699288

shadowing

To avoid this I implemented the function `initialize_axis_ticks` in 4e3c5fb
which add axis ticks next to existing axis ticks (generated by the geometry of objects) where the ratio of the distances to the neighboring ticks (left and right) exceeds a certain limit. The limit ratio can be tuned via a keyword. Default is `2`.

Here an example:
The black lines are the existing ticks and the red lines are all ticks after `initialize_axis_ticks`:

gridspacing

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -47,7 +47,7 @@ end
             signalsum += abs(evt.waveforms[i].value[end])
         end
         @info signalsum
-        @test isapprox( signalsum, T(2), atol = 4e-3 )
+        @test isapprox( signalsum, T(2), atol = 0.02 )
     end
     @testset "Simulate example detector: BEGe" begin
         sim = Simulation{T}(SSD_examples[:BEGe])
@@ -59,7 +59,7 @@ end
             signalsum += abs(evt.waveforms[i].value[end])
         end
         @info signalsum
-        @test isapprox( signalsum, T(2), atol = 3e-3 )
+        @test isapprox( signalsum, T(2), atol = 3e-2 )
     end
     @testset "Simulate example detector: HexagonalPrism" begin
         sim = Simulation{T}(SSD_examples[:Hexagon])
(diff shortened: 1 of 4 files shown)


fhagemann:
Also, the function `get_world_limits_from_objects` would require an update as it relies on functions that do not longer work. Is this the PR to implement this?

lmh91:
> Also, the function `get_world_limits_from_objects` would require an update as it relies on functions that do not longer work. Is this the PR to implement this?

Right, this has to be update. But I did not intended it for this PR.

test/runtests.jl
```diff
@@ -47,7 +47,7 @@ end
             signalsum += abs(evt.waveforms[i].value[end])
         end
         @info signalsum
-        @test isapprox( signalsum, T(2), atol = 4e-3 )
+        @test isapprox( signalsum, T(2), atol = 0.02 )
```
fhagemann, line 50: Any idea why the `signalsum` is getting worse? Is it a coarser grid now?

lmh91: I believe so. By default the initial grid is really coarse now.

fhagemann: Then, I would prefer increasing the number of refinements rather than lowering the tolerance.
It might be that we miss some bugs by allowing greater tolerances.

lmh91: I agree, but the refinement will also be updated soon. 
Including keywords and default settings. 
And we might want to overhaul the tests (especially simulation of test detectors) anyhow. 
So I would prefer not to do this here.

fhagemann:
I was about to ask whether you could shortly explain what your changes are and how this enhances the initialization :sweat_smile:

PR #172, opened 2021-07-07
Change output of intersection of Plane with Line to Tuple

fhagemann:
Previously, an `intersection` of an `AbstractPlanarSurfacePrimitive` with a `Line` would return a `CartesianPoint`, whereas an `intersection` of an `AbstractCurvedSurfacePrimitive` with a `Line` would result in a `NTuple{CartesianPoint}`.

By changing the output of `intersection` of planar surfaces with lines to a single-entry `Tuple{CartesianPoint}`, we do not need to distinguish between planar and curved surfaces in painting or electric field line plotting which reduces the number of code lines.


src/PotentialSimulation/Painting/Painting.jl
```diff
@@ -35,5 +35,156 @@ get_sub_ind_ranges(p::ConstructiveSolidGeometry.AbstractSurfacePrimitive{T}, gri
...
+                i2 = searchsortednearest(ticks[2], pt[2])
+                csgtol = abs(ticks[2][i2] - pt[2])
+                Δw_max = max(
+                    widths_ax1[i1], widths_ax1[i1+1],
+                    # widths_ax2[i2], widths_ax2[i2+1],
+                    widths_ax3[i3], widths_ax3[i3+1]
+                )/4
+                if csgtol > Δw_max
+                    csgtol = Δw_max
+                end
```
fhagemann, line 83: The newest commit chooses the values for `Δw_max` and the handling of `csgtol` like it was done for `AbstractPlanarSurfacePrimitives`. For `AbstractCurvedSurfacePrimtives`,  `Δw_max` was divided by `2` instead of `4` and `csgtol` was **NOT** overwritten if it was greater than `Δw_max`. I cannot judge why the differently bent surface primitives were handled differently before but it seems like the tests fail with the implementation for curved surfaces applied to all surfaces...

lmh91:
Thanks!

PR #177, opened 2021-07-13
Update docs and transition from JSON to YAML

lmh91:
Docs PR build: https://juliaphysics.github.io/SolidStateDetectors.jl/previews/PR177/


docs/src/man/primitives.jl
```diff
@@ -4,7 +4,8 @@ using SolidStateDetectors
 import SolidStateDetectors.ConstructiveSolidGeometry as CSG
 using SolidStateDetectors.ConstructiveSolidGeometry: ClosedPrimitive
 using SolidStateDetectors.ConstructiveSolidGeometry: Box
-using Plots
+using Plots; gr()
+Plots.reset_defaults()
```
fhagemann, line 8: What does this do?

lmh91: No it didn't...

lmh91: 30e6216 solved it. I removed the `Plots.reset_defaults` again (force push).

lmh91:
Okay, it is ready I guess. 
But it reminded me that I wanted to put the implementation of
proper logging, debug and warning outputs onto the ToDo list...

docs/src/man/primitives.jl
fhagemann: well, `Plots.reset_defaults()` still seems to be in that file..

lmh91: Ah, right. It was in `primitives.jl` as well... We can clean this up later.

Issue #68, opened 2020-10-09
Improve IO regarding flexibility


lmh91:

fraseyboo:
Regarding this issue it might be useful to support writing VTK files using VTKWrite.jl, I currently use VTK for a lot of my PSA work because the wide support for large data types is unparalleled. All the fields and electrodes should be relatively simple to export as rectilinear grids and polydata and can be easily interpreted externally in Paraview, Visit, Slicer or a native VTK program. 

![ssd_potential_bottom](https://user-images.githubusercontent.com/33124758/111006264-664a5000-8384-11eb-8acb-eba16f756905.png)
![iso](https://user-images.githubusercontent.com/33124758/111006268-68acaa00-8384-11eb-809f-f273235238e6.png)

oschulz:
Sounds very useful - we should add it as an optional dependency via Requires, though.

PR #182, opened 2021-07-15
Update configuration file conversion script

fhagemann:
Following the changes in #168, the configuration conversion script had to be updated.
While testing it, I found a bug in the `Cone` read-in (90806ee) that would lead to misinterpretations of`VaryingCylinder`.

To update config files, the user has to load `<SSD>/test/update_config_file.jl` 
(which defines a function `update_config_file()`) 
and pass the filename to that function `update_config_file` to update it to v0.6.0.

### DIFF
--- a/src/ConstructiveSolidGeometry/Transformations.jl
+++ b/src/ConstructiveSolidGeometry/Transformations.jl
@@ -25,3 +25,17 @@ function combine_transformations(current::Transformations, new::Transformations)
     (rotation = r, translation = t)
 end
 
+
+function Dictionary(m::SMatrix{3,3,T,9}) where {T}
+    dict = OrderedDict{String, Any}()
+    mat = RotXYZ(m)
+    if mat.theta1 == 0 && mat.theta2 == 0
+        if mat.theta3 != 0
+            dict["Z"] = string(mat.theta3)*"rad"
+        end
+    else
+        dict["M"] = m[:]
+    end
+    dict
+end
+
--- a/src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
+++ b/src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
@@ -120,28 +120,28 @@ ___
 _____\
 
 =#
-const VaryingCylinder{T,CO} = Cone{T,CO,Tuple{T,T},Nothing} # Full in φ
-const PartialVaryingCylinder{T,CO} = Cone{T,CO,Tuple{T,T},Tuple{T,T}}
+const VaryingCylinder{T,CO} = Cone{T,CO,Tuple{Tuple{T},Tuple{T}},Nothing} # Full in φ
+const PartialVaryingCylinder{T,CO} = Cone{T,CO,Tuple{Tuple{T},Tuple{T}},Tuple{T,T}}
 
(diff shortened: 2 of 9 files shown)


src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -604,6 +604,39 @@ function Geometry(::Type{T}, t::Type{Cone}, dict::AbstractDict, input_units::Nam
...
+        end
+        dict["r"]["top"] = if c.r[2] isa Tuple{<:Real} 
+            c.r[2][1]
+        elseif isnothing(c.r[2])
+            0
+        else 
+            OrderedDict{String, Any}("from" => c.r[2][1], "to" => c.r[2][2])
+        end
+    end
+    if !isnothing(c.φ) dict["phi"] = OrderedDict("from" => string(c.φ[1])*"rad", "to" => string(c.φ[2])*"rad") end
```
fhagemann, line 633: Non-2π phi-intervals will always be saved in units of `rad` with the respective unit attached.

src/ConstructiveSolidGeometry/Transformations.jl
```diff
@@ -25,3 +25,17 @@ function combine_transformations(current::Transformations, new::Transformations)
     (rotation = r, translation = t)
 end
 
+
+function Dictionary(m::SMatrix{3,3,T,9}) where {T}
+    dict = OrderedDict{String, Any}()
+    mat = RotXYZ(m)
+    if mat.theta1 == 0 && mat.theta2 == 0
+        if mat.theta3 != 0
+            dict["Z"] = string(mat.theta3)*"rad"
```
fhagemann, line 34: Angles of rotation matrices around `z` will be saved in units of `rad`.

fhagemann (commented):
One thing: `Dictionary`, right now, does not know about units. Thus, all angles are saved in units of "rad". There might be a more elegant work-around that this..

lmh91 (approved):
Looks fine to me :+1:

fhagemann:
Do we deal with units later (#68)?

lmh91:
> Do we deal with units later (#68)?

Sorry, I don't understand your question. #68 is about the IO of calculated fields, e.g. the ElectricPotential.
Not configuration files. Am I missing something?

fhagemann:
> > Do we deal with units later (#68)?
> 
> Sorry, I don't understand your question. #68 is about the IO of calculated fields, e.g. the ElectricPotential.
> Not configuration files. Am I missing something?

Oh, I thought #68 was also about how to pass units to the methods that calculate stuff (that you can give ticks of plots also in `mm` and not in unitless SI numbers) and how to deal with units in general. We should add a description to that issue.. :sweat_smile:

lmh91:
I think #68 is about some other IO of the simulated fields in addition to HDF5. 

But, yes, regarding plotting this would indeed be nice. But it does not belong here (this PR) in my opinion. 
We can later add some (optional) field to the configuration file like
```yaml
plotunits:
  lenghts: mm
  ...
```
or derive it from the general `unit` field.

fhagemann:
Ok, then let's merge this and care about units later.

Issue #173, opened 2021-07-08
Write documentation for v0.6 release


fhagemann:
Find a list of things that need to be written/updated for the v0.6 documentation:
- How to define a `ChargeDriftModel`
- More information on the `ADLChargeDriftModel` (#144)
- How to define an `ImpurityDensityModel` (#140)
- How to create/structure config files (#136), especially concerning rotations (#62), crystal orientation (#161), and other stuff that got added
- How to split config files (#104)
- How to plot potentials and fields (`ElectricPotential`, `ElectricField` + lines, ...)
- Docstrings for exported functions
- Add a reference to the SSD publication ([arXiv](https://arxiv.org/abs/2104.00109) or JINST publication) to `README.md` (?)
 -->  #175 

Feel free to complete this list.

Issue #94, opened 2020-11-12
HDF5 IO integration


leenderthayen:
Hello everyone,

It appears that currently the IO capabilities for HDF5 saving and loading are located in an unregistered package, LegendHDF5IO. Is it possible to integrate this directly into SSD, or if not bind the github repo's together?

Thanks!

lmh91:
So what SSD provides is a conversion (and reconversion) of several internal structs into NamedTuples.
LegendHDF5IO provides a generic IO of NamedTuples to HDF5. 
You are right, this should be an LEGEND independent official registered package. 
So far, we (@oschulz) had no time for this.

oschulz:
We have indeed been considering to turn (part of) LegendHDF5IO into a generic registered Julia package, but that will require some careful "surgery" and will probably still take a while. However, we wouldn't want to add heavy additional dependencies like HDF5 to SolidStateDetectors, so SolidStateDetectors still wouldn't depend on it. The current mechanism of conversion from/to NamedTuples gives the user freedom regarding which kind of I/O they want to use.

Having said that, I'm planning a custom package registry for LEGEND (since the LEGEND packages are too experiment-specific for the General registry). That will make it easier to install LegendHDF5IO in addition to SolidStateDetectors.

We should also check if JLD2 can be used out of the box (or with some tweaks) to store/load SolidStateDetectors detector definitions and fields. This may be an attractive option for non-LEGEND-users. There have been some improvements in JLD2 recently.

leenderthayen:
We are currently using JLD2, so we can close this for now.

PR #184, opened 2021-07-16
Write documentation for v0.6 release

fhagemann:
Docs preview: https://juliaphysics.github.io/SolidStateDetectors.jl/previews/PR184/

Checklist from #173 (will be updated as we go).

- [x] How to define a `ChargeDriftModel`
- [x] More information on the `ADLChargeDriftModel`
  - [x] what is different in comparison to Bruyneel et al.
  - [x] crystal orientation
- [x] How to define an `ImpurityDensityModel`
- [x] How to define a `ChargeDensityModel`
- [x] How to create/structure config files
- [x] CSG
  - [x] rotations 
  - [x] translations
- [x] How to split config files


docs/src/man/csg.md
```diff
@@ -1,56 +1,71 @@
...
+      r: 2
+      h: 1
+  - tube: # B
+      r: 1
+      h: 1
 ```
+![CSGDifference](../assets/CSGDifference.png)
+
+If more than two geometries are passed, all entries starting from the second will be subtracted from the first.
+
```
lmh91, line 52: In this case the top and bottom surface of `A` would still be fully included as `B` is turned into an 
`OpenPrimitive`. We might should explain that somewhere (maybe here it becomes crucial).
`h` of `B` should be larger than `h` of `A` probably.

fhagemann: b8ef8e6

docs/src/man/IO.md
```diff
@@ -1,3 +1,23 @@
...
+FileIO.save("<name-of-simulation-file>.jld", Dict("Simulation" => simulation)
+```
+
+It can be read back in using
+```yaml 
+using FileIO
+FileIO.load("<name-of-simulation.file>.jld", "Simulation")
+```
+
+Other more compact ways of saving simulation results are based on the HDF5 saving format and the package [HDF5.jl](https://github.com/JuliaIO/HDF5.jl).
```
fhagemann, line 23: @oschulz: Any updates #94 on making LegendHDF5IO.jl some stand-alone registered package (#94)?

fhagemann:
What is still missing:
- Docstrings for all exported functions (after clean-up)
- Detailed explanations on how to plot stuff

lmh91 (approved):
Looks very nice to me! :)

Issue #178, opened 2021-07-14
Enhance Refinement


lmh91:
Currently, only one single axis tick is inserted between two ticks if the electric potential difference, `ΔΦ`, 
is above a certain limit, `ΔΦ_max`, independent of how large the difference is.
Thus, a lot of refinements steps are needed to get a good grid.

We should allow that multiple, `N`, ticks can be inserted between two ticks in one refinement step:
`N = round(ΔΦ / ΔΦ_max)`

By this we will also have to change the keywords regarding the grid and refinement in functions like
`simulate!` and `calculate_electric_potential!`.

PR #185, opened 2021-07-19
Enhancement of the Refinement

lmh91:
In this PR the refinement of the grid is refactored: See also #178.

## Refinement

The high level keyword `max_nrefinements` is removed and the refinement is controlled via
the keyword `refinement_limits`. Prior to this PR, the keyword `refinement_limits` was a Tuple of 3 `Reals` 
to set the allowed relative (to the applied bias voltage) maximum difference in the potential between to grid points for each dimension. 

Now this keywords can take also a Vector of these tuples, where the first element sets the maximum allowed differences
for the first refinement. The second element for the second refinement.... 
Also, not only 1 tick is inserted, but multiple points can be inserted in one step depending on how large the difference is between two grid points. 

`refinement_limits` can also be just a single `Real` or a `Vector{<:Real}`.
E.g. `[0.2, 0.1, 0.05]`. Than, each element is converted into a NTuple{3, Real} with the same values:
`[(0.2, 0.2, 0.2), (0.1, 0.1, 0.1), ...]`

Here, in the first refinement, the maximum allowed relative potential difference would be 20%. 
Thus, the absolute allowed difference would be `0.2 * bias_voltage`.

I rename the old keyword `min_grid_spacing` to `min_tick_distance` in order to be consistent with 
`max_tick_distance` (see below). It also takes quantities now, see also `max_tick_distance`.
E.g. `min_tick_distance = 10u"nm"`.
(In cylindrical coordinates a single length value is translated into an angle at a radius of
`(sim.world.r[end] + sim.world.r[1])/2`)

## Initial Grid

The keywords of the initialization of the Grid have also changed. There a two keywords:
`max_tick_distance` and `max_distance_ratio`. 

`max_tick_distance` defines the maximum allowed distances between two ticks for each dimension. 
It takes units. E.g. for the cylindrical grid:
`Grid(sim_cyl, max_tick_distance = (0.3u"cm", 15u"°", 1u"m"))`
Also, a single value can be parsed: e.g. `max_tick_distance = 1u"mm"`.

For `max_distance_ratio::Real` look into #171.

Both keywords can also be given to `simulate!`, `calculate_electric_potential!` and `calculate_weighting_potential!`.

### DIFF
--- a/examples/example_config_files/public_CGD_config.yaml
+++ b/examples/example_config_files/public_CGD_config.yaml
@@ -24,7 +24,7 @@ detectors:
   - translate: 
       x: 0
     rotate: 
-      Z: 45
+      Z: 0
     bulk:
       material: HPGe
       temperature: 78
--- a/src/PotentialSimulation/ConvergenceAndRefinement.jl
+++ b/src/PotentialSimulation/ConvergenceAndRefinement.jl
@@ -234,3 +234,10 @@ function _refine_axis(ax::DiscreteAxis{T, <:Any, <:Any, ClosedInterval{T}}, ns::
     ticks[end] = ax.ticks[end]
     typeof(ax)(ax.interval, ticks)
 end
+
+
+function _extend_refinement_limits(rl::Union{<:Real, Vector{<:Real}, Tuple{<:Real,<:Real,<:Real}, Vector{<:Tuple{<:Real, <:Real, <:Real}}})
+    0
+end
+_extend_refinement_limits(rl::Real) = (rl, rl, rl )
+_extend_refinement_limits(rl::Tuple{<:Real,<:Real,<:Real}) = rl
--- a/src/PotentialSimulation/ConvergenceAndRefinement.jl
+++ b/src/PotentialSimulation/ConvergenceAndRefinement.jl
@@ -233,6 +233,13 @@ function _create_refined_grid(p::ScalarPotential{T,3}, max_diffs::NTuple{3, T}, 
             end
         end
     end
+    for i in 1:3 # always add an even number of ticks
+        if isodd(sum(ns[i])) 
+            i_max_width = findmax(sub_widths[i])[2]
+            ns[i][i_max_width] += 1 
+            sub_widths[i][i_max_width] = widths[i][i_max_width] / (ns[i][i_max_width]+1)
+        end
+    end
     new_axes = broadcast(i -> _refine_axis(p.grid.axes[i], ns[i], sub_widths[i]), (1, 2, 3))
     return typeof(p.grid)(new_axes)
 end
--- a/src/PotentialSimulation/ConvergenceAndRefinement.jl
+++ b/src/PotentialSimulation/ConvergenceAndRefinement.jl
@@ -89,7 +89,8 @@ end
 
 
 """
-    refine_scalar_potential(p::ScalarPotential{T}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}) where {T}
+    refine_scalar_potential(p::ScalarPotential{T}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}; 
+        only2d::Val{only_2d} = Val(size(p.data, 2)==1)) where {T, only_2d}
 
 Refine any scalar potential `p`. 
 
@@ -99,25 +100,42 @@ Refine any scalar potential `p`.
    smaller than `max_diff[i]` (i -> dimension) but that the distances between the ticks stays larger than `minimum_distances[i]`.
 3. Create the new data array for the refined grid and fill it by interpolation of the the initial (coarse) grid.
 """
-function refine_scalar_potential(p::ScalarPotential{T}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}) where {T}
+function refine_scalar_potential(p::ScalarPotential{T}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}; 
+        only2d::Val{only_2d} = Val(size(p.data, 2)==1)) where {T, only_2d}
     closed_potential = _get_closed_potential(p)
-    int = interpolate_closed_potential(closed_potential)
     new_grid = _create_refined_grid(closed_potential, T.(max_diffs), T.(minimum_distances))
     new_data = Array{T, 3}(undef, size(new_grid))
-    for i3 in axes(new_data, 3)
-        x3 = new_grid.axes[3].ticks[i3]
-        for i2 in axes(new_data, 2)
-            x2 = new_grid.axes[2].ticks[i2]
+    if only_2d
+        int = interpolate_closed_potential(closed_potential, only2d)
+        for i3 in axes(new_data, 3)
+            x3 = new_grid.axes[3].ticks[i3]
             for i1 in axes(new_data, 1)
                 x1 = new_grid.axes[1].ticks[i1]
-                new_data[i1, i2, i3] = int(x1, x2, x3)
+                new_data[i1, 1, i3] = int(x1, x3)
+            end
+        end
+    else
+        int = interpolate_closed_potential(closed_potential, only2d)
+        for i3 in axes(new_data, 3)
+            x3 = new_grid.axes[3].ticks[i3]
+            for i2 in axes(new_data, 2)
+                x2 = new_grid.axes[2].ticks[i2]
+                for i1 in axes(new_data, 1)
+                    x1 = new_grid.axes[1].ticks[i1]
+                    new_data[i1, i2, i3] = int(x1, x2, x3)
+                end
             end
         end
     end
     return _convert_closed_potential(typeof(p), ElectricPotential(new_data, new_grid))
 end             
 
-interpolate_closed_potential(p::ScalarPotential) = interpolate(p.grid.axes, p.data, Gridded(Linear()))
+function interpolate_closed_potential(p::ScalarPotential, ::Val{true}) where {T}
+    interpolate((p.grid.axes[1], p.grid.axes[3]), p.data[:,1,:], Gridded(Linear()))
+end
+function interpolate_closed_potential(p::ScalarPotential, ::Val{false}) where {T}
+    interpolate(p.grid.axes, p.data, Gridded(Linear()))
+end
 
 _get_closed_ticks(ticks::Vector{T}, int::ClosedInterval{T}) where {T} = ticks
 _get_closed_ticks(ticks::Vector{T}, int::Interval{:closed, :open, T}) where {T} = vcat(ticks, [int.right])
--- a/src/PotentialSimulation/ConvergenceAndRefinement.jl
+++ b/src/PotentialSimulation/ConvergenceAndRefinement.jl
@@ -170,10 +170,10 @@ end
 """
     _get_closed_potential(p::ScalarPotential{T,3,CS}) where {T, CS}
 
-Returnes an closed Grid & Potential:
+Returns an closed Grid & Potential:
 E.g. if one of the axis is {:closed,:open} it will turn this into {:closed,:closed}
 and also extend the `data` field of the potential in the respective dimension and fill
-it with the respective values. 
+it with the respective values.
 """
 function _get_closed_potential(p::ScalarPotential{T,3,CS}) where {T, CS}
     g = p.grid
@@ -185,7 +185,6 @@ function _get_closed_potential(p::ScalarPotential{T,3,CS}) where {T, CS}
     closed_grid = Grid{T, 3, CS, AT}(closed_axes)
     ScalarPotential(p, closed_values, closed_grid)
 end
-
 
 """
     _convert_to_original_potential(::Type{P}, data, grid) where {P, T, CS}
--- a/src/PotentialSimulation/ConvergenceAndRefinement.jl
+++ b/src/PotentialSimulation/ConvergenceAndRefinement.jl
@@ -88,7 +88,17 @@ function _update_till_convergence!( fssrb::PotentialSimulationSetupRB{T, N1, N2}
 end
 
 
+"""
+    refine_scalar_potential(p::ScalarPotential{T}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}) where {T}
 
+Refine any scalar potential `p`. 
+
+1. Extent the grid to be a closed grid in all dimensions. 
+2. Refine the axis of the grid based on `max_diffs` and `minimum_applied_potential`:
+   Insert N new ticks between to existing ticks such that the potential difference between each tick becomes
+   smaller than `max_diff[i]` (i -> dimension) but that the distances between the ticks stays larger than `minimum_distances[i]`.
+3. Create the new data array for the refined grid and fill it by interpolation of the the initial (coarse) grid.
+"""
 function refine_scalar_potential(p::ScalarPotential{T}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}) where {T}
     closed_potential = _get_closed_potential(p)
     int = interpolate_closed_potential(closed_potential)
@@ -111,6 +121,7 @@ interpolate_closed_potential(p::ScalarPotential) = interpolate(p.grid.axes, p.da
 
 _get_closed_ticks(ticks::Vector{T}, int::ClosedInterval{T}) where {T} = ticks
 _get_closed_ticks(ticks::Vector{T}, int::Interval{:closed, :open, T}) where {T} = vcat(ticks, [int.right])
+_get_closed_ticks(ticks::Vector{T}, int::Interval{:open, :closed, T}) where {T} = vcat([int.left], ticks)
 
 _get_closed_values(values::Array{T,3}, dim::Int, int::ClosedInterval{T}) where {T} = values
 
@@ -121,6 +132,15 @@ function _get_closed_values(values::Array{T,3}, dim::Int, int::Interval{:closed,
         cat(values, values[:,1:1,:], dims=dim)
     else # dim == 3
         cat(values, values[:,:,1:1], dims=dim)
+    end
+end
+function _get_closed_values(values::Array{T,3}, dim::Int, int::Interval{:open, :closed, T})::Array{T,3} where {T} 
+    if dim == 1
+        cat(values[end:end,:,:], values, dims=dim)
+    elseif dim == 2
+        cat(values[:,end:end,:], values, dims=dim)
+    else # dim == 3
+        cat(values[:,:,end:end], values, dims=dim)
     end
 end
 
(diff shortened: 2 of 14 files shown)


src/PotentialSimulation/ScalarPotential.jl
```diff
@@ -1,4 +1,8 @@
-const ScalarPotential{T, N, S} = Union{ElectricPotential{T, N, S}, WeightingPotential{T, N, S}, PointTypes{T, N, S}, EffectiveChargeDensity{T, N, S}}
+const ScalarPotential{T, N, S, AT} = Union{ElectricPotential{T, N, S, AT}, WeightingPotential{T, N, S, AT}, PointTypes{T, N, S, AT}, EffectiveChargeDensity{T, N, S, AT}}
```
fhagemann, line 1: Should we introduce a type here?
`ElectricPotential <: ScalarPotential <: AbstractArray`?

src/PotentialSimulation/ScalarPotential.jl
```diff
@@ -16,76 +20,79 @@ function getindex(ep::P, g::Grid{T, N, S}) where {T, N, S, P <: ScalarPotential{
...
+        new_pot[:, i, :] = sp[:, 1, :]
     end
-    new_axφ::DiscreteAxis{T, :periodic, :periodic} = DiscreteAxis{T, :periodic, :periodic}( new_int, new_ticks )
-    new_grid::Grid{T, 3, Cylindrical} = Grid{T, 3, Cylindrical}( (wp.grid[1], new_axφ, wp.grid[3]) )
-    new_wp::Potential = Potential( new_pot, new_grid )
-    return new_wp
+    new_axφ = DiscreteAxis{T, :periodic, :periodic}( new_int, new_ticks )
+    new_axes = (sp.grid[1], new_axφ, sp.grid[3])
+    new_grid = Grid{T, 3, Cylindrical, typeof(new_axes)}( new_axes )
+    PT = typeof(sp).name.name
```
fhagemann, line 81: `nameof(sp)`?

examples/example_config_files/public_CGD_config.yaml
```diff
@@ -24,7 +24,7 @@ detectors:
   - translate: 
       x: 0
     rotate: 
-      Z: 45
+      Z: 0
```
fhagemann, line 27: Why? Does this not work with rotated geometries?

src/PotentialSimulation/ScalarPotential.jl
lmh91: I am actually thinking whether we should remove `ElectricPotential`, `WeightingPotential`, `PointTypes`, `EffectiveChargeDensity` and just define `ScalarPotential` and parse the information what is stored in that object in some other why. But, this would require changes at several places.

src/PotentialSimulation/ScalarPotential.jl
lmh91: `nameof(typeof(sp))` should work.

examples/example_config_files/public_CGD_config.yaml
lmh91: It does. But you we might need to refine it more. And I wan't to keep the computations in the test at a minimum.

src/PotentialSimulation/ScalarPotential.jl
lmh91: done: 7ffbd72

src/Simulation/Simulation.jl
```diff
@@ -838,18 +840,24 @@ end
 ToDo...
 """
 function simulate!( sim::Simulation{T};  
-                    refinement_limits = [0.2, 0.1, 0.05 ],
+                    refinement_limits = [0.2, 0.1, 0.05],
```
fhagemann, line 847: `simulate!()` only had one refinement as default. How does this compare to the old default as it is refined thrice now? Is the accuracy the same/better/worse?

lmh91: It is finer now.

src/PotentialSimulation/ConvergenceAndRefinement.jl
```diff
@@ -88,282 +88,181 @@ function _update_till_convergence!( fssrb::PotentialSimulationSetupRB{T, N1, N2}
...
+function interpolate_closed_potential(p::ScalarPotential, ::Val{true}) where {T}
+    interpolate((p.grid.axes[1], p.grid.axes[3]), p.data[:,1,:], Gridded(Linear()))
+end
+function interpolate_closed_potential(p::ScalarPotential, ::Val{false}) where {T}
+    interpolate(p.grid.axes, p.data, Gridded(Linear()))
 end
 
+_get_closed_ticks(ticks::Vector{T}, int::ClosedInterval{T}) where {T} = ticks
+_get_closed_ticks(ticks::Vector{T}, int::Interval{:closed, :open, T}) where {T} = vcat(ticks, [int.right])
+_get_closed_ticks(ticks::Vector{T}, int::Interval{:open, :closed, T}) where {T} = vcat([int.left], ticks)
```
fhagemann, line 142: To reduce allocations: `vcat(ticks, int.right)` or `vcat(int.left, ticks)` without putting them in `Array` brackets

src/PotentialSimulation/ConvergenceAndRefinement.jl
```diff
@@ -88,282 +88,181 @@ function _update_till_convergence!( fssrb::PotentialSimulationSetupRB{T, N1, N2}
...
-                    for i in eachindex(axφ)
-                        if !in(i, inds_φ) push!(inds_φ, i) end
-                    end
-                    b = false
-                end
-            end
-        end    
-        @assert iseven(length(inds_φ)) "Refinement would result in uneven grid in φ."
     end
+    return _convert_closed_potential(typeof(p), ElectricPotential(new_data, new_grid))
```
fhagemann, line 130: Can this be solved more elegantly by making `ScalarPotential` a type?
Here, it looks like you first save it as `ElectricPotential` to convert it to whatever you need afterwards (?)

src/PotentialSimulation/ConvergenceAndRefinement.jl
```diff
@@ -88,282 +88,181 @@ function _update_till_convergence!( fssrb::PotentialSimulationSetupRB{T, N1, N2}
...
+"""
+function _get_closed_potential(p::ScalarPotential{T,3,CS}) where {T, CS}
+    g = p.grid
+    closed_values = _get_closed_values(p.data, 1, g.axes[1].interval)
+    closed_values = _get_closed_values(closed_values, 2, g.axes[2].interval)
+    closed_values = _get_closed_values(closed_values, 3, g.axes[3].interval)
+    closed_axes = broadcast(i -> _get_closed_axis(g.axes[i]), (1, 2, 3))
+    AT = typeof(closed_axes)
+    closed_grid = Grid{T, 3, CS, AT}(closed_axes)
+    ElectricPotential{T, 3, CS, AT}(closed_values, closed_grid)
```
fhagemann, line 186: Same here: this is by default `ElectricPotential` rather than the initial type of `p`?

src/PotentialSimulation/ConvergenceAndRefinement.jl
```diff
@@ -88,282 +88,181 @@ function _update_till_convergence!( fssrb::PotentialSimulationSetupRB{T, N1, N2}
...
 
+_get_ind_range_of_closed_axis(l::Int, ::Type{DiscreteAxis{T, BL, BR, ClosedInterval{T}}}) where {T, BL, BR} = 1:l
+_get_ind_range_of_closed_axis(l::Int, ::Type{DiscreteAxis{T, BL, BR, Interval{:closed, :open, T}}}) where {T, BL, BR} = 1:(l-1)
+_get_ind_range_of_closed_axis(l::Int, ::Type{DiscreteAxis{T, BL, BR, Interval{:open, :closed, T}}}) where {T, BL, BR} = 2:l
+
+
+function _create_refined_grid(p::ScalarPotential{T,3}, max_diffs::NTuple{3, T}, minimum_distances::NTuple{3, T}) where {T}
+    n_1 = Int.(floor.([maximum(abs.(p.data[i+1,:,:] .- p.data[i,:,:])) for i in 1:size(p.data, 1)-1] ./ max_diffs[1], digits=0)) 
+    n_2 = Int.(floor.([maximum(abs.(p.data[:,i+1,:] .- p.data[:,i,:])) for i in 1:size(p.data, 2)-1] ./ max_diffs[2], digits=0)) 
+    n_3 = Int.(floor.([maximum(abs.(p.data[:,:,i+1] .- p.data[:,:,i])) for i in 1:size(p.data, 3)-1] ./ max_diffs[3], digits=0))
```
fhagemann, line 224: `Int.(floor.(..., digits = 0))` is just `floor.(Int, ...)`?

src/PotentialSimulation/ConvergenceAndRefinement.jl
```diff
@@ -88,282 +88,181 @@ function _update_till_convergence!( fssrb::PotentialSimulationSetupRB{T, N1, N2}
...
-    potential, grid = add_points_and_interpolate(p.data, p.grid, refine_at_inds...)
-    return typeof(p)(potential, grid)
-end
\ No newline at end of file
+
+function _extend_refinement_limits(rl::Union{<:Real, Vector{<:Real}, Tuple{<:Real,<:Real,<:Real}, Vector{<:Tuple{<:Real, <:Real, <:Real}}})
+    0
+end
+_extend_refinement_limits(rl::Real) = (rl, rl, rl )
+_extend_refinement_limits(rl::Tuple{<:Real,<:Real,<:Real}) = rl
```
fhagemann, line 268: `_extend_refinement_limits` is defined twice for `<:Real` (L.265/267) and `Tuple[<:Real, <:Real, <:Real}` (L.265/268). Is this intended?

src/PotentialSimulation/ScalarPotential.jl
```diff
@@ -16,76 +20,79 @@ function getindex(ep::P, g::Grid{T, N, S}) where {T, N, S, P <: ScalarPotential{
...
+function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::DiscreteAxis{T, :periodic, :periodic}, int::Interval{:closed, :open, T}) where {T}
     @assert int.right != 0 "Right boundary of φ interval is not allowed to be 0"
-    Potential::Type = typeof(wp)
     l::Int = length( axφ )
     Δφ::T = int.right - int.left
     new_int::Interval{:closed, :open, T} = Interval{:closed, :open, T}(0, 2π)
     n::Int = Int(round(T(2π) / Δφ, sigdigits = 6))
     new_ticks::Vector{T} = Vector{T}(undef, l * n)
-    new_pot::Array{T, 3} = Array{T, 3}(undef, size(wp, 1), l * n, size(wp, 3))
+    new_pot = Array{eltype(sp.data), 3}(undef, size(sp, 1), l * n, size(sp, 3))
```
fhagemann, line 30: Why use `eltype(sp.data)` instead of `T`?

src/PotentialSimulation/ScalarPotential.jl
```diff
@@ -16,76 +20,79 @@ function getindex(ep::P, g::Grid{T, N, S}) where {T, N, S, P <: ScalarPotential{
...
     @assert int.right != 0 "Right boundary of φ interval is not allowed to be 0"
-    Potential::Type = typeof(wp)
     l::Int = 2 * length( axφ ) - 2
     Δφ::T = int.right - int.left
     new_int::Interval{:closed, :open, T} = Interval{:closed, :open, T}(int.left, int.right + Δφ)
     new_ticks::Vector{T} = Vector{T}(undef, l)
     new_ticks[1:length( axφ )] = collect(axφ.ticks)
-    new_pot::Array{T, 3} = Array{T, 3}(undef, size(wp, 1), l, size(wp, 3))
-    new_pot[:, 1:length( axφ ), :] = wp.data[:, :, :]
+    new_pot = Array{eltype(sp.data), 3}(undef, size(sp, 1), l, size(sp, 3))
```
fhagemann, line 53: Same here

src/PotentialSimulation/ScalarPotential.jl
```diff
@@ -16,76 +20,79 @@ function getindex(ep::P, g::Grid{T, N, S}) where {T, N, S, P <: ScalarPotential{
...
 end
 
-function extend_2D_to_3D_by_n_points(wp::ScalarPotential{T, 3, Cylindrical}, axφ::DiscreteAxis{T, :reflecting, :reflecting}, int::Interval{:closed, :closed, T},
+function extend_2D_to_3D_by_n_points(sp::ScalarPotential{T, 3, Cylindrical}, axφ::DiscreteAxis{T, :reflecting, :reflecting}, int::Interval{:closed, :closed, T},
         n_points_in_φ::Int ) where {T}
-    Potential::Type = typeof(wp)
     new_int::Interval{:closed, :open, T} = Interval{:closed, :open, T}(0, 2π)
     new_ticks::Vector{T} = Vector{T}(undef, n_points_in_φ)
-    new_pot::Array{T, 3} = Array{T, 3}(undef, size(wp, 1), n_points_in_φ, size(wp, 3))
+    new_pot = Array{eltype(sp.data), 3}(undef, size(sp, 1), n_points_in_φ, size(sp, 3))
```
fhagemann, line 72: And here.

src/Simulation/Simulation.jl
```diff
@@ -176,112 +176,85 @@ end
...
+    world_Δz = world.intervals[3].right - world.intervals[3].left
+    world_Δφ = world.intervals[2].right - world.intervals[2].left
+    world_Δr = world.intervals[1].right - world.intervals[1].left
+    world_r_mid = (world.intervals[1].right + world.intervals[1].left)/2
+    
+    max_distance_z = T(world_Δz / 4)
+    max_distance_φ = T(world_Δφ / 4)
+    max_distance_r = T(world_Δr / 4)
+    if !ismissing(max_tick_distance)
+        if max_tick_distance isa Quantity{<:Real, Unitful.𝐋}
```
fhagemann, line 200: You defined `Quantity{<:Real, Unitful.𝐋}` as `length_unit` and I would prefer to use it here.

src/Simulation/Simulation.jl
```diff
@@ -811,19 +734,21 @@ There are serveral `<keyword arguments>` which can be used to tune the computati
...
-    For normal coordinates the unit is meter. For angular coordinates, the unit is radiance.
-    It prevents the refinement to make the grid to fine. Default is [`1e-6`, `1e-6`, `1e-6`].
-- `grid::Grid{T, N, S}`: Initial grid used to start the simulation. Default is `Grid(sim, init_grid_spacing=init_grid_spacing)`.
+- refinement_limits `rl`: Defines the maximum relative (to applied bias voltage) allowed differences 
+    of the potential value of neighboured grid points 
+    in each dimension for each refinement.
+    - `rl::Real` -> One refinement with `l` equal in all 3 dimensions
+    - `rl::Tuple{<:Real,<:Real,<:Real}`` -> One refinement with `l` set individual for each dimension
+    - `rl::Vector{<:Real}`` -> `length(l)` refinements with `l[i]` being the limit for the i-th refinement. 
+    - `rl::Vector{<:Real,<:Real,<:Real}}`` `length(l)` refinements with `l[i]` being the limits for the i-th refinement.
```
fhagemann, line 743: `rl` also after `->`

src/Simulation/Simulation.jl
```diff
@@ -874,19 +780,21 @@ There are serveral `<keyword arguments>` which can be used to tune the computati
...
-    For normal coordinates the unit is meter. For angular coordinates, the unit is radiance.
-    It prevents the refinement to make the grid to fine. Default is [`1e-6`, `1e-6`, `1e-6`].
-- `grid::Grid{T, N, S}`: Initial grid used to start the simulation. Default is `Grid(sim, init_grid_spacing=init_grid_spacing)`.
+- `refinement_limits: Defines the maximum relative (to applied bias voltage) allowed differences 
+    of the potential value of neighboured grid points 
+    in each dimension for each refinement.
+    - rl::Real -> One refinement with `l` equal in all 3 dimensions
+    - rl::Tuple{<:Real,<:Real,<:Real} -> One refinement with `l` set individual for each dimension
+    - rl::Vector{<:Real} -> `length(l)` refinements with `l[i]` being the limit for the i-th refinement. 
+    - rl::Vector{<:Real,<:Real,<:Real}} `length(l)` refinements with `l[i]` being the limits for the i-th refinement.
```
fhagemann, line 789: Same here

src/PotentialSimulation/ScalarPotential.jl
lmh91: Because it is not the same. `PointTypes` for example have `T = UInt8` whereas the grid is usually in e.g. `AT = Float32`.
Its better solved now.

src/PotentialSimulation/ConvergenceAndRefinement.jl
lmh91: Nope. I removed the wrong one.

src/PotentialSimulation/ConvergenceAndRefinement.jl
lmh91: Done.

src/PotentialSimulation/ConvergenceAndRefinement.jl
lmh91: I added methods for `ScalarPotential` to make this more general.

src/Simulation/Simulation.jl
lmh91: Done.

src/Simulation/Simulation.jl
lmh91: Done.

src/Simulation/Simulation.jl
```diff
@@ -681,23 +611,45 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
+                    world_r_mid = (sim.world.intervals[1].right + sim.world.intervals[1].left)/2
+                    min_distance_z = min_distance_r = T(to_internal_units(internal_length_unit, min_tick_distance))
+                    min_distance_r, min_distance_z / world_r_mid, min_distance_z
+                else 
+                    T(to_internal_units(internal_length_unit, min_tick_distance[1])),
+                    T(to_internal_units(internal_angle_unit,  min_tick_distance[2])),
+                    T(to_internal_units(internal_length_unit, min_tick_distance[3]))
+                end 
+            else
+                (T(1e-5), T(1e-5) / (0.25 * grid.axes[1][end]), T(1e-5))
```
fhagemann, line 626: Are these default values documented somewhere (docs?)

src/Simulation/Simulation.jl
```diff
@@ -681,23 +611,45 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
+                if min_tick_distance isa length_unit
+                    min_distance = T(to_internal_units(internal_length_unit, min_tick_distance))
+                    min_distance, min_distance, min_distance
+                else
+                    T(to_internal_units(internal_length_unit, min_tick_distance[1])),
+                    T(to_internal_units(internal_length_unit, min_tick_distance[2])),
+                    T(to_internal_units(internal_length_unit, min_tick_distance[3]))
+                end
+            else
+                (T(1e-5), T(1e-5), T(1e-5))
```
fhagemann, line 639: Same here.

src/Simulation/Simulation.jl
```diff
@@ -964,23 +875,37 @@ function get_signal(sim::Simulation{T, CS}, drift_paths::Vector{EHDriftPath{T}},
...
-    function simulate!(sim::Simulation{T};  max_refinements::Int = 1, verbose::Bool = false,
-                                        depletion_handling::Bool = false, convergence_limit::Real = 1e-5 ) where {T <: SSDFloat}
+    function simulate!(sim::Simulation{T};  refinement_limits = [0.2, 0.1, 0.05],
+                            max_tick_distance::Union{Missing, length_unit, Tuple{length_unit, angle_unit, length_unit}} = missing,
+                            max_distance_ratio::Real = 5,
+                            verbose::Bool = false,
+                            depletion_handling::Bool = false, 
+                            convergence_limit::Real = 1e-5 ) where {T <: SSDFloat}
 
 ToDo...
```
fhagemann, line 885: ?

test/comparison_to_analytic_solutions.jl
```diff
@@ -93,27 +92,26 @@ struct DummyImpurityDensity{T} <: SolidStateDetectors.AbstractImpurityDensity{T}
...
+    idxR1 = searchsortedfirst( sim_car.electric_potential.grid.axes[1], ustrip(R1));
+    rs = sim_car.electric_potential.grid.axes[1].ticks[idxR1:end];
+    idy0 = searchsortedfirst( sim_car.electric_potential.grid.axes[2], 0);
+    potential_rms_car = sqrt(sum((ustrip.(map(r -> potential_analytic(r * u"m"), rs)) .- sim_car.electric_potential.data[idxR1:end, idy0, 1]).^2))/ (ustrip(BV_true)*length(rs));
 
     @testset "Potential RMS" begin
-        @test potential_rms_cyl < 0.01
-        @test potential_rms_car < 1.1
+        @test potential_rms_cyl < 1e-3
+        @test potential_rms_car < 1e-3
```
fhagemann, line 115: Why did this change so drastically?

lmh91: It is now normalized to the bias voltage and number of ticks.

src/Simulation/Simulation.jl
lmh91: Now they are. In the docstrings: b7faa70

fhagemann (approved):

Issue #186, opened 2021-07-22
Painting contacts fails for specific configuration


fhagemann:
When updating the configuration files to avoid deprecation warnings, I stumbled across a possible bug (?) in the painting algorithm.

### Minimum working example

Configuration file: [`public_CGD_config_new.txt`](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/6859422/public_CGD_config_new.txt) --> rename to `public_CGD_config_new.yaml`
(Sorry, but files that end with `.yaml` cannot be uploaded easily here)
```julia
using SolidStateDetectors
T = Float32

sim = Simulation{T}("public_CGD_config_new.yaml")
simulate!(sim, convergence_limit = 1e-6, refinement_limits = [0.2], verbose = false)
evt = Event([CartesianPoint{T}(0,2e-3,0)])
simulate!(evt, sim, Δt = 1e-9, max_nsteps = 10000)
signalsum = sum(broadcast(wf -> abs(wf.value[end]), evt.waveforms))

@info signalsum
@assert isapprox( signalsum, T(2), atol = 1e-3 )
```
It looks like the second contact does not get painted onto the grid.

### Trouble shooting

The problem seems to be this `geometry` in the `difference` of the `contact` with `id: 2`:
```yaml
- box:
    widths: [0.5, 10, 10]
    origin: 
      x: 4.75
```
This does not happen when `widths: [0.6, 10, 10]`, when `òrigin: 4.8`, or when `T = Float64`. 
So, I am assuming that this is connected to some kind of rounding error (?)

The `geometry` in the initial configuration file `public_CGD_config.yaml` also works perfectly and looks like this:
```yaml
- box:
    x: 
      from: 4.5
      to: 5
    y: 10
    z: 10
```

lmh91:
I don't think the issue is the paining. At least not alone. 
Some debugging information I found out so far:

I disabled the painting and compared your config file with the example config file:

```julia
T = Float32
sim1 = Simulation{T}("public_CGD_config_new.yaml")
sim2 = Simulation{T}(SSD_examples[:CGD])
p_det = plot(sim1.detector);

apply_initial_state!(sim1, ElectricPotential, Grid(sim, max_tick_distance = 1u"mm"))
apply_initial_state!(sim2, ElectricPotential, Grid(sim, max_tick_distance = 1u"mm"))

plot(
    plot(sim1.point_types, y = 0, size = (600,600)),
[...]

PR #187, opened 2021-07-22
Bugfix for `paint!`

lmh91:
Regarding #186

### DIFF
--- a/src/PotentialSimulation/Painting/Painting.jl
+++ b/src/PotentialSimulation/Painting/Painting.jl
@@ -56,7 +56,7 @@ function paint!(pointtypes, potential, face::AbstractSurfacePrimitive{T}, geomet
                     widths_ax2[i2], widths_ax2[i2+1],
                     # widths_ax3[i3], widths_ax3[i3+1]
                 )/4
-                if csgtol > Δw_max
+                if csgtol < Δw_max
                     csgtol = Δw_max
                 end
                 if in(pt, geometry, csgtol = csgtol)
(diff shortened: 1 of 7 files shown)


fhagemann:
Can you please post the example how the plots for the file that revealed the bug looks like?

fhagemann:
I ran your code example from #186 and it looks weird (left my file, right the example config file):

after_PR

fhagemann:
The `difference` does not seem to work properly (the inner box is not properly cut out)

after_PR2

fhagemann:
And after refinement, it looks like this:
after_PR3

Are we somehow making sure to not paint the parts of the surfaces that are subtracted?

src/PotentialSimulation/Painting/Painting.jl
```diff
@@ -44,18 +44,30 @@ function paint!(pointtypes, potential, face::AbstractSurfacePrimitive{T}, geomet
     widths_ax1 = diff(get_extended_ticks(grid[1]))
     widths_ax2 = diff(get_extended_ticks(grid[2]))
     widths_ax3 = diff(get_extended_ticks(grid[3]))
+    Δw_max_factor = T(1e-2)
```
fhagemann, line 47: This is just a scaling factor?
I must admit, I haven’t really looked into Δw_mac ...

lmh91: Basically, yes. 
It's a bit tricky to adjust the tolerance in the `in` method for the local grid. 
One parameter, like it is now, is certainly not the perfect solution. 
But it works in most cases.

lmh91:
So now it looks like this:
test

fhagemann (approved):

PR #195, opened 2021-08-05
Improvements and fixes in IO for v0.6

fhagemann:
Two fixes:
- The `Cone` IO was adapted to the cases that are implemented (previously radii of type `Tuple{T, Tuple{T,T}}` were possible if the radii at the top were identical. However, this is not implemented as special case yet)
- Floating passives had a `floating` potential that is not compatible with the type `T <: SSDFloat`. Thus, `:floating` was changed to `NaN`. I added a small entry to the docstring.

For consistency: 
- Change the name of `Grid{T,3,...}` to `CartesianGrid3D{T}` or `CylindricalGrid{T}` everywhere possible.

### DIFF
--- a/src/Config/Config.jl
+++ b/src/Config/Config.jl
@@ -10,18 +10,18 @@
 # For cylindrical grids:
 # 
 # * in(pt::CylindricalPoint{T}, config::UserConfig{T})::Bool where {T <: SSDFloat}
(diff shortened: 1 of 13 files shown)


fhagemann:
We might need to think about restricting the parametric types to only the cases that are implemented, such that things like these can be noticed more easily.

PR #198, opened 2021-08-18
Update `get_crossing_pos` in ChargeDrift.jl with methods from the new CSG

fhagemann:
The function `get_crossing_pos` is used in the charge drift simulation, if the current position is inside the detector and the point after going one step would be out.

Previously, the crossing point was determined by some kind of bisection algorithm (check if the middle point is in or out and bring the two points together "until they meet").

Now, the two points define a `CSG.Line` which intersects either with `Contacts` or the surfaces of the `Semiconductor` that it is leaving. This gets rid of some functions (`point_type()`) and some unnecessary point conversions (in `_is_next_point_in_det`)

Other changes include minor tweaks to the code.


src/ChargeDrift/ChargeDrift.jl
```diff
@@ -66,11 +66,12 @@ function modulate_driftvector(sv::CartesianVector{T}, pt::CartesianPoint{T}, vdv
...
-@inline function _is_next_point_in_det(pt_car::CartesianPoint{T}, pt_cyl::CylindricalPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cylindrical})::Bool where {T <: SSDFloat}
+@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cylindrical})::Bool where {T <: SSDFloat}
+    pt_cyl::CylindricalPoint{T} = CylindricalPoint(pt)
     pt_cyl in point_types || pt_cyl in det
 end
-@inline function _is_next_point_in_det(pt_car::CartesianPoint{T}, pt_cyl::CylindricalPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cartesian})::Bool where {T <: SSDFloat}
-    pt_car in point_types || pt_car in det
+@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cartesian})::Bool where {T <: SSDFloat}
+    pt in point_types || pt in det
 end
```
lmh91, line 75: There might be cases at surfaces where `pt` is just outside the detector but marked as inside as it still lies in the volume of a grid point which `PointType` is marked as semiconductor. 
`PointType` informations are stored bitwise. I think we have some bits left. 
We could add a flag indicating whether all neighboured grid points are also marked as semiconductor points. 
Maybe also including diagonal neighbours due to tilted surfaces.

lmh91 (approved):
Looks good to me 👍

src/ChargeDrift/ChargeDrift.jl
fhagemann: This, in fact, leads to problems in some cases, as then then `Line` connects a point outside (but marked as "inside") and a point outside (marked as "outside") that does not intersect with the detector surface. I'll look into this and make this PR a draft for now.

fhagemann:
`PointTypes` now have an additional bit to check, whether a `pn_junction` point is only surrounded by `pn_junction` points or not.
This new bit (`surface_bit = 0x08`) is true if at least one direct or diagonal neighbor point is **NOT** a `pn_junction` point.
This is checked in `PotentialSimulationSetup`, right after paint. This was used to update `in(pt, point_types)`.

For now, the boundary conditions of the world are not taken into account which might be needed for detectors that end where the world ends. However, the `surface_bit` is only used in the charge drift to check whether the point is still in the detector. And in cases where the charges surpass the dimensions of the world, this should be nothing to worry about.

When plotting `PointTypes`, `surface_bits` are ignored right now. This results in the same plots as before.
Here a plot where one can see the `surface_bits`:

surface_bits

fhagemann:
I changed the target branch for this PR to be 0.7
(because people might need to recalculate their point_types in order to make this work)

However, as the `surface_bit` will be 0 for all previous calculations, it will return the same results as before. This is why we could also push it to the master right now...

src/PotentialSimulation/Painting/Painting.jl
```diff
@@ -210,4 +210,14 @@ function paint!(point_types, potential, face::AbstractSurfacePrimitive{T}, geome
         end
     end 
     nothing
+end
+
+function mark_surface_bits!(point_types::Array{PointType, 3})
+    i1max, i2max, i3max = size(point_types)
+    for i in findall(point_types .& pn_junction_bit .> 0)
+        i1, i2, i3 = i[1], i[2], i[3]
+        point_types[i1,i2,i3] += surface_bit * !(all([point_types[j1,j2,j3] & pn_junction_bit > 0
```
fhagemann, line 219: Here, the `surface_bit` is set for all points with `pn_junction_bit` if not all neighboring points are inside the pn-junction. Changing `+=` to `=` in line 219, we could set the `pn_junction_bit` to 0.

src/Types/point_types.jl
```diff
@@ -52,14 +55,12 @@ end
...
 
 function in(pt::AbstractCoordinatePoint{T}, point_types::PointTypes{T, 3, S})::Bool where {T <: SSDFloat, S}
     cpt = _convert_point(pt, S)
-    grid::Grid{T, 3, S} = point_types.grid
-    i1::Int = searchsortednearest(grid.axes[1].ticks, cpt[1])
-    i2::Int = searchsortednearest(grid.axes[2].ticks, cpt[2])
-    i3::Int = searchsortednearest(grid.axes[3].ticks, cpt[3])
-    return point_types.data[i1, i2, i3] & pn_junction_bit > 0
+    point_type::PointType = point_types[cpt]
+    return (point_type & pn_junction_bit > 0) && (point_type & surface_bit == 0)
```
fhagemann, line 63: Here, only points are "in" the `PointTypes if their `pn_junction_bit` is 1, but the `surface_bit` is 0.

fhagemann (commented):
We could think about setting the `pn_junction_bit` to 0 for all surface points with `surface_bit` being 1. Then, in `in(pt, point_types)`, we could again only check for `pn_junction_bit` being non-zero.

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -66,11 +66,12 @@ function modulate_driftvector(sv::CartesianVector{T}, pt::CartesianPoint{T}, vdv
...
 
-@inline function _is_next_point_in_det(pt_car::CartesianPoint{T}, pt_cyl::CylindricalPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cylindrical})::Bool where {T <: SSDFloat}
+@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cylindrical})::Bool where {T <: SSDFloat}
+    pt_cyl::CylindricalPoint{T} = CylindricalPoint(pt)
     pt_cyl in point_types || pt_cyl in det
 end
-@inline function _is_next_point_in_det(pt_car::CartesianPoint{T}, pt_cyl::CylindricalPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cartesian})::Bool where {T <: SSDFloat}
-    pt_car in point_types || pt_car in det
+@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cartesian})::Bool where {T <: SSDFloat}
+    pt in point_types || pt in det
```
lmh91, line 74: Do we actually have to check here for `pt in det`?
Or would be `pt in det.semiconductor` be enough?

src/PotentialSimulation/PotentialSimulationSetups/PotentialSimulationSetupRBCartesian3D.jl
```diff
@@ -60,6 +60,7 @@ function set_point_types_and_fixed_potentials!(point_types::Array{PointType, N},
             end
         end
     end
+    mark_surface_bits!(point_types)
```
lmh91, line 63: How long does it take to mark a grid (e.g. 300x300x300)?
We only need this for the final grid. So we might want to skip this step for the coarse grids.

src/Types/point_types.jl
lmh91: We could change `surface_bit` to something like `bulk_bit` so that we only have to check one bit here.

fhagemann: That is a very good proposal, I changed this.

src/ChargeDrift/ChargeDrift.jl
fhagemann: `pt in det.semiconductor` should be enough because we only care about if we should continue the drift without any special treatment. Thanks!

src/PotentialSimulation/PotentialSimulationSetups/PotentialSimulationSetupRBCartesian3D.jl
fhagemann: For the example `BEGe` with a grid dimensions `(206, 60, 310)`, it takes around 300ms.

```julia
using SolidStateDetectors
using BenchmarkTools
sim = Simulation(SSD_examples[:BEGe])
calculate_electric_potential!(sim, refinement_limits = [0.2,0.1,0.05,0.02,0.01])
point_types = SolidStateDetectors.get_2π_potential(PointTypes(sim.point_types.data .& 0x07, sim.point_types.grid), n_points_in_φ = 60).data
@btime SolidStateDetectors.mark_bulk_bits!($point_types);
```
```
306.982 ms (1916682 allocations: 249.05 MiB)
```
However, there should be a way to get this allocation-free.
The number of allocations is a bit more than half the number of grid points..

fhagemann: After the latest commit (34ea3a8), the execution time is shorter:
```
133.343 ms (6 allocations: 40.26 MiB)
```

fhagemann:
The determination of `bulk_bits` seems to be free of allocations (and errors) now.

Here is a GIF that shows the `bulk_bits` in yellow for different refinement limits of our example inverted coaxial detector:
jl_PKKeUT

PR #199, opened 2021-08-19
Enhance SOR

lmh91:
Before starting to implement GPU support for the field calculations I had a look again at the current CPU implementation of the 
SOR algorithm. 

The `update!` method, which runs over all grid points (either red or black points), consists out of 3 loops over each axis. 
E.g.:
```julia
for z in z-axis
    for y in y-axis
       for x in x-axis
           update_grid_point(x,y,z)
```
The outer loop is now defined in `outerloop!`. Multithreading is used here (as before).

Prior to this PR, the middle loop and inner loop were defined within one function called in the outer loop. 
Now, the middle loop is defined in `middleloop!` and calls `innerloop!` (and `load_weights_for_innerloop!`) within the middle loop.

By splitting this into functions and actually two middle loops (one over even ticks and one over odd ticks), 
some minor type instability could be removed. 

I splitted the innerloop into two parts. First: `load_weights_for_innerloop!` and second: `innerloop!`.
`load_weights_for_innerloop!` precalculates  the geometrical weights needed for all grid points along the "line" of points 
of the inner loop. E.g. for all grid points in `z` for a given `r` and `phi`.
With the precalulated weights the potential values of those grid points can be updated via SOR efficiently using SIMD operations:
innerloop_vectorization
Vectorization now also works nicely with depletion handling enabled even though depletion handling adds some additional `ifelse` cases to `innerloop!`.
(However, the gain is sadly not that much as calculating the weights takes some time reducing the gain factor of SIMD).

In this PR I also made a small change to the Grid. I force all axes of the grid to have an even number of ticks. 
Not only the inner dimension (due to red black division) and in case of the cylindrical grid the `phi` dimension. 
I did this due to symmetry reasons and it simplifies the application of boundary conditions. 

I performed some benchmark for comparison. 
For this I had to add the just mentioned change of the grid to the current master of SSD:
https://github.com/lmh91/SolidStateDetectors.jl/tree/even_number_of_ticks.
I used the branch as a comparison to the new implementation of the SOR.

SOR_BM_CylindricalGrid_104_34_34

SOR_BM_CylindricalGrid_34_34_34

SOR_BM_CartesianGrid_166_174_174

Especially for larger grids there is some significant gain in performance. 
I should mention though that for the 2D special case in cylindrical coordinates the old implementation is slightly faster (not shown in the plots). However, above grid sizes of around (100, 1, 200) the new implementation becomes faster. 

For smaller grids, e.g. picture 2, the usage of too many threads actually worsens the performance because of threading overhead. 

Note that the difference between no depletion handling and depletion handling enabled became also much smaller as the latter one is now also vectorized. 

---------------------------

Between the cylindrical and Cartesian case we might be able to reduce some code since a lot of it is very similar.
E.g. we might only need one method definition for `innerloop!` and `load_weights_for_innerloop!`. 
I did not do this here, because in the Cartesian case the inner loop runs along the X-axis whereas in the cylindrical case
the inner loop runs along the Z-axis. I would like to also change this in the future such that also in the Cartesian case the
inner loop runs along the Z-axis.


oschulz (approved):
Very nice work!

oschulz:
Btw: We could give [KernelAbstractions](https://github.com/JuliaGPU/KernelAbstractions.jl) a try - it has some Julia core people behind it and brings automatic CPU+GPU support.

lmh91:
> Btw: We could give [KernelAbstractions](https://github.com/JuliaGPU/KernelAbstractions.jl) a try - it has some Julia core people behind it and brings automatic CPU+GPU support.

I will have a look into that.

PR #200, opened 2021-08-19
Bugfix in sampling `Edges`

lmh91:
There was some wrong array indexing. 

It resulted in different Grids (randomly).

### DIFF
--- a/src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
+++ b/src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
@@ -29,7 +29,7 @@ end
 function sample(es::Vector{Edge{T}}; n = 2)::Vector{CartesianPoint{T}} where {T}
     pts = Vector{CartesianPoint{T}}(undef, length(es) * n)
     for i in eachindex(es)
-        pts[i:i+n-1] = sample(es[i], n = n)
+        pts[(i-1)*n+1:i*n] = sample(es[i], n = n)
     end
     pts
 end


lmh91:
This might (hopefully) also fixes the issue of randomly failings tests.

src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
```diff
@@ -29,7 +29,7 @@ end
 function sample(es::Vector{Edge{T}}; n = 2)::Vector{CartesianPoint{T}} where {T}
     pts = Vector{CartesianPoint{T}}(undef, length(es) * n)
     for i in eachindex(es)
-        pts[i:i+n-1] = sample(es[i], n = n)
+        pts[(2i)-1:(2i)] = sample(es[i], n = n)
```
fhagemann, line 32: Shouldn't this somehow depend on `n` (`pts[1 + n*(i-1):n*i]`)?

fhagemann (changes requested):

src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
lmh91: Correct, I'll add this.

lmh91: I forced pushed.

fhagemann (approved):

fhagemann:
Now that we got rid of the randomness, can you have a look at whether we can adjust the tolerances of the test to lower values again?

lmh91:
> Now that we got rid of the randomness, can you have a look at whether we can adjust the tolerances of the test to lower values again?

Yes, we can have a look. But not in this PR.

Issue #107, opened 2020-11-19
Refactor Tests


lmh91:
At the moment, we are simulating quite a number of detectors in our tests. 
We should reduce that.

I guess besides the comparison to analytic solutions we only need to simulate 2 real detectors.
One for cartesian and one for cylindrical coordinates. 

We should add then smaller, more specific tests for the geometry, e.g., primitives. 
But we don't need a full simulation for each primitive.

schustermartin:
I second this.

I think there still should be a test script that basically runs every single feature and function we have to offer. This should be made available in the package but not as  the one that is  run automatically. People can run it locally for somewhat larger PRs to ensure nothing really breaks.

fhagemann:
This is already in the making! 
#122 already has script that tests primitives, transformations, sets and read-in.
Whenever we merge the new CSG with the already existing simulation code, the current and new test scripts can be combined.

PR #201, opened 2021-08-20
Improvements `ssd_write`/`ssd_read` and conversion `Simulation` <-> `NamedTuple`

fhagemann:
Related to #68 and #107.
(To be honest, I'm not exactly sure what #68 is addressing, so it might even be fixed with this PR.)

I added tests to `runtests.jl` for the conversion from `Simulation` to `NamedTuple` and back (2cf6e0b).

This conversion failed when parts of the simulation were `missing`, 
e.g. the `electric_field`, `electron_drift_field`, `hole_drift_field` or any of the `weighting_potentials`.

To fix this:
- the conversion from `Simulation` to `NamedTuple` and back was updated (bfc0aa5).
  This update still allows to "understand" HDF5 files with `Simulations` converted `NamedTuples` since v0.6 and does not come with breaking changes.
- the conversion from `Missing` to `NamedTuple` was updated (1b85a4e) such that it is stored as `String` now. 
  If `Missing` is converted to a `NamedTuple` with value `missing`, [LegendHDF5IO.jl](https://github.com/legend-exp/LegendHDF5IO.jl) will not be able to write it to a HDF5 file. This results in `ssd_write(filename, sim)` failing with an `ArgumentError: Value to write, of type Missing,is not a table`.
- the operator `==` was overridden for most of the structs (0cff6d2) to compare simulations before and after read-in.
  This is used in the tests.

In addition, the methods `ssd_write` and `ssd_read` are now exported (7c65895), have docstrings (f09e71b), and are somewhat explained in the manual (4717677).


fhagemann:
The latest commit (props to @lmh91 for finding out what the problem was) prevents the tests from crashing.
It comes from `Int` overflows on 32bit machines, when big values in the config files are defined as integers.

This could be prevented by adding a `float` statement to [`_parse_value`](https://github.com/fhagemann/SolidStateDetectors.jl/blob/IO/src/ConstructiveSolidGeometry/IO.jl#L60), i.e. changing
```julia
@inline _parse_value(::Type{T}, x::Real, unit::Unitful.Units) where {T} = T(to_internal_units(x * unit))
```
to
```julia
@inline _parse_value(::Type{T}, x::Real, unit::Unitful.Units) where {T} = T(to_internal_units(float(x) * unit))
```

I am not sure whether this might cause rounding error issues. I believe not (`Floats` will be parsed as they were, only `Ints` will be converted to `Floats`), but better be safe than sorry and ask. `T` is always a `Float` anyhow...

fhagemann:
Also, is #68 fixed with this PR?

lmh91:
> Also, is #68 fixed with this PR?

#68 is very general, but I would close it with this PR.

lmh91:
> The latest commit (props to @lmh91 for finding out what the problem was) prevents the tests from crashing.
> It comes from `Int` overflows on 32bit machines, when big values in the config files are defined as integers.
> 
> This could be prevented by adding a `float` statement to [`_parse_value`](https://github.com/fhagemann/SolidStateDetectors.jl/blob/IO/src/ConstructiveSolidGeometry/IO.jl#L60), i.e. changing
> 
> ```julia
> @inline _parse_value(::Type{T}, x::Real, unit::Unitful.Units) where {T} = T(to_internal_units(x * unit))
> ```
> 
> to
> 
> ```julia
> @inline _parse_value(::Type{T}, x::Real, unit::Unitful.Units) where {T} = T(to_internal_units(float(x) * unit))
> ```
> 
> I am not sure whether this might cause rounding error issues. I believe not (`Floats` will be parsed as they were, only `Ints` will be converted to `Floats`), but better be safe than sorry and ask. `T` is always a `Float` anyhow...

Isn't the problem how the `x` is interpreted in the first place?

I think we had this before and i caused numerical issues sometimes. But I am not sure. Also, with the new way we initialize the grid those issues should not matter anymore.

fhagemann:
Regarding my latest commit (4cae7a5):

If the configuration file parser interprets `x` as `Float`, `float(x)` will not do anything.
However, when `x` is an `Int`, the conversion to `float` should not result in any rounding errors.

Only difference is, that the unit conversion for lengths returns a `Rational`, if `x` is an `Int`:
```julia
julia> uconvert(u"m", 25u"mm")
1//40 m

julia> uconvert(u"m", 25.0u"mm")
0.025 m
```

Evaluating the units might not be any problem for `lengths` (where we only multiply powers of 10), and the conversion from `Int` to `Float` will be trivial. For conversion of angles (`deg` <-> `rad`), we have to convert `x` to a `Float` anyhow.

lmh91 (approved):
Thanks :)

PR #202, opened 2021-08-20
Enhance (automated) usage of multi threading in field calculations

lmh91:
Follow-up PR to #199.

Using too many threads for smaller grids actually results in worse performance of the field calculation.

Thus, we can now define the number of (maximal) used threads for each step in the field calculations 
individually by giving a vector to `use_threads`. E.g.:

```julia
calculate_electric_potential!( sim, 
    use_nthreads = [2, 4, 4, 8], # if shorter than `length(refinement_limits)+1`, `maximum(use_nthreads)` is used for further refinements
    refinement_limits = [0.2, 0.1, 0.05, 0.02, 0.01],
)
```

In addition, if only a single value is passed to `use_nthreads`, e.g. `use_nthreads = 16`, it is now interpreted as an upper limit. 
SSD uses `_guess_optimal_number_of_threads_for_SOR`, in each stage of the field calculation, to guess the optimal 
number of threads depending on the current size of the grid.


src/PotentialSimulation/SimulationAlgorithms/SOR.jl
```diff
@@ -24,3 +24,28 @@ end
...
+    elseif n < 400
+        16
+    elseif n < 800
+        32
+    elseif n < 1600
+        64
+    else
+        max_nthreads
+    end
+    nt < max_nthreads ? nt : max_nthreads
```
fhagemann, line 50: Compactify lines 37-50 to:
```nt = min(nextpow(2, max(cld(n, 25),4)), max_nthreads)```

Explanation:
- `cld(n,25) = ceil(n / 25)` defines the cases that you use in the `if`-statements here
- `max( ...., 4)` ensures that `nt` is minimum `4`
- `nextpow(2, nt)` find the nearest power of `2` that is greater than `nt`
- `min( ... ,max_nthreads)` ensures that the return value is smaller than `max_nthreads`

nthreads

File: src/PotentialSimulation/SimulationAlgorithms/SOR.jl [hunk 2]
```diff
@@ -24,3 +24,28 @@ end
     end 
     nothing
 end
+
+
+function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nthreads::Integer, S::Union{Type{Cylindrical}, Type{Cartesian}})::Int
+    max_nthreads = max_nthreads > Base.Threads.nthreads() ? Base.Threads.nthreads() : max_nthreads
+    n = S == Cylindrical ? gs[2] * gs[3] : gs[1] * gs[2] # Number of grid points to be updated in each iteration of the outer loop
```
fhagemann, line 31: Not sure whether this is more performant than the `?`-`:`- syntax but just as an inspiration:
```julia
n = gs[2] * g[1 + 2 * (S==Cylindrical)]
```

src/Simulation/Simulation.jl
```diff
@@ -706,24 +706,6 @@ Takes the current state of `sim.electric_potential` and refines it with respect
 * `sim::Simulation{T}`: [`Simulation`](@ref) for which `sim.electric_potential` will be refined.
 * `max_diffs::Tuple{<:Real,<:Real,<:Real}`: Maximum potential difference between two discrete ticks of `sim.electric_potential.grid` after refinement.
 * `minimum_distances::Tuple{<:Real,<:Real,<:Real}`: Minimum distance (in SI Units) between two discrete ticks of `sim.electric_potential.grid` after refinement.
-
-## Keywords
```
fhagemann, line 727: Why delete the `Keywords` here?
Do you refer to where the explanation for the keywords can be found?

src/Simulation/Simulation.jl
```diff
@@ -862,13 +826,22 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
-            @warn "`use_nthreads` was set to `1`. The environment variable `JULIA_NUM_THREADS` must be set appropriately before the julia session is started."
-        end
         refine = !ismissing(refinement_limits)
         if !(refinement_limits isa Vector) refinement_limits = [refinement_limits] end
         n_refinement_steps = length(refinement_limits)
+
+        max_nthreads, guess_nt = if use_nthreads isa Int 
+            if use_nthreads > Base.Threads.nthreads()
+                use_nthreads = Base.Threads.nthreads();
+                @warn "`use_nthreads` was set to `1`. The environment variable `JULIA_NUM_THREADS` must be set appropriately before the julia session is started."
```
fhagemann, line 836: Why `1`? It is set to `Base.Threads.nthreads()`?

src/Simulation/Simulation.jl
```diff
@@ -862,13 +826,22 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
+            end
+            fill(use_nthreads, n_refinement_steps+1), true
+        else
+            _nt = fill(maximum(use_nthreads), n_refinement_steps+1)
+            _nt[1:length(use_nthreads)] = use_nthreads
```
fhagemann, line 841: What happens if `use_nthreads` is longer than `n_refinement_steps+1`?
This would look like a `BoundsError` to me

src/Simulation/Simulation.jl
```diff
@@ -862,13 +826,22 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
         refine = !ismissing(refinement_limits)
         if !(refinement_limits isa Vector) refinement_limits = [refinement_limits] end
         n_refinement_steps = length(refinement_limits)
+
+        max_nthreads, guess_nt = if use_nthreads isa Int 
+            if use_nthreads > Base.Threads.nthreads()
+                use_nthreads = Base.Threads.nthreads();
+                @warn "`use_nthreads` was set to `1`. The environment variable `JULIA_NUM_THREADS` must be set appropriately before the julia session is started."
+            end
+            fill(use_nthreads, n_refinement_steps+1), true
```
fhagemann, line 838: I guess this will be filled later?

src/Simulation/Simulation.jl
```diff
@@ -1063,7 +1038,9 @@ There are several keyword arguments which can be used to tune the calculation.
         Default is `5`.
 * `grid::Grid`: Initial grid used to start the simulation. Default is `Grid(sim)`.
 * `depletion_handling::Bool`: Enables the handling of undepleted regions. Default is `false`.
-* `use_nthreads::Int`: Number of threads to use in the computation. Default is `Base.Threads.nthreads()`.
+* `use_nthreads::Union{Int, Vector{Int}}`: If `<:Int`, `use_nthreads` defines the maximum number of threads to be used in the computation. 
+    Less thread might be used depending on the current grid size due to threading overhead. Default is `Base.Threads.nthreads()`.
```
fhagemann, line 1042: `thread` -> `threads`

src/Simulation/Simulation.jl
```diff
@@ -928,7 +901,7 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
                                                 n_iterations_between_checks = n_iterations_between_checks,
                                                 max_n_iterations = max_n_iterations,
                                                 depletion_handling = depletion_handling,
-                                                use_nthreads = use_nthreads,
+                                                use_nthreads = guess_nt ? _guess_optimal_number_of_threads_for_SOR(size(sim.electric_potential.grid), max_nthreads[iref+1], CS) : max_nthreads[iref+1],
```
fhagemann, line 904: Will it increase the performance to call  `_guess_optimal_number_of_threads_for_SOR` once at the beginning for all `iref` using broadcast?

src/PotentialSimulation/SimulationAlgorithms/SOR.jl
lmh91: I also don't whether this is more performant. However, this is not a high performant critically code anyway. 
I would keep the current implementation due to better (clearer) readability.

src/Simulation/Simulation.jl
lmh91: Because they are not used in those functions. 

There explanation still exists in the docstrings where they are exists.

src/Simulation/Simulation.jl
lmh91: I'll fix that.

src/Simulation/Simulation.jl
lmh91: `fill(x, n)` creates an vector of length `n` with entries `x`. 

Type stability: `max_nthreads` is otherwise either `Int` or a `Vector{Int}`.
You could still make it type stable but than you produce additional compiled code (methods) for all functions using `max_nthreads`.

src/Simulation/Simulation.jl
lmh91: Correct, we need the size of the grid.

src/PotentialSimulation/SimulationAlgorithms/SOR.jl
```diff
@@ -27,25 +27,26 @@ end
...
-    elseif n < 800
-        32
-    elseif n < 1600
-        64
-    else
-        max_nthreads
-    end
-    nt < max_nthreads ? nt : max_nthreads
+    nt = min(nextpow(2, max(cld(n+1, 25), 4)), max_nthreads)
+    return min(nt, max_nthreads)
```
fhagemann, line 51: Then `min(..., max_nthreads)` is already covered by the line before

fhagemann (changes requested):

src/Simulation/Simulation.jl
```diff
@@ -833,10 +833,13 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         max_nthreads, guess_nt = if use_nthreads isa Int 
             if use_nthreads > Base.Threads.nthreads()
                 use_nthreads = Base.Threads.nthreads();
-                @warn "`use_nthreads` was set to `1`. The environment variable `JULIA_NUM_THREADS` must be set appropriately before the julia session is started."
+                @warn "`use_nthreads` was set to `Base.Threads.nthreads()`. The environment variable `JULIA_NUM_THREADS` must be set appropriately before the julia session is started."
```
fhagemann, line 836: Do you want to print the actual value as well (`$(Base.Threads.nthreads())`) ?

src/PotentialSimulation/SimulationAlgorithms/SOR.jl
lmh91: Ah, right...

fhagemann (approved):

PR #203, opened 2021-08-20
Speed-up in set `point_types`

lmh91:
We do not need to check whether a point is in any contact here.


fhagemann (approved):

fhagemann:
Perfect, thanks so much for your endless efforts!

PR #204, opened 2021-08-24
Change determination of `csgtol` in `paint!`

lmh91:
In this PR the determination/definition of the parameter `csgtol` in the painting of the contacts onto the grid (`paint!`) is changed.

Before, the were sometimes artefacts in cases of coarse (and very unevenly spaced grids). 

This seems to improve this.


fhagemann (approved):
Amazing! This will also fix other issues I encountered ;)

PR #205, opened 2021-08-24
Support `bulletize_PC` keyword in SigGen config files

### LINKED ISSUE #196: Implement `bulletize_PC` for SigGen config files
The field `bulletize_PC` in SigGen config files defines the point contact as spherical rather than as a tube.
This could be implemented using our `Ellipsoid` primitive.

fhagemann:
This PR closes #196. 

SigGen allows the user to choose between contacts with sharp edges or rounded edges (more hemispherical) via the keyword `bulletize_PC`. This is now supported also in SSD.

If the radius is larger than the height, SSD uses the `Torus` primitive with a `Tube` in the center, if the radius is smaller than the height, SSD uses the `Ellipsoid` primitive with a `Tube` underneath.

bulletize

Here are some plots on `PointTypes` and `WeightingPotential`:
siggen_big_r_initial_state
siggen_big_r
siggen_big_r_wp1
siggen_big_r_wp2

Note that SSD does not seem to support true coaxial detector geometries in SigGen. If a height higher than the crystal height is given, SSD still tries to describe this with a "point contact":

siggen_coax

However, semi-coaxial detector geometries are fine.

### DIFF
--- a/docs/src/tutorial_lit.jl
+++ b/docs/src/tutorial_lit.jl
@@ -15,8 +15,7 @@ plot(sim.detector, size = (700, 700))
 
 # One can also have a look at how the initial conditions look like on the grid (its starts with a very coarse grid):
 
-apply_initial_state!(sim, ElectricPotential, 
-    Grid(sim, max_tick_distance = 3u"mm")) # optional
+apply_initial_state!(sim, ElectricPotential) # optional
 plot(
     plot(sim.electric_potential), # initial electric potential (boundary conditions)
     plot(sim.point_types), # map of different point types: fixed point / inside or outside detector volume / depleted/undepleted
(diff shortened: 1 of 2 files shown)


fhagemann:
There are two SigGen keywords in the detector definition that are still not supported: `top_bullet_radius` and `bottom_bullet_radius`.
They would allow describing the top and bottom corners as rounded corners, again using `Torus` primitives. 
If this is needed, this could also be implemented in this PR.

fhagemann:
The tests will not cover what I implemented, but all three cases (`R > Z`, `R = Z`, `R < Z`) look fine.

# `R < Z`
siggen_biggerZ

# `R = Z`
siggen_equalLZ

# `R > Z`

siggen_biggerR

Issue #61, opened 2020-10-09
Charge cloud effects: Diffusion & self-repulsion


lmh91:

lmh91:
#43 is related

fhagemann:
To implement charge cloud effects in SSD, I am thinking about implementing a new scalar fields for the electron and hole densities that evolve in time in `Event` rather than having point-like charges drifting along a drift path.

Find a short proposal on my approach on implementing it [here](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/5535709/Charge_cloud_effects_in_SSD.pdf).

fhagemann:
Diffusion and self-repulsion can also be implementing by extending our current implementation with point-like charges. Then, charge clusters would be simulated as some distribution of point-like charges with weighted charge values.

Find a second proposal for implementing diffusion and self-repulsion 
[here](https://github.com/JuliaPhysics/SolidStateDetectors.jl/files/5590240/Charge_cloud_effects_in_SSD_2.pdf).

fhagemann:
> Diffusion and self-repulsion can also be implemented by extending our current implementation with point-like charges. Then, charge clusters would be simulated as some distribution of point-like charges with weighted charge values.

I have something in the making. However, it is still not (yet) performant and needs some cross-check with analytic solutions to find the corresponding parameters for the diffusion.

Diffusion is implemented by a random walk with the same step length for each point charge. 
Self-repulsion is right now accounted for by calculated the electric field between pairs of charges and adding that to the external field. This requires to calculate the drift velocities for each time step rather than using the interpolated precalculated drift fields.

![Platon](https://user-images.githubusercontent.com/30291312/104365788-bc338080-5518-11eb-9dc2-5540e74c04c5.jpg)

Using the platonic bodies twice around the charge cloud centre as initial point charge distributions results in some promising GIFs:

![Tetraeder2_dr](https://user-images.githubusercontent.com/30291312/104365126-e8023680-5517-11eb-9bb8-0647d18255c0.gif)
![Oktaeder2_dr](https://user-images.githubusercontent.com/30291312/104365841-ce152380-5518-11eb-87f7-b5078a01158a.gif)
![Hexaeder2_dr](https://user-images.githubusercontent.com/30291312/104365868-d5d4c800-5518-11eb-8018-6454332a50d8.gif)
[...]

PR #209, opened 2021-09-01
Implement charge clouds / refactor drift code / separate contributions to waveform

### LINKED ISSUE #64: Generated Signal: Add option to separate electron & hole contribution

fhagemann:
This PR is quite massive, but let me break it down for you!

All of these changes should not be breaking for the users.
The current [legend-julia-tutorial](https://github.com/legend-exp/legend-julia-tutorial) runs through without problems, it only shows a warning when calling `calculate_drift_fields!`.

### Additions to `ChargeCloudModels`
The following things were added:
  - The structs `PointCharge` and `NBodyChargeClouds` (0fad909) which are `AbstractChargeClouds`.
    `NBodyChargeCloud` is just a collection of `AbstractChargeCloud` and stores additional information on the energy distribution and the structure of the charge cloud.
  - Platonic solids `Tetrahedron`, `Hexahedron`, `Octahedron`, `Dodecahedron` and `Icosahedron` (1a01d21) that I proposed in https://github.com/JuliaPhysics/SolidStateDetectors.jl/issues/61#issuecomment-758901706
  - `ParticeTypes` and `Alpha`, `Beta` and `Gamma` (211fbb0) with functions `radius_guess` to estimate the initial size of a charge cloud. Right now, no energy dependence (or anything else) is taken into account.
  - The method `create_charge_cloud` (9714382) which builds an `NBodyChargeCloud` from a given `center` and `energy` of the charge cloud, as well as the `ParticleType` that created it.
  - A method to create an `Event` from an `NBodyChargeCloud` (b077a22).

This is an example, of an `NBodyChargeCloud` that is defined as a center point with two `Icosahedrons` around it:

Icosahedron_initial

  
### Refactor drift code

The drift code needed some new structure (ab149ae) in order to simulate all charge carrier drifts of an `Event` at once (#61).
- The positions of all charge carriers are now stored in a single, two-dimensional `Array` (0c8e735).
- I removed the `electron_drift_field` and `hole_drift_field` from `Simulation` (d829d3d). The electron and hole drift velocities are now calculated separately for each step in the charge drift simulation. Diffusion and self-repulsion, as they are introduced later, will change the drift velocity from what is expected from the precalculated drift fields.
- I removed `calculate_drift_fields!` everywhere. If it is called, it results in a warning that states that it does nothing.
- I renamed `get_interpolated_drift_field` to `interpolated_vectorfield` to be more consistent with `interpolated_scalarfield`.
- I broke down the charge drift code into smaller pieces. At the beginning of the charge drift for one `Event`, an `Array{CartesianVector{T}}` is initialized. In each iteration step:
  - the vectors in this array are set to zero (aac6105),
  - the electric field vector as determined from linear interpolation of `sim.electric_field` is added (e7e82b3),
  - the drift vectors are calculated from the electric field vector (9994f06). and 
  - the resulting drift vectors are modulated (99dd2a8) according to the virtual drift volumes in the detector.
- I added a function to simulate self-repulsion between all charge carriers of the same type (ef198ef). Right now, it is the most basic implementation of Coulomb's law and this process scales with O(N²). It can be activated using the `self_repulsion` keyword that I added to the respective functions (and docstrings).
- I added a function to simulate diffusion (1ebe55f) using random walk. Right now, the step length of a random step is fixed and cannot be changed in a code. I would propose to change this in a future PR.

The functions that simulate diffusion and self-repulsion are currently written such that they return an electric field vector that is added to the interpolated electric field vector **before** determining the step vector. This is done to take the crystal axis effects into account. For self-repulsion, this should be accurate. This might not be the best approach for diffusion, but I'll leave it for now.

Icosahedron_drift

### Separating electron and hole contributions to a waveform

This PR closes #64.
- I added a function `get_electron_and_hole_contribution` (with docstring) (88859fa) that takes an `Event`, `Simulation` and `contact_id` and returns a `NamedTuple` with two fields (`electron_contribution` and `hole_contribution`) that both are `RDWaveforms` with the contributions to the final waveform stored in `evt.waveforms[contact_id]`.
- For convenience, I also added `plot_electron_and_hole_contribution` (4c38862) that creates a plot with the waveforms and the two contributions from electron and holes. This is now added at the very end of our tutorial (35c371a).
- The function `plot_electron_and_hole_contribution` calls `add_baseline_and extend_tail` such that all three `RDWaveforms` are plotted with the same length. I discovered that `add_baseline_and_extend_tail` has a docstring but is not exported, so I exported it (bdc0c02).

Example:
```julia
using Plots; pyplot()
using SolidStateDetectors
T = Float32

simulation = Simulation{T}(SSD_examples[:InvertedCoax])
simulate!(simulation, refinement_limits = [0.2, 0.1, 0.05])
event = Event([
        CylindricalPoint{T}(0.01,deg2rad(10),0.01), 
        CylindricalPoint{T}(0.03,deg2rad(50),0.02),
        CylindricalPoint{T}(0.02,deg2rad(80),0.04)
])
simulate!(event, simulation)

plot_electron_and_hole_contribution(event, simulation, 1, legend = :topleft)
```

Electron_and_hole_contribution


src/ChargeCloudModels/ChargeCloudModels.jl
```diff
@@ -71,4 +71,48 @@ function create_charge_cloud(center::CartesianPoint{T}, energy::RealQuantity, pa
...
+            φ = 2π*n/Mφ
+            Ncount += 1
+            points[Ncount] = CartesianPoint{T}(R*cos(φ)*sin(θ), R*sin(φ)*sin(θ), R*cos(θ))
+        end
+    end
+    
+    return points[1:Ncount]
+end
+
+function create_charge_cloud(center::CartesianPoint{T}, energy::RealQuantity, N::Integer, particle_type::Type{PT} = Gamma;
```
fhagemann, line 99: Now, when calling `create_charge_cloud` with an integer `N` after the energy, e.g. `create_charge_cloud(CartesianPoint{T}(0,0,0), 1460u"keV", 100)`, approximately `N` point-like charges are created on regular spheres. The number of spheres can be defined via `number_of_shells`.

For `N = 100`:

regular_cloud_initial

Creating an `Event` from the `NBodyChargeCloud` and drifting the charges with `diffusion` or `self-repulsion` leads a nice behavior. With this many point-like charges, I would not connect the points in the plot recipe.

Diffusion (`length = 5e-4`):
regular_sphere_diffusion_5e-4

Self-repulsion:
regular_sphere_self_repulsion

Diffusion (`length = 2e-4`) and self-repulsion:
regular_sphere_both_2e-4

src/ChargeCloudModels/ChargeCloudModels.jl
```diff
@@ -1,10 +1,118 @@
...
+energy = 1460u"keV"
+create_charge_cloud(center, energy, number_of_shells = 3, shell_structure = SolidStateDetectors.Icosahedron)
+```
+
+!!! note
+    Using values with units for `energy` requires the package [Unitful.jl](https://github.com/PainterQubits/Unitful.jl).
+
+See also [`NBodyChargeCloud`](@ref).
+"""
+function create_charge_cloud(center::CartesianPoint{T}, energy::RealQuantity, particle_type::Type{PT} = Gamma;
```
lmh91, line 58: This looks like a constructor for `NBodyChargeCloud`.
So I think it should also be named `NBodyChargeCloud`.

lmh91 (commented):
Very nice work :+1:

src/ChargeCloudModels/ChargeCloudModels.jl
fhagemann: I agree! Thanks for pointing that out :smiley:

fhagemann:
Regarding diffusion:
- Right now, the diffusion `length` is a keyword argument of `_add_fieldvector_diffusion!` that cannot be modified from the outermost functions. We would need a way of passing it from the outermost layer (or in the config file) down to the internal functions.
- It would also make sense to include some words (on how to use diffusion as an experimental feature of SSD) in the manual. 

Should this be a part of this PR or a follow-up PR to have this merged for now?

lmh91:
> Regarding diffusion:
> 
> * Right now, the diffusion `length` is a keyword argument of `_add_fieldvector_diffusion!` that cannot be modified from the outermost functions. We would need a way of passing it from the outermost layer (or in the config file) down to the internal functions.
> * It would also make sense to include some words (on how to use diffusion as an experimental feature of SSD) in the manual.
> 
> Should this be a part of this PR or a follow-up PR to have this merged for now?

Hmm, I think in general this is material property, right? 
So I would suggest to add this property to our materials (so that we can dispatch on it). 
It's probably also related to the temperature.
So we can create a function, e.g. `diffusion_length(material::NamedTuple{<MaterialName>}, temperatur)`. 

So we can provide a default fallback method and the user can overwrite
this function in order to try out different things. 

I would suggest to handle this in a follow-up PR.

fhagemann:
Well, we could not only make it temperature-dependent but also field-dependent and determine `length` from the electric field at that point. 


Screenshot from 2021-09-06 17-29-02
Source: [Canali _et al._ (1978)](https://doi.org/10.1016/0026-2714(78)90832-6)

But I agree to postpone this to a follow-up PR. This PR should be good for now.

lmh91:
Than you can add the field strength as an additional argument. It does not need to be used,
but than we don't have to break the syntax later on.

fhagemann:
I would also postpone that to a possible follow-up. Breaking syntax of an internal function is acceptable I would say.. :sweat_smile:

PR #211, opened 2021-09-06
New experimental feature: Calculation of weighting potentials for undepleted detectors

lmh91:
In `calculate_weighting_potential!(sim, contact_id)` the keyword `depletion_handling::Bool` is now actually used.
If `depletion_handling == true` the possible undepleted region of the detector (`sim.point_types`) is used
in the application of the boundary conditions. 

The dielectric permittivity is scaled by a factor in undepleted regions. 
The factor is defined in the function `scaling_factor_for_permittivity_in_undepleted_region(::Semiconductor)`.
So the scaling can be overwritten by overwriting this internal function. 

Here, some example cases for three different scaling factors:

The upper one for an scaling factor of 1 (no scaling):

eps_1

The second one for an scaling factor of 100:

eps_100

The last one for an scaling factor of 100000 which basically mimics perfect conductivity:

eps_100000

### DIFF
--- a/src/PotentialSimulation/PotentialSimulationSetups/PotentialSimulationSetupRBCartesian3D.jl
+++ b/src/PotentialSimulation/PotentialSimulationSetups/PotentialSimulationSetupRBCartesian3D.jl
@@ -78,10 +78,13 @@ function PotentialSimulationSetupRB(det::SolidStateDetector{T}, grid::CartesianG
                 medium::NamedTuple = material_properties[materials["vacuum"]],
                 potential_array::Union{Missing, Array{T, 3}} = missing; 
                 weighting_potential_contact_id::Union{Missing, Int} = missing,
+                point_types = missing,
                 use_nthreads::Int = Base.Threads.nthreads(),
                 sor_consts::T = T(1), not_only_paint_contacts::Bool = true, paint_contacts::Bool = true ) where {T}
-                is_weighting_potential::Bool = !ismissing(weighting_potential_contact_id)
-
+    
+    is_weighting_potential::Bool = !ismissing(weighting_potential_contact_id)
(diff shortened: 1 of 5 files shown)


fhagemann (approved):

src/PotentialSimulation/PotentialSimulationSetups/PotentialSimulationSetupRBCartesian3D.jl
```diff
@@ -185,6 +188,20 @@ function PotentialSimulationSetupRB(det::SolidStateDetector{T}, grid::CartesianG
                 fill_ρ_and_ϵ!(ϵ, ρ_tmp, q_eff_fix_tmp, Cartesian, mpz, mpy, mpx, use_nthreads, passive)
             end
         end
+        if depletion_handling
+            for iz in axes(ϵ, 3)
+                for iy in axes(ϵ, 2)
+                    for ix in axes(ϵ, 1)
+                        pt::CartesianPoint{T} = CartesianPoint{T}(mpx[ix], mpy[iy], mpz[iz])
+                        ig = find_closest_gridpoint(pt, point_types.grid)
+                        if is_undepleted_point_type(point_types.data[ig...])
```
fhagemann, line 197: There is a function `getindex(point_types::PointTypes, pt::AbstractCoordinatePoint)` that does exactly this [on branch v0.7](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/v0.7/src/Types/point_types.jl#L58) (`point_types.data[find_closest_gridpoint(pt, point_types.grid)...]`). We could think about changing this once everything is merged.

lmh91: I will merge when the tests are through. 
Then, we could merge the current master into the v0.7 branch.

Issue #215, opened 2021-09-13
Enhance support for symmetries


lmh91:
Right the user is able to define the world in the config file. 
Symmetries can be used there. For example mirror symmetries. 

The issue is that the symmetries might differ for different things to simulate. 
For example, the example coax detector has in general a periodicity of 60° in φ.
In addition, within this period, it has a mirror symmetry. 
Thus, the electric potential only needs to be calculated on the interval [0°, 30°] (with reflecting boundary conditions in φ).

However, this is not possible for the weighting potentials of individual segments. Those, in this case, have
only a mirror symmetry but not a periodicity. Thus, they are calculated over 360°. 
It could be reduced to [x°, x°+180°]. 

For this we need to add support for defining those symmetries in the configuration files for each individual contact.

fhagemann:
Concerning this: can we handle stuff like `ConeMantle` going from `-30°` to `30°`? The angle in `CylindricalPoint` always goes from `0°``to `360°`, so I guess a point at `350°` will return `false` when checking if it is in `[-30°, 30°]`..

PR #216, opened 2021-09-13
Bugfix: Don't use symmetries for weighting potentials

lmh91:
Crucial bugfix for the calculation of the weighting potentials. 

Bug: The grid for the weighting potential were initialized with the symmetries of the world which are only valid for the
electric potential and not the individual contacts.

### DIFF
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -218,17 +218,12 @@ The grid initialization can be tuned using a set of keyword arguments listed bel
 * `for_weighting_potential::Bool = false`: Grid will be optimized for the calculation of 
     an [`ElectricPotential`](@ref) if set to `true`, and of a [`WeightingPotential`](@ref)
     if set to `false`.
-
-## Additional Keyword for a `CylindricalGrid`
-* `full_2π::Bool = false`: Grid will be extended to `2π` if set to `true` and be left as is
-    if set to `false`.
 """
 function Grid(sim::Simulation{T, Cylindrical};
                 for_weighting_potential::Bool = false,
                 max_tick_distance::Union{Missing, LengthQuantity, Tuple{LengthQuantity, AngleQuantity, LengthQuantity}} = missing,
                 max_distance_ratio::Real = 5,
-                add_points_between_important_point::Bool = true,
-                full_2π::Bool = false)::CylindricalGrid{T} where {T}
+                add_points_between_important_point::Bool = true)::CylindricalGrid{T} where {T}
     det = sim.detector
     world = sim.world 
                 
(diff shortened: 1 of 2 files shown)


fhagemann (changes requested):
The tests still fail. The tolerance needs to be adjusted or the Event needs to be modified.

src/Simulation/Simulation.jl
```diff
@@ -239,7 +239,13 @@ function Grid(sim::Simulation{T, Cylindrical};
 
     world_Δr, world_Δφ, world_Δz = width.(world.intervals)
     world_r_mid = (world.intervals[1].right + world.intervals[1].left)/2
-    
+    if for_weighting_potential && world_Δφ > 0 
+        world_φ_int = SSDInterval{T, :closed, :open, :periodic, :periodic}(0, 2π)
+        world_Δφ = width(world_φ_int)
```
fhagemann, line 244: Do we **always** want it to be full `2π `? I think this makes the `full_2π` keyword obsolete?

We could set `full_2π = true` here...

src/Simulation/Simulation.jl
```diff
@@ -296,9 +302,9 @@ function Grid(sim::Simulation{T, Cylindrical};
     ax_r = even_tick_axis(DiscreteAxis{T, BL, BR}(int_r, important_r_points))
 
     # φ
-    L, R, BL, BR = get_boundary_types(world.intervals[2])
-    int_φ = Interval{L, R, T}(endpoints(world.intervals[2])...)
-    if full_2π || (for_weighting_potential && (world.intervals[2].left != world.intervals[2].right))
+    L, R, BL, BR = get_boundary_types(world_φ_int)
+    int_φ = Interval{L, R, T}(endpoints(world_φ_int)...)
+    if full_2π || (for_weighting_potential && (world_φ_int.left != world_φ_int.right))
```
fhagemann, line 307: .. and then we can check only the statement `if full_2π` here?

fhagemann (commented):
We are ignoring the keyword `full_2π` here. Do we want this?

src/Simulation/Simulation.jl
lmh91: Yes, I think it is obsolete. It's nowhere used. I will remove it.

I would keep it as we did it before. Always calculate the weigting potentials for 2π. 
Only if it is completly phi-symmetric. Than, also only 2D. 

I think we should refactor the symmetry handling, #215, in general. But for now we just need a quick fix.

fhagemann:
I am a bit surprised why the tests for the segBEGe detector did not fail (`@test signalsum == 2`) if the weighting potentials were not calculated the right way and what we can do tr avoid tests passing when things like these happen again.

lmh91:
> I am a bit surprised that the tests for the segBEGe detector did not fail (`@test signalsum == 2`) if the weighting potentials were not calculated the right way and what we can do tr avoid tests passing when things like these happen again.

Because contact 1 lied (at least partially) inside the segment of the world which was covered by the symmetry defined for the electric potential. And also the event was spawned inside this volume. So a sum of 2 was produced. However, the pulse shape is probably also wrong.

PR #223, opened 2021-10-11
Restructure `Event` struct

fhagemann:
In this PR, the `Event` struct and corresponding charge drift functions are slightly modified.
The `locations` and `energies` in `Event` will from now on be `VectorsOfVectors`.
Each entry (= hit) will be a `Vector{CartesianPoint{T}}` (= charge carriers that describe the hit).

Each `Vector{CartesianPoint{T}` is drifted separately. 
Thus, only hits inside one `Vector{CartesianPoint{T}}` will be able to self-repel.

Exaggerated example:
```julia
using SolidStateDetectors
using Plots
using Unitful
T = Float32

simulate!(sim, verbose = false, refinement_limits = [0.2,0.1,0.05])
pos1 = CartesianPoint{T}(0.02,0,0.05); E1 = 1u"GeV"
pos2 = CartesianPoint{T}(0.02,0,0.0505); E2 = 1u"GeV"
```

`Event` with one hit consisting of two charge carriers (that self-repel):
```julia
evt = Event([pos1, pos2], [E1, E2])
simulate!(evt, sim, self_repulsion = true)

plot(evt.drift_paths)
A = plot!(sim.detector)
plot(A, plot(evt.waveforms[1]), size = (1000,500), legend = false)
```
onehit

`Event` with two hits consisting of one charge carrier each (that do not self-repel):
```julia
evt2 = Event([[pos1], [pos2]], [[E1], [E2]])
simulate!(evt2, sim, self_repulsion = false)

plot(evt2.drift_paths)
A = plot!(sim.detector)
plot(A, plot(evt2.waveforms[1]), size = (1000,500), legend = false)
```
twohits

### DIFF
--- a/src/ChargeCloudModels/ChargeCloudModels.jl
+++ b/src/ChargeCloudModels/ChargeCloudModels.jl
@@ -15,12 +15,12 @@ initially distributed around a given center.
 * `SH`: Geometry of the shells.
 
 ## Fields
-* `points::Vector{CartesianPoint{T}}`: Positions of the charge carriers that are part of the charge cloud.
-* `energies::Vector{T}`: Energies of the respective charge carriers, in the same order as `points`.
+* `locations::Vector{CartesianPoint{T}}`: Positions of the charge carriers that are part of the charge cloud.
+* `energies::Vector{T}`: Energies of the respective charge carriers, in the same order as `locations`.
 * `shell_structure::SH`: Initial geometry of the charge carriers around the `center` point, relevant for plotting.
 """
 struct NBodyChargeCloud{T <: SSDFloat, N, SH} <: AbstractChargeCloud
-    points::Vector{CartesianPoint{T}}
+    locations::Vector{CartesianPoint{T}}
     energies::Vector{T} # in units of eV
     shell_structure::SH
 end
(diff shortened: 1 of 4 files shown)


fhagemann:
Now, the question remains how we want to pass the information of an `NBodyChargeCloud` to `Event`
(before or after constructing `Event`, before, during or after calling `simulate!(evt, sim)`, ... ?).

Now it is:
```julia
nbcc = NBodyChargeCloud(pos1, E1, 100, σ = 1u"mm", number_of_shells = 2)
evt = Event(nbcc)
simulate!(evt, sim, self_repulsion = true)
```
I still need to implement that `Event` can take a `Vector{NBodyChargeCloud}`.

We could think of something like this:
```julia
evt = Event([pos1, pos2], [E1, E2], number_of_charges = 100, σ = 1u"mm", number_of_shells = 2)
simulate!(evt, sim, self_repulsion = true)
```
which would make the `NBodyChargeCloud` struct obsolete.

fhagemann:
> Sorry, I didn't had time today to look in detail but I recognized one thing: Shouldn't your first example ("old" syntax) not be converted into the "old" behaviour (independent drift)?
> 
> ```julia
> [pos1, pos2] -> [[pos1], [pos2]] 
> ```

Then, the question is what syntax to use if you want to drift two point-charges together..
Using `[pos1, pos2]` will still result in the "old" behavior if `diffusion = false, self_repulsion = false`.

PR #224, opened 2021-10-15
v0.7

lmh91:
[Documentation](https://juliaphysics.github.io/SolidStateDetectors.jl/previews/PR224/)


lmh91:
So we now can also use the generation of charge clouds via `NBodyChargeCloud` in `simulate_waveforms`.
However, this is really a mess but the roadmap is to refactor the internal code after `v0.8`.

lmh91:
Alright. Is anything missing? Otherwise I would merge.

PR #232, opened 2021-10-21
fix volume primitives double label

hervasa2:
Noticed that on documentation volume primitives had duplicated labels for plots that use `st = :csg`. Fixed.


fhagemann:
They are also missing a symbol in the legend entry. Do we care about this?

lmh91:
> They are also missing a symbol in the legend entry. Do we care about this?

Could you please fix that David? You can probably use the same workaround you did for the mesh, right?

PR #234, opened 2021-10-22
WIP: Sample Surface Plotting

hervasa2:
The `vertices` method can be easily extended to accommodate sampling of each surface. This is then used for the `:samplesurface` seriestype. Please take a look and comment! 
The tricky part is point density. `n_samples` keyword is provided to control point density. Point density is also determined by the scale of the CSG object. This is trivial at the SurfacePrimitive level. But at the CSG level the scale is determined by finding the maximum extremum of the surfaces + norm(origins).  Not always perfect, but there is good visual results. 
Plotting of each CSG takes around 10-80ms
## Coax Example Detector Semiconductor
Screen Shot 2021-10-22 at 5 04 01 PM
## InvertedCoax Example Detector Semiconductor
Screen Shot 2021-10-22 at 5 04 12 PM
## InvertedCoax Example Detector
Screen Shot 2021-10-22 at 5 09 43 PM
s


hervasa2:
On new PR

PR #241, opened 2021-11-04
Plotting docs and defaults

hervasa2:
Updated plotting docs with examples for all seriestypes. 
Made small changes to default values for plotting (eg. linewidths)

### DIFF
--- a/src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
+++ b/src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
@@ -25,7 +25,7 @@ end
                     fillalpha --> 0.2
                     if haskey(plotattributes, :seriestype) 
                         if plotattributes[:seriestype] in [:csg, :wireframe] 
-                            linewidth := 0.5
+                            linewidth := 1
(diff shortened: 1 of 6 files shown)


src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -25,7 +25,7 @@ end
                     fillalpha --> 0.2
                     if haskey(plotattributes, :seriestype) 
                         if plotattributes[:seriestype] in [:csg, :wireframe] 
-                            linewidth := 0.5
+                            linewidth := 1
```
lmh91, line 28: Do we actually want to force the linewidth here? Why not `-->`?

hervasa2: I think it's the best option because if the user changes the linewidth, all linewidths would change if we use -->. As written a user set linewidth will only change the linewidth of positive geometries and will force negative geometries to `linewidth := 1`.

fhagemann (approved):

PR #244, opened 2021-11-16
Fix in plot recipe for `EHDriftPath`

fhagemann:
Based on the newest Plots.jl release (v1.23.6), the plot recipe for `EHDriftPath` results in an error.
This is fixed with this PR. The code shown here also works with older Plots versions.


fhagemann:
So right now, if people get the newest SSD version and the newest plot version, plotting `EHDriftPath` will not work, as the master is not released yet. Do we want to release it (but this comes also with the `PartialTorus` update and the change of the internal representation with `φ` being a `T` instead of `Tuple{T,T}`)?

lmh91:
I would suggest to make a new release.

PR #245, opened 2021-11-20
Adjust conversion of `Geometry` to `Dictionary` for the new definition of `φ` in `Cone` and `Torus`

fhagemann:
With the new implementation of `φ` in `Cone` and `Torus` that is now on the master branch (#243), the conversion of `Geometry` to a `Dictionary` needs to be updated.

Right now, the geometry
```yaml
tube:
  r : 2
  phi:
    from: 30°
    to: 210°
  h: 2
```
will be`parsed as 
```julia
Cone(
  r = 2
  φ = π
  hZ = 1
  origin = [0,0,0]
  rotation = RotZ(π/6)
)
```
When converting this back to a configuration file with `CSG.Dictionary`, this gives
```yaml
tube:
  r : 2
  phi:
    from: 0
    to: 3.141592653589793rad
  h: 2
  rotation:
    Z: 0.5235987755982988rad
```

This PR will increase the readability by
- saving the angles in units of degree (3b92123f0494afe2a6d0e20624ffb93d4a86e4c2)
- adding the offset of `φ` back to the `phi` entry in the config file if the rotation matrix is only a rotation around `Z` (d265da7c8cc0edc2bf04841b8c8043c3b72febde)
- running additional tests to ensure that the conversion between `Geometry` and `Dictionary` is done correctly (118dfb3f7dcb0738604ef6101aa532a43e524157)

resulting in
```yaml
tube:
  r : 2
  phi:
    from: 30°
    to: 210°
  h: 2
```
However, there might be rounding errors, so the initial and final dictionary are only identical within the type precision.

The `Dictionary` function is used when converting config files from v0.5 or before to v0.6 or after.
I think it would be beneficial to have the updated config file as readable and editable as possible.

After this, I would propose a new release v0.7.3.

### DIFF
--- a/src/ConstructiveSolidGeometry/Transformations.jl
+++ b/src/ConstructiveSolidGeometry/Transformations.jl
@@ -31,7 +31,7 @@ function Dictionary(m::SMatrix{3,3,T,9}) where {T}
     mat = RotXYZ(m)
     if mat.theta1 == 0 && mat.theta2 == 0
         if mat.theta3 != 0
-            dict["Z"] = string(mat.theta3)*"rad"
+            dict["Z"] = string(rad2deg(mat.theta3))*"°"
         end
     else
         dict["M"] = m[:]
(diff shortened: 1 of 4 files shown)


lmh91:
Looks good to me. 👍

PR #250, opened 2022-01-18
Add function to determine the depletion voltage

### LINKED ISSUE #193: Add function to determine depletion voltage
Something like `get_depletion_voltage(sim)` that takes the `Simulation`, performs some kind of calculation and returns the depletion voltage in V.

fhagemann:
This PR would close #193.
I wrote a method `get_depletion_voltage` to estimate the potential needed at a given contact to fully deplete the detector.

## How does it work?
The electric potential inside a detector (`ϕ`) is given by the contribution from the ionized impurities in the semiconductor when all contacts are grounded (`ϕρ`) and the contribution from the potentials applied to the contacts of the detector (`ϕV`).

ElectricPotentialContributions

Mathematically, the contribution from the potentials applied to the contacts corresponds to the `WeightingPotential` of the contacts times the potentials applied to the contacts. Thus, `ϕρ` can be determined by subtracting the `WeightingPotentials` times the potentials applied to the contacts from the `ElectricPotential`.

The detector is depleted when no point in the detector has a potential value outside of the range given by the potentials applied to the contacts. In the picture above, `ϕρ` has its maximum in the center of the bulk at around 1100V.

By applying a potential (in this case to the point contact at the top of the detector), `ϕV` can be varied such that the final `ϕ` does not reach its maximum in the bulk of the detector but on the point contact. It is then depleted.
VoltageScan

The idea of `get_depletion_voltage` is to find the "first" potential, at which the minimum and maximum values of the resulting `ElectricPotential` are reached on the contacts and not in the bulk of the detector.

## How is it implemented?

This method only works, if the original `ElectricPotential` is calculated for a depleted detector. Thus, it is first checked whether the original detector setup is depleted.

Then, all the `WeightingPotentials` of the contacts, at which the potentials are non-zero, are interpolated to the `Grid` of the already calculated `ElectricPotential` (via `_adapt_weighting_potential_to_electric_potential_grid!`). By not only interpolating but also updating the resulting `WeightingPotential` until convergence, the interpolation errors are reduced.

`ϕρ` is calculated by subtracting the adapted `WeightingPotentials` times the potential of the contact from the `ElectricPotential`.
Then, the extrema values for all grid points in the semiconductor, `ϕmin` and  `ϕmax`, are determined for the initial configuration, from which can be deducted if the setup is initially depleted or not (`ìnitial_depletion`).

Then, a loop iterates through all voltages given by the `potential_range`. For each potential value, `ϕmin` and  `ϕmax` are again calculated which determine whether the detector is depleted, ie. if  `ϕmax - ϕmin` is smaller than the applied bias voltage.
If the `initial_depletion` does not match the current depletion behavior, we have reached the depletion voltage which is, then, returned by the function. If  the `initial_depletion` does not change throughout the given potential range, this means that the depletion voltage most probably is not within that range.

## Things to take into account

- `get_depletion_voltage` overwrites the `WeightingPotentials` of the `Simulation` when they are adapted (which leaves them with a new grid). Is that a problem?
- I did not add any tests for this.

### DIFF
--- a/docs/src/tutorial_lit.jl
+++ b/docs/src/tutorial_lit.jl
@@ -90,6 +90,11 @@ is_depleted(sim_undep.point_types)
 println("Depleted:   ", get_active_volume(sim.point_types))
 println("Undepleted: ", get_active_volume(sim_undep.point_types));
 
+# The depletion voltage can also be estimated from the simulation with the depleted detector.
+# For this, the user has to choose a contact to which to apply the voltage, as well as a range to probe:
+
+get_depletion_voltage(sim, 2, 1700:0.5:2000)
+
 
 # ## Electric field calculation
 
--- a/src/Simulation/DepletionVoltage.jl
+++ b/src/Simulation/DepletionVoltage.jl
@@ -1,0 +1,105 @@
+# """
+#     _adapt_weighting_potential_to_electric_potential_grid!(sim::Simulation, contact_id::Int)
+# 
+# Interpolates the [`WeightingPotetial`](@ref) of the [`Contact`](@ref) with id `contact_id`
+# onto the [`Grid`](@ref) of the [`ElectricPotential`](@ref) and updates it until it converges.
+# 
+# ## Arguments 
+# * `sim::Simulation{T}`: [`Simulation`](@ref) for which the [`WeightingPotential`] should be adapted.
+# * `contact_id::Int`: The `id` of the [`Contact`](@ref) at which the [`WeightingPotential`] should be adapted.
+# """
+function _adapt_weighting_potential_to_electric_potential_grid!(sim::Simulation, contact_id::Int)
+    @assert !ismissing(sim.electric_potential) "Electric potential missing"
+    if ismissing(sim.weighting_potentials[contact_id]) calculate_weighting_potential!(sim, contact_id, verbose = false) end
+    if sim.weighting_potentials[contact_id].grid != sim.electric_potential.grid
+        sim.weighting_potentials[contact_id] = sim.weighting_potentials[contact_id][sim.electric_potential.grid]
+        #apply_initial_state!(sim, WeightingPotential, contact_id, sim.electric_potential.grid)
+        update_till_convergence!(sim, WeightingPotential, contact_id)
+    end
+end
+
+"""
+    get_depletion_voltage( sim::Simulation{T}, contact_id::Int, potential_range::AbstractRange; kwargs... )::T
+
+Returns the potential (in V) needed at the [`Contact`](@ref) with id `contact_id`
+to fully deplete the detector in a given [`Simulation`](@ref). For this, all other
+contact potentials are set to `0` and the potential at the specified contact is
+increased or decreased according to the `potential_range`. The depletion voltage
+is set to the potential for which a previously undepleted detector becomes depleted,
+or for which a previously depleted detector becomes undepleted.
+
+## Arguments 
+* `sim::Simulation{T}`: [`Simulation`](@ref) for which the depletion voltage should be determined.
+* `contact_id::Int`: The `id` of the [`Contact`](@ref) at which the potential is applied.
+* `potential_range::AbstractRange`: Range of potentials to be tested.
+    
+## Keywords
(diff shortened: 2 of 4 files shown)


lmh91:
Thanks for starting this!

I would rename the function to `calculate_depletion_voltgate` to go along with all the other functions like `calculate_electric_potential`, `calculate_capacitance_matrix`, ...

I don't think this statement is correct (strong enough) though:
> The detector is depleted when no point in the detector has a potential value outside of the range given by the potentials applied to the contacts.

E.g., this condition is not enough for detectors which have an undepleted region not connected to any contact (a so-called bubble) => Looking for the global maxima/minima is not enough.

I would also leave out the weighting potentials here. Because there can also be detectors which have segmentation in 
both kind of contacts, n+ and p+. Thus, we should calculate the electric potential for two cases (almost the same as what you have described):
1) Set all contacts to `0V` and turn `depletion_handling` off to calculate the field of the impurity density `ϕρ`.
2) Set the impurity density to 0 and calculate the electric potential `ϕV` for an bias voltage of 1.

Then, find the minimum `DV` (depletion voltage) where the potential `ϕV * DV + ϕρ` changes monotonically 
from all `n+` to `p+` contacts along all ways. I think that can be done by iterating one time over all grid points 
when we have the two potentials `ϕV` and `ϕρ` on the same grid through your function `_adapt_weighting_potential_to_electric_potential_grid!`. By this we also exclude local maxima/minima.

We should not overwrite any calculated fields of the `Simulation`. 
We could add two more (optional) fields to `Simulation` for those two special fields.

fhagemann:
Thanks for the feedback!

> E.g., this condition is not enough for detectors which have an undepleted region not connected to any contact (a so-called bubble) => Looking for the global maxima/minima is not enough.

I see your point, would be enough to also make sure that there are no local maxima and minima, which could exclude bubbles (we have the `surface_bit`, so we could demand that no "`bulk_bit`" is surrounded by points with potential values that are **ALL** smaller or **ALL** larger?) Or calculate the electric field from this and checking that it should be greater than a given threshold?

> I would also leave out the weighting potentials here. Because there can also be detectors which have segmentation in 
both kind of contacts, n+ and p+.

I don't think that we differentiate between n+ and p+ contacts in the simulation, at least from what I see... If you want to apply the potential to multiple contacts, one could think about making the argument `contact_id` an Array (which can consist of mutliple contacts for the segmented case)

> 1) Set all contacts to `0V` and turn `depletion_handling` off to calculate the field of the impurity density `ϕρ`.
> 2) Set the impurity density to 0 and calculate the electric potential `ϕV` for an bias voltage of 1.

In your list, point 2 to me looks exactly what the weighting potential is..

> We should not overwrite any calculated fields of the `Simulation`.
> We could add two more (optional) fields to `Simulation` for those two special fields.

I don't think that we need extra fields. As this function will not be essential for the simulation, we can just `deepcopy` the needed potentials in the function and free the memory afterward.

lmh91:
> I see your point, would be enough to also make sure that there are no local maxima and minima, which could exclude bubbles (we have the surface_bit, so we could demand that no "bulk_bit" is surrounded by points with potential values that are ALL smaller or ALL larger?) 

I think that this is a sufficient condition. 
Iterate over all grid points and determine locally the scale factor (`sf[i]`) of `ϕV` such that 
`sf[i] * ϕV[i_neighbour_min] + ϕρ[i_neighbour_min] <=  sf[i]* ϕV[i] + ϕρ[i] <= sf[i] * ϕV[i_neighbour_max] + ϕρ[i_neighbour_max]`.
Then, the depletion voltage should be `minimum(sf[i]) * V`.

> In your list, point 2 to me looks exactly what the weighting potential is..

Only when the weighting potential you are using is the only contact where the bias voltage is applied. I know this is for most cases the case, but not in general. Of course we can add functionally such that the user can specify to take this particular weighting potential as `ϕV`. 

___________________

General note about depletion handling: If static charge distribution are present this all becomes unclear.
Also in `calculate_electric_potential!` with `depletion_handling = true`.

fhagemann:
Closing this in favor of #311.

PR #254, opened 2022-02-03
Add pseudo geometry for tests

SebastianRuffert:
A pseudo detector containing all volume primitives, boolean operators and transformations will be used for tests. Before, several geometries had to be calculated individually.


SebastianRuffert:
```
using Plots
using SolidStateDetectors
sim = Simulation(SSD_examples[:TestDetector])
plot(sim.detector, size = (700, 700), show_semiconductor = true)
```
Detector

Contacts in the middle are biased at 0V

PR #255, opened 2022-02-14
Pseudo geometry

SebastianRuffert:
Implement Pseudo Geometry for Tests


SebastianRuffert:
image
image


Pseudo Geometry

SebastianRuffert:
image
Issue with the electron drift path. Electron leaves semiconductor at the border of the partial torus.

SebastianRuffert:
image

Add torus contact to prevent charges from getting lost inside the torus

SebastianRuffert:
After discussing, we have agreed to test all primitives separately instead of testing a pseudo geometry consisting of all geometries.

Issue #256, opened 2022-02-15
`:samplesurface` does not work when important points are not in the final geometry


fhagemann:
The function `get_scale` that scales the distance between points for `:samplesurface` plotting does not work for geometries, for which the important points are not in the final geometry (excluded in `CSGDifference` or `CSGIntersection`).

## Minimum working example

Detector configuration file
```yaml
units:
  length: mm
  angle: deg
grid: 
  coordinates: cartesian
  axes:
    x: 
      from: -3
      to: 3
    y:
      from: -3
      to: 3
    z: 
      from: -3
      to: 3
medium: vacuum
detectors:
- semiconductor:
    material: HPGe
    geometry:
      intersection:
        - box: 
            widths: [2,2,2]
        - sphere:
            r: 1.41421356237
  contacts:
    - id: 1
      potential: 100
      geometry:
        tube:
          r: 1
          h: 0
          origin: [0,0,1]
    - id: 2
      potential: 0
      geometry:
        tube:
          r: 1
          h: 0
          origin: [0,0,-1]
```

Plotting the detector works for `:csg`, `:mesh3d` and `:wireframe`:
```julia
using SolidStateDetectors
using Plots; gr(fmt = :png)
sim = Simulation("MWEsamplesurface.yaml");
plot(sim.detector.semiconductor, size = (500,500), st = :mesh3d, color = :lightgray, lw = 0.5)
plot!(sim.detector)
```
![csg](https://user-images.githubusercontent.com/30291312/154164421-b68cf8f9-919d-4087-8850-5dfc69845d29.png)

but fails for `:samplesurface`:
```julia
plot(sim.detector, show_semiconductor = true, st = :samplesurface, size = (500,500), n_samples = 50)
```

```
ArgumentError: reducing over an empty collection is not allowed

Stacktrace:
  [1] _empty_reduce_error()
    @ Base ./reduce.jl:301
  [2] reduce_empty(op::Function, #unused#::Type{Float32})
    @ Base ./reduce.jl:311
  [3] mapreduce_empty(#unused#::typeof(identity), op::Function, T::Type)
    @ Base ./reduce.jl:345
  [4] reduce_empty(op::Base.MappingRF{typeof(identity), typeof(max)}, #unused#::Type{Float32})
    @ Base ./reduce.jl:331
  [5] reduce_empty_iter
    @ ./reduce.jl:357 [inlined]
[...]

PR #258, opened 2022-02-16
Bugfixes for cases in which important points are not in the final geometry

fhagemann:
This PR fixes the errors described in #256.

### DIFF
--- a/src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
+++ b/src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
@@ -6,10 +6,12 @@ function primitives(csg::AbstractConstructiveGeometry)
     ps
 end
 
-@recipe function f(csg::AbstractConstructiveGeometry{T}; n_samples = 40) where {T}
+@recipe function f(csg::AbstractConstructiveGeometry{T}; n_samples = 40, CSG_scale = missing) where {T}
     seriestype --> :csg
     ps = primitives(csg)
-    spacing = (haskey(plotattributes, :seriestype) && plotattributes[:seriestype] == :samplesurface) ?  T(get_scale(csg)/n_samples) : nothing
+    if haskey(plotattributes, :seriestype) && plotattributes[:seriestype] == :samplesurface
+        spacing::T = T((ismissing(CSG_scale) ? get_scale(csg) : CSG_scale)/n_samples)
+    end
     @series begin
         label --> "CSG"
         if haskey(plotattributes, :seriestype) 
--- a/src/ConstructiveSolidGeometry/CSG.jl
+++ b/src/ConstructiveSolidGeometry/CSG.jl
@@ -219,6 +219,10 @@ sample(csg::AbstractConstructiveGeometry{T}, sampling...) where {T} = vcat(sampl
 function get_scale(csg::AbstractConstructiveGeometry{T}) where {T}
     points = [point for p in primitives(csg) for s in surfaces(p) for point in vertices(s, 4)]
     filter!(p -> in(p,csg), points)
+    if isempty(points) 
+        @warn "The number of samples might be smaller than expected. Consider increasing `n_samples` to obtain the desired result."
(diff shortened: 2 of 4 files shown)


hervasa2 (approved):
Looks good! Thank you

fhagemann:
The only thing is that the warning appears twice when creating this plot
(apparently `get_scale` is called twice):
samplesurface

Is this intended? @hervasa2

hervasa2:
> The only thing is that the warning appears twice when creating this plot (apparently `get_scale` is called twice):
> 
> ```julia
> using SolidStateDetectors
> using Plots; gr(fmt = :png)
> sim = Simulation("MWEsamplesurface.yaml");
> plot(sim.detector.semiconductor, size = (500,500), st = :mesh3d, color = :lightgray, lw = 0.5)
> plot(sim.detector, show_semiconductor = true, st = :samplesurface, size = (500,500), n_samples = 50)
> ```
> 
> samplesurface
> 
> Is this intended? @hervasa2

`get_scale` gets called for each csg object so this is expected when plotting contacts.

fhagemann:
> `get_scale` gets called for each csg object so this is expected when plotting contacts.

The thing is that the only CSG object here is the semiconductor. If I only plot the contacts (both are just simple tubes, no CSG), I don’t get the warning at all.

lmh91:
> The thing is that the only CSG object here is the semiconductor. If I only plot the contacts (both are just simple tubes, no CSG), I don’t get the warning at all.
> 
> It also looks like I'm plotting more points for the contacts by just "removing" the semiconductor:

In the plot recipe where the `get_scaled` is actually called one would need to get the scales for all objects, not just the ones which are to be plotted. I think it would be okay to do so.

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -6,10 +6,12 @@ function primitives(csg::AbstractConstructiveGeometry)
     ps
 end
 
-@recipe function f(csg::AbstractConstructiveGeometry{T}; n_samples = 40) where {T}
+@recipe function f(csg::AbstractConstructiveGeometry{T}; n_samples = 40, CSG_scale = missing) where {T}
     seriestype --> :csg
     ps = primitives(csg)
-    spacing = (haskey(plotattributes, :seriestype) && plotattributes[:seriestype] == :samplesurface) ?  T(get_scale(csg)/n_samples) : nothing
+    if haskey(plotattributes, :seriestype) && plotattributes[:seriestype] == :samplesurface
+        spacing::T = T((ismissing(CSG_scale) ? get_scale(csg) : CSG_scale)/n_samples)
```
fhagemann, line 13: Now, the scales are only determined once for a detector and passed to the plot recipes of the individual geometries.
However, if the geometries/objects are plotted by themselves, the scales are determined within that

fhagemann (commented):
The scales are now only determined once when plotting.

Examples:
```julia
plot(sim.detector, show_semiconductor = true, n_samples = 50, st = :samplesurface, ratio = 1, size = (500,500))
```
```
┌ Warning: The number of samples might be smaller than expected. Consider increasing `n_samples` to obtain the desired result.
└ @ SolidStateDetectors.ConstructiveSolidGeometry /home/mpp/Software/SolidStateDetectors.jl/src/ConstructiveSolidGeometry/CSG.jl:223
```
detector

```julia
plot(sim.detector.semiconductor, n_samples = 50, st = :samplesurface, ratio = 1, size = (500,500))
```
```
┌ Warning: The number of samples might be smaller than expected. Consider increasing `n_samples` to obtain the desired result.
└ @ SolidStateDetectors.ConstructiveSolidGeometry /home/mpp/Software/SolidStateDetectors.jl/src/ConstructiveSolidGeometry/CSG.jl:223
```
semiconductor

```julia
plot(sim.detector, n_samples = 50, st = :samplesurface, ratio = 1, size = (500,500))
```
contacts

PR #261, opened 2022-02-21
Unify SOR-convergence check (CPU&GPU) and improve depletion handling

lmh91:
Alright, this PR unifies the convergence of CPU and GPU.
It also changes how the depletion is handled.
Please have a look @oschulz @fhagemann @schustermartin @hervasa2

# Boolean Issue in Depletion Handling

Prior to this PR, a grid point was either depleted or undepleted. This information was stored bitwise in the point type bit of the grid point. This had a big issue: The convergence was difficult to check because grid points, especially at the edge of the depleted zone, were constantly flipping between those to states which resulted always in a large change in potential of this grid point. This increased also the runtime of the field calculation quite a bit as the convergence was hard to determine. 
The flipping also resulted in some points being marked depleted even as they should be marked as undepleted:
old

Therefore, in the past, I did added some special treatments for these issues, 
which improved the situation but were far from perfect. 

# Impurity Scale

Now, this special treatment is not necessary anymore as I introduced a new scalar field to the calculation: 
`imp_scale` for impurity scale. The values of `imp_scale` are in `[0, 1]`. It controls how much of the charge of the impurity density of a grid point is taken into account. So now, grid points can also be treated as partially depleted, which makes sense
for grid points at the edge of a depleted zone. 
The impurity scale of a grid points is getting updated in the SOR such that it converges against some value. 
Now, the convergence check do not need a special treatment anymore and works quite nicely.
And, the field calculation becomes much faster. 
At the very end, after the SOR, the points types are updated once: Points where `imp_scale < 1` are marked as undepleted. 
Here the respective plot for comparison to the one above:
new

The code to reproduce the plots:
```julia
using Plots, SolidStateDetectors
sim.detector = SolidStateDetector(sim.detector, contact_id = 2, contact_potential = 500); # V  <-- Bias Voltage of Mantle
isNDH = hasfield(typeof(sim), :imp_scale) 
fn = isNDH ? "new" : "old"

calculate_electric_potential!(
    sim,
    depletion_handling = true,
    convergence_limit = 1e-6,
    refinement_limits = [0.2, 0.1, 0.05, 0.025],
    verbose = true,
    max_n_iterations = 50000,
    use_nthreads = 1
)
if isNDH
    plot(
        plot(sim.electric_potential),
        plot(sim.point_types),
        plot(sim.imp_scale),
        title = fn,
        layout = (1, 3), 
        size = (1500, 700)
    )
else
    plot(
        plot(sim.electric_potential),
        plot(sim.point_types),
        title = fn,
        layout = (1, 2), 
        size = (900, 700)
    )
end
```


src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCylindrical.jl
```diff
@@ -81,7 +81,8 @@ end
 
 function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CylindricalGrid{T}, 
                 medium::NamedTuple = material_properties[materials["vacuum"]],
-                potential_array::Union{Missing, Array{T, 3}} = missing; 
+                potential_array::Union{Missing, Array{T, 3}} = missing,
+                imp_scale::Union{Missing, ElectricPotential{T, 3}} = missing;
```
fhagemann, line 85: Is there a reason not to pass `imp_scale` as `Array{T,3}` like you do for the `potential_array`.
It looks like you only access the `data` field and I guess the `grid` is the same as in `point_types`..

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCylindrical.jl
```diff
@@ -372,14 +371,20 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::Cylindrical
                 paint_contacts = Val(paint_contacts)  )
         rbpotential  = RBExtBy2Array( potential, grid )
         rbpoint_types = RBExtBy2Array( point_types, grid )
+
+        imp_scale = ismissing(imp_scale) || is_weighting_potential ? zeros(T, size(q_eff_imp)) : RBExtBy2Array(imp_scale.data, grid)
+        for i in eachindex(imp_scale)
+            imp_scale[i] = is_pn_junction_point_type(rbpoint_types[i]) ? one(T) : zero(T)
```
fhagemann, line 377: Same thing here: looks like `statement ? true : false`

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_update.jl
```diff
@@ -0,0 +1,33 @@
+"""
+    function update!(   
+        pssrb::PotentialCalculationSetup{T};
```
fhagemann, line 3: I thought this was renamed to `pcs` ?

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_update.jl
```diff
@@ -0,0 +1,33 @@
+"""
+    function update!(   
+        pssrb::PotentialCalculationSetup{T}; 
+        ::Nothing, # these to unused arguments are used such that the method
```
fhagemann, line 4: Typo: two

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_update.jl
```diff
@@ -0,0 +1,33 @@
...
+        is_weighting_potential::Val{_is_weighting_potential} = Val{false}(),
+        only2d::Val{only_2d} = Val{false}()
+)::Nothing where {T, depletion_handling_enabled, only_2d, _is_weighting_potential}
+
+This function performs one iteration of the SOR. One iteration consists out of 4 steps:
+
+    1) Iterate in parallel over all even points and update their potential. 
+    2) Apply the boundary conditions at the ends of the grid for all even points. 
+    3) Iterate in parallel over all odd points and update their potential. 
+    2) Apply the boundary conditions at the ends of the grid for all odd points.
```
fhagemann, line 17: Typo: 4)

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_update.jl
```diff
@@ -0,0 +1,33 @@
...
+This function performs one iteration of the SOR. One iteration consists out of 4 steps:
+
+    1) Iterate in parallel over all even points and update their potential. 
+    2) Apply the boundary conditions at the ends of the grid for all even points. 
+    3) Iterate in parallel over all odd points and update their potential. 
+    2) Apply the boundary conditions at the ends of the grid for all odd points. 
+"""
+@inline function update!(   
+    pssrb::PotentialCalculationSetup{T},
+    ::Nothing, # these to unused arguments are used such that the method
```
fhagemann, line 21: Typo: two

src/PotentialCalculation/SuccessiveOverRelaxation/GPU_kernel_Cartesian3D.jl
```diff
@@ -1,20 +1,22 @@
...
-    q_eff_fix::AbstractArray{T, 4},
-    ϵ_r::AbstractArray{T, 3},
-    geom_weights::NTuple{3, <:AbstractArray{T, 2}},
-    sor_const::AbstractArray{T, 1},
-    update_even_points::Bool,
-    depletion_handling_enabled::Bool,
-    is_weighting_potential::Bool,
-    only2d::Bool
+    imp_scale::AbstractArray{T, 4},
+    @Const(point_types::AbstractArray{PointType, 4}),
```
fhagemann, line 4: What does `@Const` do? Not modify the array within the function?

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -42,22 +42,23 @@ end
...
-        point_type::PointType, 
+        imp_scale::T,
         neighbor_potentials::NTuple{6,T}, 
         q_eff_imp::T, 
         volume_weight::T,
         sor_const::T
     )::Tuple{T, PointType} where {T}
 
-This function checks whether the current grid point is depleted or undepleted.
+This function handles the grid points which volume is not fully depleted.
```
fhagemann, line 52: Reads weird:
 .. grid points with volumes which are not fully depleted. (?)

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,58 @@ Thus, there cannot be local extrema.
...
     neighbor_potentials::NTuple{6,T}, 
     q_eff_imp::T, 
     volume_weight::T,
     sor_const::T
-)::Tuple{T, PointType} where {T}
+)::Tuple{T, T} where {T}
     vmin::T = min(neighbor_potentials...)
     vmax::T = max(neighbor_potentials...)
+    
+    isLess = new_potential < vmin
```
fhagemann, line 84: Do we have variables named like this (uppercase letters in between)? Why not `is_less`?

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,58 @@ Thus, there cannot be local extrema.
...
 
-    if new_potential < vmin || new_potential > vmax
-        new_potential -= q_eff_imp * volume_weight * sor_const
-        if (point_type & undepleted_bit == 0 && point_type & pn_junction_bit > 0) 
-            point_type += undepleted_bit
+    Δnp_ext = new_potential - neighbor_relevant_extremum
+    imp_contribution_fix = q_eff_imp * volume_weight * sor_const 
+    imp_contribution_current = imp_contribution_fix * imp_scale
+    
+    if isLess || new_potential > vmax
```
fhagemann, line 91: why have `isLess` but not `isMore`?

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,58 @@ Thus, there cannot be local extrema.
...
+            new_imp_scale::T = x / imp_contribution_fix
+            if new_imp_scale > 1
+                new_imp_scale = 1
+            end
+            imp_scale::T = new_imp_scale
         end
-    elseif point_type & undepleted_bit > 0
-        point_type -= undepleted_bit
+    else
+        imp_scale *= 1.1 # Just chosen. Seems to work. There might be a better solution.
```
fhagemann, line 102: Please elaborate on this.

fhagemann (commented):
I had a short glimpse, find my comments here.

I didn't test your code: when plotting `sim.q_eff_imp`, will that still be the "full" effective charge or will that also be scaled by `imp_scale`?

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -290,14 +291,20 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                 paint_contacts = Val(paint_contacts)  )
         rbpotential = RBExtBy2Array( potential, grid )
         rbpoint_types = RBExtBy2Array( point_types, grid )
+        
+        imp_scale = ismissing(imp_scale) || is_weighting_potential ? zeros(T, size(q_eff_imp)) : RBExtBy2Array(imp_scale.data, grid)
+        for i in eachindex(imp_scale)
+            imp_scale[i] = is_pn_junction_point_type(rbpoint_types[i]) ? one(T) : zero(T)
```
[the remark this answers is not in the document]
lmh91: No, there is no reason. Just a remnant of (unoptimized) first lines of code of the trying out the idea (of the PR). 
I will change it as `imp_scale = is_pn_junction_point_type(rbpoint_types[i])` also generates less code.

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCylindrical.jl
lmh91: Yes, I think we actually do not need the `point_types` anymore (in that function).
We can just multiple all elements of `ϵ` with `(1 - imp_scale)` and do not need to check 
if it is marked as an undepleted point type.

src/PotentialCalculation/SuccessiveOverRelaxation/GPU_kernel_Cartesian3D.jl
lmh91: https://juliagpu.github.io/KernelAbstractions.jl/dev/api/#KernelAbstractions.@Const

It gives some additional information to the compiler, that those fields are not modified (read-only) during the kernel. 
This information can sometimes improve performance. 
Though, in our case, I did not observe any speed up.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: We only need `isMore` once.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: Did not really think about that name here. I can change it.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: At the edge of the depleted zone, `imp_scale` should converge against some value `x` in `[0, 1]`. 
The `else`-part should be to increase it again a little bit (towards `x`) in case it was reduced to some value smaller than `x`
in the previous iteration.

lmh91:
> I didn't test your code: when plotting `sim.q_eff_imp`, will that still be the "full" effective charge or will that also be scal
ed by `imp_scale`?

No, it will show the "full" charge. 

I believe we also don't "modify" it with `point_types` on the current master.

src/PotentialCalculation/SuccessiveOverRelaxation/GPU_kernel.jl
```diff
@@ -1,77 +1,81 @@
...
-            S, 
-            ϵ_r,
-            geom_weights[3],
-            i2, in2, i3, in3,
-            geom_weights_2...
-        )
+    # Comparison to CPU indices: (Cyl / Car)
+    # i3 <-> idx3 / ir / iz
+    # i2 <-> idx2 / iφ / iy
+    # i1 <-> idx1 / iz / ix
```
fhagemann, line 21: Did you test this on a non-symmetric detector in `Cartesian` coordinates to see if the indices are not mixed up?

src/PotentialCalculation/SuccessiveOverRelaxation/GPU_update.jl
```diff
@@ -0,0 +1,25 @@
...
+        ndrange = ndrange
+    ))
+    apply_boundary_conditions!(pcs, Val(update_even_points), only2d)
+    update_even_points = false
+    wait(kernel( 
+        pcs.potential, pcs.imp_scale, pcs.point_types, pcs.volume_weights, pcs.q_eff_imp, pcs.q_eff_fix, pcs.ϵ_r,
+        pcs.geom_weights, pcs.sor_const, update_even_points, depletion_handling_enabled, _is_weighting_potential, only_2d,
+        ndrange = ndrange
+    ))
+    apply_boundary_conditions!(pcs, Val(update_even_points), only2d)
```
fhagemann, line 23: Code duplication: Is there a reason we do not go for
```julia
for update_even_points in (true, false)
    wait(kernel( 
        pcs.potential, pcs.imp_scale, pcs.point_types, pcs.volume_weights, pcs.q_eff_imp, pcs.q_eff_fix, pcs.ϵ_r,
        pcs.geom_weights, pcs.sor_const, update_even_points, depletion_handling_enabled, _is_weighting_potential, only_2d, 
        ndrange = ndrange
    ))
    apply_boundary_conditions!(pcs, Val(update_even_points), only2d)
end
```
instead of having the same function with the same arguments called twice?

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,55 @@ Thus, there cannot be local extrema.
...
-    point_type::PointType, 
+    imp_scale::T,
     neighbor_potentials::NTuple{6,T}, 
     q_eff_imp::T, 
     volume_weight::T,
     sor_const::T
-)::Tuple{T, PointType} where {T}
+)::Tuple{T, T} where {T}
     vmin::T = min(neighbor_potentials...)
     vmax::T = max(neighbor_potentials...)
```
fhagemann, line 82: Not sure if this is faster but there is the function `minmax`:
```julia
vmin::T, vmax::T = minmax(neighbor_potentials...)
```
Also, is there a reason not to use `maximum` and `minimum` (or `extremum`) without `...`?
```julia
vmin::T , vmax::T = extremum(neighbor_potentials)
```

fhagemann: I just checked: for "only" 6 values in `neighbor_potentials` it makes no difference, but `extremum` is better if its 40 values:

```julia
t = Tuple(randn(6))
# (-0.9443487734649653, -0.37544349629655793, -1.6098211403752445, -0.7270684898446809, 1.941463962273232, -0.9060439688986951)

determine_max(t) = max(t...)
@btime determine_max($t);

determine_maximum(t) = maximum(t)
@btime determine_maximum($t);
```
```
  9.698 ns (0 allocations: 0 bytes)
  9.686 ns (0 allocations: 0 bytes)
```

I just checked: for "only" 6 values in `neighbor_potentials` it makes no difference, but `extremum` is better if its 40 values:

```julia
t = Tuple(randn(40))

determine_max(t) = max(t...)
@btime determine_max($t);

determine_maximum(t) = maximum(t)
@btime determine_maximum($t);
```
```
  3.079 μs (82 allocations: 1.88 KiB)
  1.254 μs (42 allocations: 992 bytes)

File: src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl [hunk 16]
```diff
@@ -71,33 +72,55 @@ Thus, there cannot be local extrema.
...
+    is_ptype = q_eff_imp < 0 
+    neighbor_relevant_extremum = is_ptype ? vmin : vmax
+    imp_contribution = q_eff_imp * volume_weight * sor_const 
+    
+    if !iszero(imp_contribution) 
+        new_potential_proposal = new_potential - (T(1)- imp_scale) * imp_contribution
+        if (new_potential_proposal > vmax || new_potential_proposal < vmin)
+            imp_scale::T = (imp_contribution - (new_potential - neighbor_relevant_extremum)) / (imp_contribution * sor_const)
+            if imp_scale < 0 imp_scale = T(0) end
+            if imp_scale > 1 imp_scale = T(1) end
```
fhagemann, line 93: This is exactly what `min` and `max` do:
```julia
imp_scale = max(imp_scale, T(0))
imp_scale = min(imp_scale, T(1))
```
or
```julia
imp_scale = min(max(imp_scale, T(0)), T(1))
```

There might even be a function where you give the edges of an interval and it does exactly this.

src/Simulation/Simulation.jl
```diff
@@ -930,7 +910,7 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
                                                 device_array_type = device_array_type,
                                                 not_only_paint_contacts = not_only_paint_contacts, 
                                                 paint_contacts = paint_contacts,
-                                                sor_consts = sor_consts )
+                                                sor_consts = is_last_ref ? T(1) : sor_consts )
```
fhagemann, line 913: Out of curiosity, is there a difference in performance between `T(0)` and `zero(T)` (or between `T(1)` and `one(T)`)?
And should we come up with a consistent way of defining a zero or a one in SSD, as we use both ways in different parts of the code?

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,55 @@ Thus, there cannot be local extrema.
...
+include("GPU_update.jl")
+
+include("convergence.jl")
 
+
+function mark_undep_bits!(point_types::Array{PointType, 3}, imp_scale::Array{T, 3}) where {T}
+    @inbounds for i in eachindex(imp_scale)
+        if is_fixed_point_type(point_types[i])
+            imp_scale[i] = T(0)
+        end
```
fhagemann, line 120: If we want to avoid this `if` statement:
```
imp_scale[i] *= !is_fixed_point_type(point_types[i])
```

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,55 @@ Thus, there cannot be local extrema.
...
+        if is_pn_junction_point_type(point_types[i]) && imp_scale[i] < 1
+            point_types[i] += undepleted_bit
+        end
```
fhagemann, line 123: If we want to avoid this `if` statement:
```
point_types[i] += undepleted_bit * (is_pn_junction_point_type(point_types[i]) && imp_scale[i] < 1)
```

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
```diff
@@ -71,33 +72,55 @@ Thus, there cannot be local extrema.
...
 include("GPU_kernel_Cylindrical.jl")
 include("GPU_kernel_Cartesian3D.jl")
 include("GPU_kernel.jl")
-include("GPU_convergence.jl")
+include("GPU_update.jl")
+
+include("convergence.jl")
 
+
+function mark_undep_bits!(point_types::Array{PointType, 3}, imp_scale::Array{T, 3}) where {T}
```
fhagemann, line 116: This function not only marks undepleted bits but also changes the `imp_scale` values for fixed point types.
Should this be included in the function name?

src/PotentialCalculation/SuccessiveOverRelaxation/convergence.jl
```diff
@@ -0,0 +1,49 @@
...
+                                ) where {T, S, depletion_handling_enabled, only_2d, _is_weighting_potential, DAT} #  <: GPUArrays.AbstractGPUArray
+    device = get_device(DAT)
+    ndrange = size(pcs.potential)[1:3] .- 2
+    kernel = get_sor_kernel(S, device)
+    c_limit = _is_weighting_potential ? convergence_limit : abs(convergence_limit * pcs.bias_voltage)
+    c = (one(c_limit) + c_limit) * 10 # Has to be larger than c_limit at the beginning
+    n_performed_iterations = 0
+    tmp_potential = similar(pcs.potential, ndrange)
+    inner_ranges = broadcast(i -> 2:size(tmp_potential, i) + 1, (1, 2, 3))
+    is_logging(io) = isa(io, Base.TTY) == false || (get(ENV, "CI", nothing) == "true")
```
fhagemann, line 20: Not a fan of `== false`, alternative would be `!isa(io, Base.TTY)`

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: This can be (is) done.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: I changed it.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: This would change all the values of `imp_scale` in the bulk.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: I don't like this as it is hard to read.
But I tested it and is seems to be slower. (Though might depend on the simulation and grid.)

src/Simulation/Simulation.jl
```diff
@@ -686,14 +652,23 @@ function update_till_convergence!( sim::Simulation{T, CS},
...
+        use_nthreads = _guess_optimal_number_of_threads_for_SOR(size(sim.weighting_potentials[contact_id].grid), Base.Threads.nthreads(), CS),    
+        not_only_paint_contacts = not_only_paint_contacts, 
+        paint_contacts = paint_contacts, 
+        point_types = depletion_handling ? sim.point_types : missing)
+    );
 
     cf::T = _update_till_convergence!( pcs, T(convergence_limit), device_array_type;
                                        only2d = Val{only_2d}(),
-                                       depletion_handling = Val{depletion_handling}(),
+                                       depletion_handling = Val{false}(),
```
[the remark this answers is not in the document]
lmh91: Is never needed for weighting potentials.

src/Simulation/Simulation.jl
lmh91: I don't think it does. Did a simple test. 
I guess we should at some point. PR is welcome.

test/comparison_to_analytic_solutions.jl
```diff
@@ -27,7 +27,7 @@ end
 
 @testset "Two Spheres Capacitor" begin
     sim = Simulation{T}(SSD_examples[:TwoSpheresCapacitor])
-    simulate!(sim, device_array_type = device_array_type, refinement_limits = [0.2, 0.1, 0.05, 0.02], convergence_limit = 1e-7)
+    simulate!(sim, device_array_type = device_array_type, refinement_limits = [0.2, 0.1, 0.05, 0.025, 0.01], convergence_limit = 1e-7)
```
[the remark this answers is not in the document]
lmh91: Yes and no. We used `@fastmath` before. I am not sure where exactly it came from.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
fhagemann: Why? If you multiply something by 1 (`true`) it will be the same (?)

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -198,7 +199,7 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
         q_eff_fix_tmp *= ϵ0_inv
 
         volume_weights::Array{T, 4} = RBExtBy2Array(T, grid)
-        ρ::Array{T, 4} = RBExtBy2Array(T, grid)
+        q_eff_imp::Array{T, 4} = RBExtBy2Array(T, grid)
```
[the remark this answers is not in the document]
fhagemann: Why is it `ρ_imp` but `ρ_eff_fix` (why `eff` for the fixed charge densities?)

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -275,8 +276,8 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                     volume_weights[ irbx, iy, iz, rbi ] = inv(volume_weight)
 
                     dV::T = Δmpx[inx] * Δmpy[iny] * Δmpz[inz]
-                    ρ[ irbx, iy, iz, rbi ] = dV * ρ_cell
-                    q_eff_fix[ irbx, iy, iz, rbi ] = dV * q_eff_fix_cell
+                    q_eff_imp[ irbx, iy, iz, rbi ] = dV * ρ_imp_cell
+                    q_eff_fix[ irbx, iy, iz, rbi ] = dV * ρ_eff_fix_cell
```
fhagemann, line 280: I get the `q_eff` here but not the `ρ_eff_...`..

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -290,14 +291,27 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
...
+        
+        new_imp_scale::Array{T, 4} = if ismissing(imp_scale) || is_weighting_potential 
+            broadcast(is_pn_junction_point_type, rbpoint_types)
+        else
+            new_imp_scale = RBExtBy2Array(imp_scale, grid)
+            for i in eachindex(new_imp_scale)
+                if !is_pn_junction_point_type(rbpoint_types[i]) 
+                    new_imp_scale[i] = 0
+                end
+            end
```
fhagemann, line 303: Instead of the loop:
```julia
new_imp_scale .*= broadcast(is_pn_junction_point_type, rbpoint_types)
```
??

Also, I guess julia is smart enough that we don't need to define the type of 0, i.e. `T(0)` or `zero(T)`?

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -57,15 +57,15 @@ function set_point_types_and_fixed_potentials!(point_types::Array{PointType, 3},
     nothing
 end
 
-function fill_ρ_and_ϵ!(ϵ::Array{T}, ρ_tmp::Array{T}, q_eff_fix_tmp::Array{T}, 
+function fill_ρimp_ϵ_ρfix(ϵ::Array{T}, ρ_imp_tmp::Array{T}, ρ_eff_fix_tmp::Array{T},
```
fhagemann, line 60: The order of the arguments of this function is not consistent with the output of `get_ρimp_ϵ_ρfix`.
Should we change the order of the arguments or adjust the function name, i.e. `fill_ϵ_ρimp_ρfix`?

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_update.jl
fhagemann: This typo is still in there.

lmh91: Ah, ups. Just corrected the one in the docstring.

src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl
lmh91: Ah, my head. Sure... I will try it out.

lmh91:
There seems to be a problem with the over relaxation.
There is some strange behaviour near the contact:
issue_sor

This only is present when the SOR constant is larger than 1.0.
I guess I will disable the over relaxation for electric potentials if `depletion_handling` is activated.
Its not that important performance vise anymore since we refactored and improved the refinement.

I tried using the over relaxation only in the first two refinement steps and later force it to 1.0, 
but this does not help for some reason I don't understand so far...

src/Simulation/Simulation.jl
fhagemann: @oschulz

lmh91:
> Looks to me like the grid points that were fixed in the initial grid are marked as undepleted later..

Yes there was some issue for `depletion_handling = false`. This is fixed now. 

I think this is ready to merge now. 
I would set the version to `v0.8.0-dev` and merge if there are no further comments.

PR #263, opened 2022-03-07
Add convenient constructors for `Ellipsoid`

SebastianRuffert:


src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -47,13 +47,32 @@ This is a sphere with inner radius 1 and outer radius 2.
...
+    rotation::SMatrix{3,3,T,9}
+    
+    function Ellipsoid{T,CO}(r::TR,φ::TP,θ::TT,origin::CartesianPoint{T}=zero(CartesianPoint{T}),
+        rotation::SMatrix{3,3,T,9}=one(SMatrix{3, 3, T, 9})) where {T,CO,TR<:T,TP<:Nothing,TT<:Nothing} 
+        new{T,CO,TR,TP,TT}(r,φ,θ,origin,rotation)
+    end
+end
+
+
+function Ellipsoid(::Type{T}=Float64; r = 1.0, φ = nothing, θ = nothing, origin = zero(CartesianPoint{T}),
```
fhagemann, line 65: Do we want `Float64` to be the default? The default for `Simulation` is `Float32`.

fhagemann, line 65: Right now, `φ` and `θ` can only be `nothing`. 
If we keep them here, there will be an error if they are not `nothing`.

We could expand this constructor with returning `CSGs` for non-full spheres, e.g. `CSGIntersection` of `Sphere` and `PartialTube` for partial `φ` or `CSGIntersection` of `Sphere` and `PartialCone` for partial `θ`.

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -47,13 +47,32 @@ This is a sphere with inner radius 1 and outer radius 2.
...
+    φ::TP 
+    θ::TT 
 
-    origin::CartesianPoint{T} = zero(CartesianPoint{T})
-    rotation::SMatrix{3,3,T,9} = one(SMatrix{3, 3, T, 9})
+    origin::CartesianPoint{T}
+    rotation::SMatrix{3,3,T,9}
+    
+    function Ellipsoid{T,CO}(r::TR,φ::TP,θ::TT,origin::CartesianPoint{T}=zero(CartesianPoint{T}),
+        rotation::SMatrix{3,3,T,9}=one(SMatrix{3, 3, T, 9})) where {T,CO,TR<:T,TP<:Nothing,TT<:Nothing}
```
fhagemann, line 59: This `where {T,CO,TR<:T,TP<:Nothing,TT<:Nothing}` is done to limit `TR` to `T`, and `TP` and `TT` to `Nothing`?

SebastianRuffert: Correct, at the moment, we only support full spheres

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -47,19 +47,32 @@ This is a sphere with inner radius 1 and outer radius 2.
...
+
+    CO = _boundary_from_symbol(boundary)
+    return Ellipsoid{T,CO}(T.(r), φ, θ, origin, rotation)
 end
 
 Ellipsoid{T,CO,TR,TP,TT}( e::Ellipsoid{T,CO,TR,TP,TT}; COT = CO,
             origin::CartesianPoint{T} = e.origin,
             rotation::SMatrix{3,3,T,9} = e.rotation) where {T,CO<:Union{ClosedPrimitive, OpenPrimitive},TR,TP,TT} =
-    Ellipsoid{T,COT,TR,TP,TT}(e.r, e.φ, e.θ, origin, rotation)
+    Ellipsoid{T,COT,TR,TP,TT}(T; r = e.r, φ = e.φ, θ = e.θ, origin = origin, rotation = rotation)
```
fhagemann, line 75: You cannot use keyword arguments here if you use the methods with curly brackets (now that `@with_kw` was removed)

fhagemann (dismissed):

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
lmh91: We do want `Float64` here.

src/ConstructiveSolidGeometry/VolumePrimitives/VolumePrimitives.jl
```diff
@@ -15,6 +15,16 @@ function sample(vp::AbstractVolumePrimitive)
...
 
+function _boundary_from_symbol(boundary::Symbol)
+    return if (boundary == :closed)
+        ClosedPrimitive
+    elseif (boundary == :open)
+        OpenPrimitive
+    else
+        error("Only `:closed` and `:open` are viable boundary options for CSG constructors")
+    end
+end
```
lmh91, line 26: Please remove this. We can just use `ClosedPrimitive` or `OpenPrimitive` directly.

We could change `CO` to `B` for boundary in general. But lets stick to `CO` in this PR (dedicated PR in case we want to).

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
lmh91: I guess we can leave out `φ` and `θ` in the constructors for now and just always force it to be nothing. 

These constructors should not return CSGs like `CSGDifference`, but always an `Ellipsoid`.

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -47,19 +47,32 @@ This is a sphere with inner radius 1 and outer radius 2.
...
+
+    origin::CartesianPoint{T}
+    rotation::SMatrix{3,3,T,9}
+        
+end
+
+function Ellipsoid{T,CO}(r::TR,φ::TP,θ::TT,origin::CartesianPoint{T}=zero(CartesianPoint{T}),
+    rotation::SMatrix{3,3,T,9}=one(SMatrix{3, 3, T, 9})) where {T,CO,TR<:T,TP<:Nothing,TT<:Nothing} 
+    Ellipsoid{T,CO,TR,TP,TT}(r,φ,θ,origin,rotation)
+end
```
lmh91, line 63: I think keywords would be better here:

```julia
function Ellipsoid{T,CO}(;
    r::TR = 1,
    origin = zero(CartesianPoint{T}),
    rotation = one(SMatrix{3, 3, T, 9})
) where {T,CO,TR} 
    Ellipsoid{T,CO,TR,Nothing,Nothing}(r,nothing,nothing,origin,rotation)
end
```

I think we can drop the type annotations for `origin` and `rotation` here. 
We should add `convert` methods for e.g. `CartesianPoint` if we don't already have them
so Julia will automatically convert them to the correct type with elements of type `T` in the default constructor.

lmh91 (changes requested):

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -47,19 +47,33 @@ This is a sphere with inner radius 1 and outer radius 2.
...
+# end
+
+function Ellipsoid(::Type{T}=Float64, ::Type{CO}=ClosedPrimitive; r::TR = 1.0, origin = zero(CartesianPoint{T}), 
+    rotation = one(SMatrix{3, 3, T, 9})) where {T,CO,TR<:T}
 
-    origin::CartesianPoint{T} = zero(CartesianPoint{T})
-    rotation::SMatrix{3,3,T,9} = one(SMatrix{3, 3, T, 9})
+    TP=Nothing
+    TT=Nothing
+    return Ellipsoid{T,CO,TR,TP,TT}(r, nothing, nothing, origin, rotation)
```
lmh91, line 70: Please put `Nothing` directly in the constructor at the end to keep it small. 

Hmm, I am still unsure whether it would be better here if we put `T` and `CO` into 
the curly brackets: `Ellipsoid{T, CO}(...`
And then add methods which call this method to add defaults. 
E.g. `Ellipsoid(::Type{T} = Float64, ::Type{CO} = ClosedPrimitve; kwargs...) = Ellipsoid{T, CO}(; kwargs...)`

src/ConstructiveSolidGeometry/VolumePrimitives/Ellipsoid.jl
```diff
@@ -81,26 +95,26 @@ function Geometry(::Type{T}, ::Type{Ellipsoid}, dict::AbstractDict, input_units:
         error("Partial Ellipsoid (`θ = θ`) is not yet supported.")
     end
     e = if r isa Tuple{T,T}
-        Ellipsoid{T,ClosedPrimitive,T,typeof(φ),typeof(θ)}(
+        Ellipsoid(T,ClosedPrimitive,
             r = r[2], 
-            φ = φ, 
-            θ = θ,
+            # φ = φ, 
+            # θ = θ,
```
lmh91, line 101: Please remove those.

lmh91 (changes requested):
Just mostly some style and clean up things. 

Almost there 👍

lmh91 (approved):

lmh91:
Please just add some more tests:

* Tests for all constructor methods. 
* The read in from a dictionary. 
* `in` method should be tested for several points. Also in combination with translation and rotations.
Here, also closed and open primitives should be tested (points on the surface). 

Regarding the tests for the constructors, please also have a look at the `@inferred` macro from the `Test` package.

lmh91:
Thanks 👍

PR #268, opened 2022-03-09
Don't overwrite imp_scale in contacts

lmh91:
Don't overwrite `imp_scale` in contacts as this unnecessary.


fhagemann:
What is the status now if simulations from v0.7 and before are read in. Is `imp_scale` still `missing` or is it determined from the `point_types`?

lmh91:
> What is the status now if simulations from v0.7 and before are read in. Is `imp_scale` still `missing` or is it determined from the `point_types`?

It is still missing. I will add determining it from the point_types in this PR.

lmh91:
`imp_scale` does now look like this: 

depl

PR #271, opened 2022-03-16
Add tutorial for custom impurity (of a p-n junction)

lmh91:


docs/src/man/electric_potential.md
```diff
@@ -39,6 +39,16 @@ plot(
 One contribution to the charge density $\rho(\vec{r})$ is the impurity density of the semiconductor of a detector.
 Some simple impurity density profiles are already implemented in SolidStateDetectors.jl and can be easily accessed in the configuration files. Note that all impurity densities are given in units of **atoms / particles** per volume.
 
+!!! note "Sign of the impurity density"
+    The sign of the impurity density determines whether it is p-type or n-type.
```
fhagemann, line 43: "it" --> "the semiconductor"

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+# The plates span over `y` and `z`. It is made infinite by defining the world smaller than the contacts
+# and using reflecting boundaries in `y` and `z`.
+# Let's get started by loading the packages and the detector and have a look at its geometry:
+
+using Plots
+using SolidStateDetectors
+using Unitful
+
+sim = Simulation{Float32}(SSD_examples[:InfiniteParallelPlateCapacitor])
+plot(sim.detector)
```
fhagemann, line 25: The plot looks very squeezed together. Is that because the world is much larger in `x` than the detector itself? Is this something we want to change?

Parallel Plate Capacitor

I think it would make sense to also set `show_semiconductor = true`, as this tutorial primarily deals with the semiconductor and not only the contacts.

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+
+# ## Define the Impurity Density
+
+# Now, we define the model for impurity density of the p-n junction.
```
fhagemann, line 29: Missing word: ... for **the** impurity density ...

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+# In order to do so, we need two define two things.
+
+# 1. We have to define a `struct` which has to be 
+# subtype of `SolidStateDetectors.AbstractImpurityDensity{T}`.
+# We define the model with two parameters in order to be able to vary the density level 
+# on the p-type and n-type regions independently.
```
fhagemann, line 35: on -> in ?

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+# ## Define the Impurity Density
+
+# Now, we define the model for impurity density of the p-n junction.
+# In order to do so, we need two define two things.
+
+# 1. We have to define a `struct` which has to be 
+# subtype of `SolidStateDetectors.AbstractImpurityDensity{T}`.
+# We define the model with two parameters in order to be able to vary the density level 
+# on the p-type and n-type regions independently. 
+# More parameters could be added. E.g., the position of the p-n junction.
```
fhagemann, line 36: Punctuation: ... added, e.g. the position ...

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+struct PNJunctionImpurityDensity{T} <: SolidStateDetectors.AbstractImpurityDensity{T} 
+    p_type_density::T
+    n_type_density::T
+end
+
+# 2. We have to define a method for `SolidStateDetectors.get_impurity_density` for our 
+# just defined impurity density. 
+# It has to take two arguments. An instance of our model and a point.
+# Since the infinite parallel plate capacitor is configured to be calculated on a cartesian grid,
+# the point has to be a `CartesianPoint`.
```
fhagemann, line 47: If I weren't used to SSD, I would read this as: for every new detector, you need to define a new impurity density profile. So what we are doing here sounds not applicable/usable for cylindrical grids.

However, (and this is also something we do for the pre-defined impurity density profiles), we can just do
```julia
function SolidStateDetectors.get_impurity_density(cdm::PNJunctionImpurityDensity{T},  pt::CylindricalPoint{T})::T where {T}
    SolidStateDetectors.get_impurity_density(cdm, CartesianPoint(pt))
end
```
Can we somehow add that here?

Ànd a general comment: Is there any reason, why we don't generally go for:
```julia
function SolidStateDetectors.get_impurity_density(
    cdm::PNJunctionImpurityDensity{T}, 
    pt::AbstractCoordinatePoint{T}
)::T where {T}
    cpt = CartesianPoint(pt) # converts it to CartesianPoint, does nothing if it is a CartesianPoint
    x = cpt[1] # In this example, we only need the `x` coordinate of the point.
    if x > 0
        -cdm.p_type_density # p-type region -> electrons are fixed -> negative charge 
    else 
        cdm.n_type_density  # n-type region -> holes are fixed -> positive charge
    end
end
```

`

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+# The function returns the impurity density at the respective position.
+# The returned value has to be in units of 1/m``^{3}``.
+
+#=
+!!! note "Sign of the impurity density"
+    The sign of the impurity density determines whether it is p-type or n-type.
```
fhagemann, line 53: Same here: "it" -> "the semiconductor"

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+end
+
+# ## Assign the Density and Calculate the Fields
+
+# Now, we create an instance of our model with reasonable parameters 
+# (for the geometry and bias voltage of the example detector)
+# and assign it to the detector.
+
+pn_junction_impurity_density = PNJunctionImpurityDensity{Float32}(3e16, 1.5e16)
+
```
fhagemann, line 81: Remove this line (?)

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+sim.detector = SolidStateDetector(sim.detector, pn_junction_impurity_density);
+
+# We are now ready to calculate the electric potential. 
+# Depletion handling has to be turned on, via the keyword `depletion_handling = true`,
+# in order to detect the regions where the detector is depleted.
+# The other settings for `calculate_electric_potential!` are very specific for this effectively 1D problem
+# to be calculated in 3D.
+# Afterwards, we also calculate the electric field and plot the 
+# electric potential, impurity scale map and the point type map. 
+# The two latter ons show us where the detector is depleted (`impurity scale == 1`).
```
fhagemann, line 91: Typo: "ons" -> "ones"

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+
+calculate_electric_potential!(
+    sim, 
+    convergence_limit = 1e-6,
+    max_tick_distance = (0.002, 1, 1) .* u"cm",
+    min_tick_distance = (1e-7, 1, 1) .* u"m",
+    refinement_limits = [0.005, 0.001],
+    depletion_handling = true,
+    verbose = true
```
fhagemann, line 100: `verbose = true` does not seems to give output in the tutorial..

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
fhagemann: If you use typewriter font style, you might also want to use an underscore: `impurity_scale == 1`

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+    plot(sim.imp_scale, y = 0),
+    plot(sim.point_types, y = 0),
+    layout = (3, 1),
+    aspect_ratio = :none,
+    size = (800, 800)
+)
+
+# ## 1D Plots of the P-N Junction
+
+# As this is in principle a 1D simulation, lets better plot the
```
fhagemann, line 115: Typo: "lets" -> "let's"
But I would prefer to rephrase this to: simulation, **it is most reasonable to** plot ...

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+    plot(sim.point_types, y = 0),
+    layout = (3, 1),
+    aspect_ratio = :none,
+    size = (800, 800)
+)
+
+# ## 1D Plots of the P-N Junction
+
+# As this is in principle a 1D simulation, lets better plot the 
+# charge density from the impurities, the electric potential and the electric field only over `x`.
```
fhagemann, line 116: electric field **strength** (?)

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+
+xs = sim.electric_potential.grid[1] * 1000;
```
fhagemann, line 118: You might want to add units here, then you don't have to add them to the xlabel later.

```julia
xs = sim.electric_potential.grid[1] * 1000u"mm"
```

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+plot(
+    begin
+        plot(xs, sim.q_eff_imp.data[:, 1, 1] .* sim.imp_scale[:, 1, 1], lw = 4, label ="", yguide = "\$\\rho\$ [e/m\$^3\$]")
```
fhagemann, line 121: Is the unit right? I thought `ϵ0` would already be in here, including its units...

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+        vline!([0], lw = 4, color = "black", label = "p-n junction")
+    end,
+    begin
+        plot(xs, sim.electric_potential.data[:,1,1], lw = 4, label ="", yguide = "\$\\Phi\$ [V]")
+        vline!([sim.detector.contacts[1].geometry.origin.x] * 1000, lw = 4, color = "green", label = "N+ Contact")
+        vline!([sim.detector.contacts[2].geometry.origin.x] * 1000, lw = 4, color = "red", label = "P+ Contact")
+        vline!([0], lw = 4, color = "black", label = "p-n junction")
+    end,
+    begin
+        plot(xs, map(E -> E[1]/1000, sim.electric_field.data[:,1,1]), lw = 4, label ="", yguide = "\$\\mathcal{E}_z\$ [V/mm]")
```
fhagemann, line 133: Shouldn't this be `E_x` instead of `E_z` then?

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -0,0 +1,143 @@
...
+        vline!([sim.detector.contacts[1].geometry.origin.x] * 1000, lw = 4, color = "green", label = "N+ Contact")
+        vline!([sim.detector.contacts[2].geometry.origin.x] * 1000, lw = 4, color = "red", label = "P+ Contact")
+        vline!([0], lw = 4, color = "black", label = "p-n junction")
+    end,
+    layout = (3, 1),
+    xlims = (-5.1, 5.1),
+    size = (800, 800),
+    lw = 2,
+    xguide = "x [mm]"
```
fhagemann, line 142: if you give `xs` units, you might not need to add `[mm]` here.

fhagemann (commented):
Thanks for taking care of updating the tutorials!

In the former "old" tutorial on the inverted coaxial point contact detector, some of the plots are missing data. Find some screenshots attached. However, the full plot is there, when you click on them. Do you know why that is?


Screenshot from 2022-03-20 17-26-56

Screenshot from 2022-03-20 17-27-06

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
lmh91: Hmm, its the `x` component of the electric field. 
Which is, in this case, the same as the field strength. 

I just write it out even if it becomes a bit lengthy.

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
lmh91: I changed some plotting parameters.

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
lmh91: Ah, true... I forgot to change this.

lmh91 (commented):
Thanks for the comments. 
I think I integrated them all.

docs/src/tutorials/custom_impurity_density_pn_junction_lit.jl
```diff
@@ -22,18 +22,19 @@ using SolidStateDetectors
...
+# Now, we define the model for the impurity density of the p-n junction.
 # In order to do so, we need two define two things.
 
 # 1. We have to define a `struct` which has to be 
 # subtype of `SolidStateDetectors.AbstractImpurityDensity{T}`.
 # We define the model with two parameters in order to be able to vary the density level 
-# on the p-type and n-type regions independently. 
-# More parameters could be added. E.g., the position of the p-n junction.
+# in the p-type and n-type regions independently. 
+# More parameters could be added, e.g., the position of the p-n junction.
```
fhagemann, line 37: I think we don't need the comma after e.g. but I might be wrong.

lmh91: I just looked it and it seems to be another British vs American English thing...

lmh91:
> > Thanks for taking care of updating the tutorials!
> > In the former "old" tutorial on the inverted coaxial point contact detector, some of the plots are missing data. Find some screenshots attached. However, the full plot is there, when you click on them. Do you know why that is?
> > Screenshot from 2022-03-20 17-26-56
> > Screenshot from 2022-03-20 17-27-06
> 
> Also, this problem still seems to be there.

In my browser it looks just fine. Seems to be some issue with your browser.

fhagemann (approved):
Well, if I reload the page, the plots in the tutorial seem to appear... Strange!

PR #275, opened 2022-03-25
Improve grid creation for weighting potentials

lmh91:
After calculating the electric potential, we can use the calculated potential and also the impurity scale field 
to add ticks in the creation of the initial grid of the weighting potentials.

This is especially important for cases where the detector is not fully depleted.

### DIFF
--- a/src/Axes/DiscreteAxis.jl
+++ b/src/Axes/DiscreteAxis.jl
@@ -313,26 +313,15 @@ end
 """
     merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
 
-Merges two vectors, removes the entries from the second vector if they are too close 
-to entries of the first vector (via `min_diff`) and returns one sorted vector.
+Merge all elements of the second vector, `imp_second_order`, into the first vector, `imp`, if they are not 
+to close (via `min_diff`) to elements of the first vector.
+Returns the sorted and uniqued vector.
 """
 function merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
     sorted_imp = sort(imp)
-    all_merged = unique!(sort!(vcat(sorted_imp, imp_second_order)))
-    Δall_merged = diff(all_merged)
-    inds_diff_too_small = findall(Δ -> Δ < min_diff, Δall_merged)
-    delete_inds = Int[]
-    for i in inds_diff_too_small
-        if insorted(all_merged[i], sorted_imp)
-            push!(delete_inds, i+1)
-        elseif insorted(all_merged[i+1], sorted_imp)
-            push!(delete_inds, i)
-        else
-            error("This should never happen.") # I will leave this for now just to be sure.
-        end
-    end
-    deleteat!(all_merged, delete_inds)
-    all_merged
+    nearest_inds = map(x -> searchsortednearest(sorted_imp, x), imp_second_order)
+    merge_inds = filter(i -> abs(sorted_imp[nearest_inds[i]] - imp_second_order[i]) >= min_diff, eachindex(imp_second_order))
+    return unique!(sort!(vcat(imp, imp_second_order[merge_inds])))
 end
 
 function get_new_ticks_to_equalize_ratios_on_side(t::AbstractVector{T}; max_ratio = T(2)) where {T}
--- a/src/Axes/DiscreteAxis.jl
+++ b/src/Axes/DiscreteAxis.jl
@@ -311,13 +311,13 @@ function merge_close_ticks(v::AbstractVector{T}; min_diff::T = T(1e-6)) where {T
 end
 
 """
-    merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
+    merge_second_order_important_ticks(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
 
-Merge all elements of the second vector, `imp_second_order`, into the first vector, `imp`, if they are not 
-to close (via `min_diff`) to elements of the first vector.
-Returns the sorted and uniqued vector.
+Merge all elements of the second vector, `imp_second_order`, into the first vector, `imp`, 
+if they are not too close (via `min_diff`) to elements of the first vector.
+Returns the merged vector sorted and uniqued.
 """
-function merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
+function merge_second_order_important_ticks(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
     sorted_imp = sort(imp)
     nearest_inds = map(x -> searchsortednearest(sorted_imp, x), imp_second_order)
     merge_inds = filter(i -> abs(sorted_imp[nearest_inds[i]] - imp_second_order[i]) >= min_diff, eachindex(imp_second_order))
--- a/src/Axes/DiscreteAxis.jl
+++ b/src/Axes/DiscreteAxis.jl
@@ -310,6 +310,30 @@ function merge_close_ticks(v::AbstractVector{T}; min_diff::T = T(1e-6)) where {T
     v[1:n]
 end
 
+"""
+    merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
+
+Merges and sorts two vectors into one sorted vector. But only add elements of the second vector, `imp_second_order`, if
+they are not too close (via `min_diff`) to elements of the first vector. 
+"""
+function merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
+    sorted_imp = sort(imp)
+    all_merged = unique!(sort!(vcat(sorted_imp, imp_second_order)))
+    Δall_merged = diff(all_merged)
+    inds_diff_too_small = findall(Δ -> Δ < min_diff, Δall_merged)
+    delete_inds = Int[]
+    for i in inds_diff_too_small
+        if insorted(all_merged[i], sorted_imp)
+            push!(delete_inds, i+1)
+        elseif insorted(all_merged[i+1], sorted_imp)
+            push!(delete_inds, i)
+        else
+            error("This should never happen.") # I will leave this for now just to be sure.
+        end
+    end
+    deleteat!(all_merged, delete_inds)
+    all_merged
+end
 
 function get_new_ticks_to_equalize_ratios_on_side(t::AbstractVector{T}; max_ratio = T(2)) where {T}
     @assert length(t) == 3 
--- a/src/Axes/DiscreteAxis.jl
+++ b/src/Axes/DiscreteAxis.jl
@@ -298,7 +298,7 @@ function merge_closest_ticks!(v::AbstractVector{T}, n::Int = length(v); min_diff
 end
 function merge_close_ticks(v::AbstractVector{T}; min_diff::T = T(1e-6)) where {T}
     l = length(v)
-    if l <= 1 return v end
+    l <= 1 && return v
     n = merge_closest_ticks!(v, min_diff = min_diff)
     reduced = n < l
     l = n
@@ -313,8 +313,8 @@ end
 """
     merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
 
-Merges and sorts two vectors into one sorted vector. But only add elements of the second vector, `imp_second_order`, if
-they are not too close (via `min_diff`) to elements of the first vector. 
+Merges two vectors, removes the entries from the second vector if they are too close 
+to entries of the first vector (via `min_diff`) and returns one sorted vector.
 """
 function merge_second_order_important_points(imp::Vector{T}, imp_second_order::Vector{T}; min_diff::T = T(1e-6)) where {T}
     sorted_imp = sort(imp)
--- a/src/Axes/DiscreteAxis.jl
+++ b/src/Axes/DiscreteAxis.jl
@@ -280,6 +280,7 @@ end
 Base.convert(T::Type{NamedTuple}, x::DiscreteAxis; unit = u"m/m") = T(x)
 
 function merge_closest_ticks!(v::AbstractVector{T}, n::Int = length(v); min_diff::T = T(1e-6)) where {T}
+    n == 1 && return n
     Δv = diff(v[1:n])
     Δ_min, Δv_min_indx = findmin(Δv)
     vFirst = v[1]
--- a/src/Axes/DiscreteAxis.jl
+++ b/src/Axes/DiscreteAxis.jl
@@ -298,7 +298,7 @@ function merge_closest_ticks!(v::AbstractVector{T}, n::Int = length(v); min_diff
 end
 function merge_close_ticks(v::AbstractVector{T}; min_diff::T = T(1e-6)) where {T}
     l = length(v)
-    if l == 1 return v end
+    if l <= 1 return v end
     n = merge_closest_ticks!(v, min_diff = min_diff)
     reduced = n < l
     l = n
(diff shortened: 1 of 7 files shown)


src/ScalarPotentials/ImpurityScale.jl
```diff
@@ -46,4 +46,20 @@ function ImpurityScale(nt::NamedTuple)
...
 end
-Base.convert(T::Type{ImpurityScale}, x::NamedTuple) = T(x)
\ No newline at end of file
+Base.convert(T::Type{ImpurityScale}, x::NamedTuple) = T(x)
+
+
+"""
+    get_ticks_at_positions_of_edge_of_depleted_volumes(field::ImpurityScale{T, 3})
+
+The impurity scale field is analyzed in order to find and return ticks where the gradient is strong,
```
fhagemann, line 55: Can you specify this a bit more? What does `analyze` mean and what is `strong`?
You use the value 0.5 later on in this function, maybe add that here into the definition of `strong`..

src/ScalarPotentials/ImpurityScale.jl
```diff
@@ -46,4 +46,20 @@ function ImpurityScale(nt::NamedTuple)
...
+which is the case at the surface of the depleted volume of a semiconductor.
+"""
+function get_ticks_at_positions_of_edge_of_depleted_volumes(field::ImpurityScale{T, 3}) where T
+    dims = (1, 2, 3)
+    mps::NTuple{3, Vector{T}} = broadcast(idim -> StatsBase.midpoints(field.grid[idim]), dims)
+    extended = length.(mps) .> 0
+    max_diffs = broadcast(idim -> extended[idim] ? map(i->maximum(abs.(selectdim(field.data, idim, i+1) .- selectdim(field.data, idim, i))), eachindex(mps[idim]))::Vector{T} : T[], dims);
+    inds = broadcast(idim -> extended[idim] ? findall(Δ -> 0.5 < Δ < 1, max_diffs[idim]) : Int[], dims) # 0.5 seems to be a good limit as we don't want to add too many ticks. 
+    return broadcast(idim -> extended[idim] ? mps[idim][inds[idim]] : T[], dims)
```
fhagemann, line 64: Why not summarize this into one `broadcast`?
```julia
    return broadcast(idim -> extended[idim] ? mps[idim][findall(Δ -> 0.5 < Δ < 1, max_diffs[idim])] : T[], dims)
```
For readability?

fhagemann (changes requested):

src/ScalarPotentials/ElectricPotential.jl
```diff
@@ -40,4 +40,20 @@ function ElectricPotential(nt::NamedTuple)
...
-Base.convert(T::Type{ElectricPotential}, x::NamedTuple) = T(x)
\ No newline at end of file
+Base.convert(T::Type{ElectricPotential}, x::NamedTuple) = T(x)
+
+"""
+    get_ticks_at_positions_of_large_gradient(field::ElectricPotential)
+
+The electric potential is analyzed in order to find and return ticks where the gradient (electric field) is strong.
+"""
+function get_ticks_at_positions_of_large_gradient(field::ElectricPotential{T, 3}) where T
```
fhagemann, line 50: Would it make sense to make this function for arguments of type `ScalarPotential`?
And can you rename the `field` variable to `epot` (as we try to use consistent naming for variables of the same type)?

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -188,6 +188,8 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                         ig = find_closest_gridpoint(pt, point_types.grid)
                         if is_undepleted_point_type(point_types.data[ig...]) 
                             ϵ[ix, iy, iz] *= scaling_factor_for_permittivity_in_undepleted_region(det.semiconductor) * (1 - imp_scale[ig...])
+                        elseif is_fixed_point_type(point_types.data[ig...])
+                            ϵ[ir, iφ, iz] *= scaling_factor_for_permittivity_in_undepleted_region(det.semiconductor)
```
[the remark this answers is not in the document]
lmh91: Ups. I forced pushed the fix.

Curious that the tests went through...

lmh91 (commented):
I took your comments into account.

I also changed how the ticks generated by the electric potential and impurity scale are
merged into the axis ticks. I don't add ticks which are too close to ticks generated from the geometry. I encountered a case where this led to `NaN`'s.

src/Axes/DiscreteAxis.jl
```diff
@@ -309,6 +310,30 @@ function merge_close_ticks(v::AbstractVector{T}; min_diff::T = T(1e-6)) where {T
...
+    delete_inds = Int[]
+    for i in inds_diff_too_small
+        if insorted(all_merged[i], sorted_imp)
+            push!(delete_inds, i+1)
+        elseif insorted(all_merged[i+1], sorted_imp)
+            push!(delete_inds, i)
+        else
+            error("This should never happen.") # I will leave this for now just to be sure.
+        end
+    end
```
fhagemann, line 333: Is the method using `push!` the most efficient?

If we stick to your original docstring, and we only add the ticks of the second array if they fit (rather then adding all of them and removing them afterwars), can we do something like (untested):
```julia
sorted_imp_1 = sort(imp)
sorted_imp_2 = sort(imp_second_order)
final = Vector{T}(undef, length(sorted_imp1) + length(sorted_imp_2))
i1 = 1 # index for sorted_imp_1
i2 = 1 # index for sorted_imp_2
for i in eachindex(final)
    s1 = sorted_imp_1[i1]
    s2 = sorted_imp_2[i2]
    if s1 < s2
        final[i] = s1
        i1 += 1
        if s2 - s1 < min_diff
            i2 += 1
        end
    else
        if s1 - s2 < min_diff
            final[i] = s2
        else
            final[i] = s1
            i1 += 1
        end
        i2 += 1
    end
end
```
?

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -188,6 +188,8 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                         ig = find_closest_gridpoint(pt, point_types.grid)
                         if is_undepleted_point_type(point_types.data[ig...]) 
                             ϵ[ix, iy, iz] *= scaling_factor_for_permittivity_in_undepleted_region(det.semiconductor) * (1 - imp_scale[ig...])
+                        elseif is_fixed_point_type(point_types.data[ig...])
+                            ϵ[ix, iz, iz] *= scaling_factor_for_permittivity_in_undepleted_region(det.semiconductor)
```
fhagemann, line 192: Still a typo: it is `iy`, not `iz`

src/ScalarPotentials/ElectricPotential.jl
```diff
@@ -40,4 +40,22 @@ function ElectricPotential(nt::NamedTuple)
...
+Base.convert(T::Type{ElectricPotential}, x::NamedTuple) = T(x)
+
+"""
+    get_ticks_at_positions_of_large_gradient(epot::ElectricPotential)
+
+The electric potential is analyzed in order to find and return ticks where the gradient (electric field) is strong.
+The electric field strength is the strongest at the position of the pn-junction at least for the 1D linear case.
+Thus, this function is likely to return ticks which are located close to the pn-junction of a semiconductor.
+"""
+function get_ticks_at_positions_of_large_gradient(epot::ElectricPotential{T, 3}) where T
```
fhagemann, line 52: It is a pain to compare functions on GitHub preview: what is the difference between this function and `get_ticks_at_positions_of_edge_of_undepleted_volumes`. There seem to be quite some duplications (?) Can these functions we somehow merged where similar?

src/ScalarPotentials/ImpurityScale.jl
```diff
@@ -1,7 +1,7 @@
 """
     struct ImpurityScale{T, N, S, AT} <: AbstractArray{T, N}
         
-Impurity scalar field of the simulation.
+Impurity scalar impscale of the simulation.
```
fhagemann, line 4: Find and replace?

src/ScalarPotentials/ImpurityScale.jl
```diff
@@ -15,7 +15,7 @@ It can takes values between `0` and `1`:
 * `S`: Coordinate system (`Cartesian` or `Cylindrical`).
 * `AT`: Axes type.  
         
-## Fields
+## impscales
```
fhagemann, line 18: Find and replace?

src/ScalarPotentials/ImpurityScale.jl
```diff
@@ -46,4 +46,23 @@ function ImpurityScale(nt::NamedTuple)
...
+
+The impurity scale field is analyzed in order to find and return ticks where the gradient is strong,
+which is the case at the surface of the depleted volume of a semiconductor.
+"""
+function get_ticks_at_positions_of_edge_of_depleted_volumes(impscale::ImpurityScale{T, 3}) where T
+    dims = (1, 2, 3)
+    mps::NTuple{3, Vector{T}} = broadcast(idim -> StatsBase.midpoints(impscale.grid[idim]), dims)
+    extended = length.(mps) .> 0
+    max_diffs = broadcast(idim -> extended[idim] ? map(i->maximum(abs.(selectdim(impscale.data, idim, i+1) .- selectdim(impscale.data, idim, i))), eachindex(mps[idim]))::Vector{T} : T[], dims);
+    inds = broadcast(idim -> extended[idim] ? findall(Δ -> 0.5 < Δ < 1, max_diffs[idim]) : Int[], dims) # 0.5 seems to be a good limit as we don't want to add too many ticks.
```
fhagemann, line 63: Not a fan of creating empty arrays here for dimensions in which the grid is not extended..

src/Axes/DiscreteAxis.jl
```diff
@@ -309,6 +310,30 @@ function merge_close_ticks(v::AbstractVector{T}; min_diff::T = T(1e-6)) where {T
...
+    Δall_merged = diff(all_merged)
+    inds_diff_too_small = findall(Δ -> Δ < min_diff, Δall_merged)
+    delete_inds = Int[]
+    for i in inds_diff_too_small
+        if insorted(all_merged[i], sorted_imp)
+            push!(delete_inds, i+1)
+        elseif insorted(all_merged[i+1], sorted_imp)
+            push!(delete_inds, i)
+        else
+            error("This should never happen.") # I will leave this for now just to be sure.
```
fhagemann, line 331: "for now" --> until when?

src/ScalarPotentials/ElectricPotential.jl
```diff
@@ -40,4 +40,22 @@ function ElectricPotential(nt::NamedTuple)
...
     ElectricPotential{T, N, S, typeof(grid.axes)}( ustrip.(uconvert.(u"V", nt.values)), grid)
 end
-Base.convert(T::Type{ElectricPotential}, x::NamedTuple) = T(x)
\ No newline at end of file
+Base.convert(T::Type{ElectricPotential}, x::NamedTuple) = T(x)
+
+"""
+    get_ticks_at_positions_of_large_gradient(epot::ElectricPotential)
+
+The electric potential is analyzed in order to find and return ticks where the gradient (electric field) is strong.
```
fhagemann, line 48: Define `strong`. Should "`strong`" be an optional argument for this function?

src/ScalarPotentials/ElectricPotential.jl
```diff
@@ -40,4 +40,22 @@ function ElectricPotential(nt::NamedTuple)
...
+The electric field strength is the strongest at the position of the pn-junction at least for the 1D linear case.
```
fhagemann, line 49: "The electric field is the strongest..." or "The electric field strength is the largest..."
The electric field strength is also large close to sharp edges of contacts.

fhagemann (changes requested):

src/ScalarPotentials/ImpurityScale.jl
fhagemann: ?

src/ScalarPotentials/ElectricPotential.jl
lmh91: The impurity scale field does not need to be normalized and the selection criterion is different.

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
fhagemann: It looks like these cases are not covered by our tests?
Do you have a MWE where this is important that we can add to the tests?

lmh91: Ah, I thought it is tested in the test set "Depletion Handling: PN Junction", but I just realized that only the electric potential is calculated there.

src/Simulation/Simulation.jl
```diff
@@ -260,60 +269,66 @@ function Grid(sim::Simulation{T, Cylindrical};
...
-    important_r_points = fill_up_ticks(important_r_points, max_distance_r)
-
-    append!(important_z_points, endpoints(world.intervals[3])...)
-    important_z_points = unique!(sort!(important_z_points))
+    iL = searchsortedfirst(important_r_ticks, world.intervals[1].left)
+    iR = searchsortedfirst(important_r_ticks, world.intervals[1].right)
+    important_r_ticks = unique(map(t -> isapprox(t, 0, atol = 1e-12) ? zero(T) : t, important_r_ticks[iL:iR]))
+    important_r_ticks = merge_close_ticks(important_r_ticks)
+    imp2order_r_ticks = merge_close_ticks(second_order_imp_ticks[1], min_diff = world_Δs[1] / 20)
+    important_r_ticks = merge_second_order_important_ticks(important_r_ticks, imp2order_r_ticks)
```
fhagemann, line 282: Isn't `important_r_ticks` supposed to be unique already by this implementation?
(Same thing later for `φ`, `z`, `x` and `y`)?
Do we need the `unique!` statement in `merge_second_order_important_ticks` then?

src/Simulation/Simulation.jl
```diff
@@ -260,60 +269,66 @@ function Grid(sim::Simulation{T, Cylindrical};
...
+    important_z_ticks = unique(map(t -> isapprox(t, 0, atol = 1e-12) ? zero(T) : t, important_z_ticks[iL:iR]))
+    important_z_ticks = merge_close_ticks(important_z_ticks)
+    imp2order_z_ticks = merge_close_ticks(second_order_imp_ticks[3], min_diff = world_Δs[3] / 20)
+    important_z_ticks = merge_second_order_important_ticks(important_z_ticks, imp2order_z_ticks)
+    important_z_ticks = initialize_axis_ticks(important_z_ticks; max_ratio = T(max_distance_ratio))
+    important_z_ticks = fill_up_ticks(important_z_ticks, max_distance_z)
+
+    append!(important_φ_ticks, endpoints(world_φ_int)...)
+    important_φ_ticks = unique!(sort!(important_φ_ticks))
     if add_points_between_important_point
```
fhagemann, line 302: Here again, it's `point` instead of `tick`.
Also, we use one function from `StatsBase` called `midpoints`.. 
Maybe `points` is a better choice than `ticks`?

lmh91: IMO ticks is better. In principle, both is fine. Just depends on your definition of a point 
(which also depends on the context).
Using ticks makes it a bit more clear what happens here (at least in my opinion).

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -188,6 +188,8 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                         ig = find_closest_gridpoint(pt, point_types.grid)
                         if is_undepleted_point_type(point_types.data[ig...]) 
                             ϵ[ix, iy, iz] *= scaling_factor_for_permittivity_in_undepleted_region(det.semiconductor) * (1 - imp_scale[ig...])
+                        elseif is_fixed_point_type(point_types.data[ig...])
+                            ϵ[ix, iy, iz] *= scaling_factor_for_permittivity_in_undepleted_region(det.semiconductor)
```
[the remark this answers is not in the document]
fhagemann: If you plot ϵ, can you see the different values for the different semiconductors? Or do you then just see 0 and „Inf“?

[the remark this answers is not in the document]
lmh91: No, no other values are shown. 
`sim.ϵ_r` comes from the calculation of the electric potential where the scaling is not applied.

PR #285, opened 2022-04-25
Add convenient constructor for `Torus`

SebastianRuffert:


test/ConstructiveSolidGeometry/CSG_primitives.jl
```diff
@@ -110,16 +110,52 @@ no_translations = (rotation = one(SMatrix{3, 3, T, 9}), translation = zero(Carte
...
+        @test in(CartesianPoint{Float64}(1,0,-1),torus_closed_trafo)
+        @test !in(CartesianPoint{Float64}(1,0,-1),torus_open_trafo)
+        @test in(CartesianPoint{Float64}(0,0,0),torus_closed_trafo)
+        @test !in(CartesianPoint{Float64}(0,0,0),torus_open_trafo)
+        tol = 1e-8
+        @test !in(CartesianPoint{Float64}(0,-tol,0),torus_closed_trafo)
+        @test in(CartesianPoint{Float64}(0,+tol,0),torus_open_trafo)
+        #Test all valid keyword argument types
+        @inferred CSG.Torus{Float64}(φ=nothing,θ=nothing)
+        CSG.Torus{Float64}(φ=π,θ=(π,2π)) #inferred does not work here, due to evaluation of _get_conemantle_type(θ) during runtime
```
fhagemann, line 157: Can we change the implementation of `_get_conemantle_type` such that we dont have an `if`-`else`-statement at runtime to make `@inferred` pass?

fhagemann:
It might also make sense to allow `φ` to be a `Tuple` and do the conversion to a `T` in the constructor (see #240).

test/ConstructiveSolidGeometry/CSG_primitives.jl
lmh91: Hmm, the issue is that the surface type depends on values of `θ` and not the type of `θ`...

This could be "solved" by removing `TT1` and `TT2` from `Torus`. 
But then we shift this type instability to the function `surfaces(t::Torus)`... 

I think `:inwards`, `:outwards` could be managed without type instabilities. 
About the `:flat` case I am not sure though. I will look a bit more into that.

lmh91: But I think this falls outside the scope of this PR as this is a discussion 
about the general implementation of the structs `Torus`, `ConeMantle` and the functions `surfaces` or `TorusThetaSurface`.

I would say we open an dedicated issue for this and accepts that `@inferred` does not work for this case for now.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -83,22 +83,31 @@ struct Cone{T,CO,TR,TP<:Union{Nothing,T}} <: AbstractVolumePrimitive{T, CO}
 end
 
 #Type conversion happens here
+function Cone{T,CO}(r, φ::Tuple{T,T}, hZ, origin, rotation) where {T,CO}
+    r = _csg_convert_args(T, r)
+    φ = _csg_convert_args(T, φ)
+    _φ = abs(φ[2]-φ[1])
+    rotation=rotation*SMatrix{3}(cos(φ[1]),sin(φ[1]),0,-sin(φ[1]),cos(φ[1]),0,0,0,1)
```
fhagemann, line 90: Shorten this a bit? Something like `rotation *= SMatrix{3}(RotZ(φ[1]))`

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -83,22 +83,31 @@ struct Cone{T,CO,TR,TP<:Union{Nothing,T}} <: AbstractVolumePrimitive{T, CO}
 end
 
 #Type conversion happens here
+function Cone{T,CO}(r, φ::Tuple{T,T}, hZ, origin, rotation) where {T,CO}
+    r = _csg_convert_args(T, r)
+    φ = _csg_convert_args(T, φ)
```
fhagemann, line 88: If you force `φ` to be `Tuple{T,T}`, I don't see why you would need to `_csg_convert_args` it afterwards..

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -66,14 +66,86 @@ This is a `Torus` with `r_tube` having an inner radius of 1 and an outer radius
...
+        T = float(promote_type(eltypes...))
+        Torus{T,CO}(r_torus, r_tube, φ, θ, origin, rotation)
+end
+
+function Torus(CO, r_torus::TRTo, r_tube::TRTu, φ::Nothing, θ::Nothing, 
+    origin::PT, rotation::ROT) where {TRTo, TRTu, PT, ROT}
+        eltypes = _csg_get_promoted_eltype.((TRTo, TRTu, PT, ROT))
+        T = float(promote_type(eltypes...))
+        Torus{T,CO}(r_torus, r_tube, φ, θ, origin, rotation)
+end
```
lmh91, line 127: We don't need this.

Since we define out defaults values for the primitives as integers (thanks to @fhagemann's idea)
we can also add the method `_csg_get_promoted_eltype(::Type{Nothing}) = Int` 
to get rid of those methods.

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -66,14 +66,86 @@ This is a `Torus` with `r_tube` having an inner radius of 1 and an outer radius
...
+
+function Torus(::Type{CO} = ClosedPrimitive;
+    r_torus = 1, 
+    r_tube = 1,
+    φ = nothing,
+    θ = nothing,
+    origin = zero(CartesianPoint{Int64}), 
+    rotation = one(SMatrix{3, 3, Int64, 9})
+) where {CO}
+    Torus(CO, r_torus, r_tube, φ, θ, origin, rotation)
```
lmh91, line 137: I think we should use `Int` here and not `Int64`.
On 32bit systems `1` will be of type `Int32`. Not sure if it would cause issues but using `Int` should be better.

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -66,14 +66,86 @@ This is a `Torus` with `r_tube` having an inner radius of 1 and an outer radius
...
+    rotation::SMatrix{3,3,T,9}
+end
+
+#Type conversion happens here
+function Torus{T,CO}(r_torus, r_tube, φ::Tuple{T,T}, θ, origin, rotation) where {T,CO}
+    r_torus = _csg_convert_args(T, r_torus)
+    r_tube = _csg_convert_args(T, r_tube)
+    φ = _csg_convert_args(T, φ)
+    _φ = abs(φ[2]-φ[1])
+    rotation=rotation*SMatrix{3}(cos(φ[1]),sin(φ[1]),0,-sin(φ[1]),cos(φ[1]),0,0,0,1)
```
fhagemann, line 85: Shorten?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -66,14 +66,86 @@ This is a `Torus` with `r_tube` having an inner radius of 1 and an outer radius
...
+
+    origin::CartesianPoint{T}
+    rotation::SMatrix{3,3,T,9}
+end
+
+#Type conversion happens here
+function Torus{T,CO}(r_torus, r_tube, φ::Tuple{T,T}, θ, origin, rotation) where {T,CO}
+    r_torus = _csg_convert_args(T, r_torus)
+    r_tube = _csg_convert_args(T, r_tube)
+    φ = _csg_convert_args(T, φ)
```
fhagemann, line 83: Type instability?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -117,7 +189,7 @@ function Geometry(::Type{T}, ::Type{Torus}, dict::AbstractDict, input_units::Nam
         throw(ConfigFileError("Error when trying to parse φ from configuration file."))
     end
     θ = parse_θ_of_primitive(T, dict, angle_unit)
-    TT1, TT2 = _get_conemantle_type(θ)
+    #TT1, TT2 = _get_conemantle_type(θ)
```
fhagemann, line 192: Because `TT1` and `TT2` are now determined in `Torus{T}(...)` ?

src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
```diff
@@ -66,14 +66,86 @@ This is a `Torus` with `r_tube` having an inner radius of 1 and an outer radius
...
+end
+
+function Torus{T,CO}(r_torus, r_tube, φ, θ, origin, rotation) where {T,CO}
+    r_torus = _csg_convert_args(T, r_torus)
+    r_tube = _csg_convert_args(T, r_tube)
+    φ = _csg_convert_args(T, φ)
+    θ = _csg_convert_args(T, θ)
+    TT1, TT2 = _get_conemantle_type(θ)
+    Torus{T,CO,typeof(r_tube),typeof(φ),typeof(θ),TT1,TT2}(r_torus, r_tube, φ, θ, origin, rotation)
+end
```
lmh91, line 98: We only want one method here. That's a lot of code duplication.

We would need to add some general function to process and handle `φ` and `rotation` inside it.
We also need it for other primitives I believe.

lmh91 (changes requested):

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -83,22 +83,31 @@ struct Cone{T,CO,TR,TP<:Union{Nothing,T}} <: AbstractVolumePrimitive{T, CO}
...
+
 function Cone{T,CO}(r, φ, hZ, origin, rotation) where {T,CO}
-    _r = _csg_convert_args(T, r)
-    _φ = _csg_convert_args(T, φ)
-    _hZ = _csg_convert_args(T, hZ)
-    Cone{T,CO,typeof(_r),typeof(_φ)}(_r, _φ, _hZ, origin, rotation)
+    r = _csg_convert_args(T, r)
+    φ = _csg_convert_args(T, φ)
+    hZ = _csg_convert_args(T, hZ)
+    Cone{T,CO,typeof(r),typeof(φ)}(r, φ, hZ, origin, rotation)
```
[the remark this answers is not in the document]
lmh91: Might be. Maybe Julia is clever enough. Could be tested via `@code_warntype`.

One could just put those line directly into the constructor:
`Cone{T,CO,typeof(r),typeof(φ)}(_csg_convert_args(T, r),....`

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
```diff
@@ -83,42 +83,28 @@ struct Cone{T,CO,TR,TP<:Union{Nothing,T}} <: AbstractVolumePrimitive{T, CO}
...
-end
-
 function Cone{T,CO}(r, φ, hZ, origin, rotation) where {T,CO}
     r = _csg_convert_args(T, r)
     φ = _csg_convert_args(T, φ)
+    (_φ, rotation) = _handle_phi(φ, rotation)
     hZ = _csg_convert_args(T, hZ)
-    Cone{T,CO,typeof(r),typeof(φ)}(r, φ, hZ, origin, rotation)
+    Cone{T,CO,typeof(r),typeof(_φ)}(r, _φ, hZ, origin, rotation)
 end
```
lmh91, line 92: @fhagemann already commented using the same names (`r = _csg_convert_args(T, r); ...`) could create some type instabilities. I tested it and it does. You can see it yourself with `@code_warntype Cone{Float64,CO}(1, ...)`. 

We have to use different names for the variables, e.g. `_r = _csg_convert_args(T, r); ...`.

lmh91 (changes requested):

test/ConstructiveSolidGeometry/CSG_primitives.jl
lmh91: I addressed this issue in #287.

So in this PR, this issue can be ignored.

src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl
lmh91: @SebastianRuffert  could you use `RotZ(φ[1])` instead of writing all elements?

lmh91:
I just realized that we have two rotations in the constructor due to `φ`.
The order how they are applied is important. 
I think we did not specified that in the documentation so far, did we?

`rotation * RotZ(φ[1]) != RotZ(φ[1]) * rotation` is the "issue". 

We will have to decide for one and stick to it for all primitives and have to state it in the documentation. 

Any opinions on which order is the better one? In sense of that it is easier to construct certain geometries.

SebastianRuffert:
> I just realized that we have two rotations in the constructor due to `φ`. The order how they are applied is important. I think we did not specified that in the documentation so far, did we?
> 
> `rotation * RotZ(φ[1]) != RotZ(φ[1]) * rotation` is the "issue".
> 
> We will have to decide for one and stick to it for all primitives and have to state it in the documentation.
> 
> Any opinions on which order is the better one? In sense of that it is easier to construct certain geometries.

I think we should stick to rotation*RotZ(φ[1]). The user wants to construct a specific geometry ranging from one polar angle to the other one, afterwards he rotates it. Thus this order.

lmh91:
> > I just realized that we have two rotations in the constructor due to `φ`. The order how they are applied is important. I think we did not specified that in the documentation so far, did we?
> > `rotation * RotZ(φ[1]) != RotZ(φ[1]) * rotation` is the "issue".
> > We will have to decide for one and stick to it for all primitives and have to state it in the documentation.
> > Any opinions on which order is the better one? In sense of that it is easier to construct certain geometries.
> 
> I think we should stick to rotation*RotZ(φ[1]). The user wants to construct a specific geometry ranging from one polar angle to the other one, afterwards he rotates it. Thus this order.

Sounds reasonable. 
Could please make a note into the documentation in `/docs/src/man/csg.md`?
Maybe under `### Combination of Transformation`?

With some example maybe? Or extend the example of the tube by adding also some limitation in phi?

lmh91 (changes requested):

PR #289, opened 2022-05-10
Convenient constructors for the surface primitives

SebastianRuffert:


SebastianRuffert:
I did not add a constructor for Polygon, since I am unsure wether this one makes any sense. It has one field, which is a SVector containing N cartesian points. Therefor one can also directly call the struct itself and directly hand it the SVector

src/ConstructiveSolidGeometry/SurfacePrimitives/TorusMantle.jl
```diff
@@ -86,6 +85,7 @@ get_φ_limits(tm::TorusMantle{T,Nothing}) where {T} = T(0), T(2π)
 
 get_θ_limits(tm::TorusMantle{T,<:Any,Tuple{T,T}}) where {T} = tm.θ[1], tm.θ[2]
 get_θ_limits(tm::TorusMantle{T,<:Any,Nothing}) where {T} = T(0), T(2π)
+get_θ_limits(tm::TorusMantle{T,<:Any,T}) where {T} = T(0), tm.θ
```
fhagemann, line 88: Not sure if I remember correctly but I thought `θ` could never be just a `Float`...

src/ConstructiveSolidGeometry/SurfacePrimitives/TorusMantle.jl
```diff
@@ -22,14 +22,60 @@ Surface primitive describing the mantle of a [`Torus`](@ref).
...
+end
+
+function TorusMantle(D::Symbol=:outwards;
+    # define default parameters as Int64 to not influence type promotion
+    r_torus = 1, 
+    r_tube = 1,
+    φ = nothing,
+    θ = nothing,
+    origin = zero(CartesianPoint{Int64}), 
+    rotation = one(SMatrix{3, 3, Int64, 9})
```
fhagemann, line 64: Also here, you should choose `Int` over `Int64`

src/ConstructiveSolidGeometry/SurfacePrimitives/Plane.jl
```diff
@@ -9,10 +9,30 @@ Surface primitive describing a two-dimensional flat plane in three-dimensional s
 """
 struct Plane{T} <: AbstractPlanarSurfacePrimitive{T}
     origin::CartesianPoint{T}
-    normal::CartesianVector{T}
+    normal::CartesianPoint{T}
```
fhagemann, line 12: The `normal` vector of a `Plane` should not be a `CartesianPoint` but remain a `CartesianVector`

fhagemann:
As this PR is quite massive: please open a separate PR for each `SurfacePrimitive` (as discussed).

Issue #288, opened 2022-05-07
SSD cannot handle arbitrary contact IDs 


fhagemann:
Right now, a config file can only be read in, if the contacts of the detector are enumerated in order from `1` to `N` 
(where `N` is the number of contacts).

Choosing something different from this results in an error when trying to display the `Simulation`:
```yaml
detectors:
  - semiconductor:
      material: HPGe
      geometry: 
        # ...
    contacts:
      - material: HPGe
        potential: 2000
        id: 101
        geometry:
          # ...
      - material: HPGe
        potential: 0
        id: 102
        geometry:
          # ...
```
```julia
sim = Simulation("config_file.yaml")
```
```
Simulation{Float32, Cartesian} - Coordinate system: Cartesian
  Environment Material: Vacuum
  Detector: ExampleCuboid
  Electric potential: missing
  Charge density: missing
  Impurity scale: missing
  Fix Charge density: missing
  Dielectric distribution: missing
  Point types: missing
  Electric field: missing
  Weighting potentials: 
    Contact 101: Error showing value of type Simulation{Float32, Cartesian}:
ERROR: BoundsError: attempt to access 2-element Vector{Any} at index [101]
Stacktrace:
  [1] getindex(A::Vector{Any}, i1::Int64)
    @ Base ./array.jl:861
  [2] println(io::IOContext{Base.TTY}, sim::Simulation{Float32, Cartesian})
    @ SolidStateDetectors ~/.julia/packages/SolidStateDetectors/ETZbh/src/Simulation/Simulation.jl:134
  [3] show
    @ ~/.julia/packages/SolidStateDetectors/ETZbh/src/Simulation/Simulation.jl:142 [inlined]
  [4] show(io::IOContext{Base.TTY}, #unused#::MIME{Symbol("text/plain")}, sim::Simulation{Float32, Cartesian})
    @ SolidStateDetectors ~/.julia/packages/SolidStateDetectors/ETZbh/src/Simulation/Simulation.jl:145
  [5] (::REPL.var"#43#44"{REPL.REPLDisplay{REPL.LineEditREPL}, MIME{Symbol("text/plain")}, Base.RefValue{Any}})(io::Any)
    @ REPL /opt/julia-1.7.2/share/julia/stdlib/v1.7/REPL/src/REPL.jl:266
  [6] with_repl_linfo(f::Any, repl::REPL.LineEditREPL)
[...]

PR #290, opened 2022-05-12
Allow for contact_id to be any Int

SebastianRuffert:
This should solve issue #288

### DIFF
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -78,6 +78,14 @@ function NamedTuple(sim::Simulation{T}) where {T <: SSDFloat}
 end
 Base.convert(T::Type{NamedTuple}, x::Simulation) = T(x)
 
+function _viable_contact_types(contacts)
+    for contact_i in contacts
+        for contact_j in contacts
+            
+        end
+    end
+end
+
 function Simulation(nt::NamedTuple)
     missing_tuple = NamedTuple(missing)
     epot = ElectricPotential(nt.electric_potential)
@@ -101,13 +109,17 @@ function Simulation(nt::NamedTuple)
         ImpurityScale(nt.imp_scale)
     end
     sim.electric_field = haskey(nt, :electric_field) && nt.electric_field !== missing_tuple ? ElectricField(nt.electric_field) : missing
+    contact_ids = []
     sim.weighting_potentials = if haskey(nt, :weighting_potentials) 
         DataStructures.OrderedDict(let wp = Symbol("WeightingPotential_$(contact.id)")
             contact.id =>haskey(nt.weighting_potentials, wp) && getfield(nt.weighting_potentials, wp) !== missing_tuple ? WeightingPotential(getfield(nt.weighting_potentials,wp)) : missing 
+            append!(contact_ids,contact.id)
         end for contact in sim.detector.contacts)
     else
         DataStructures.OrderedDict(contact.id=>missing for contact in sim.detector.contacts)
+        append!(contact_ids,contact.id)
     end
+    @assert size(unique(contact_ids))==size(contact_ids) "Contact IDs must be unique"
     return sim
 end
 Base.convert(T::Type{Simulation}, x::NamedTuple) = T(x)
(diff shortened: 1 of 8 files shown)


fhagemann:
Thank you.
I saw that also changes in the I/O. Will this change allow to read in saved simulations (so no changes for the user)?

fhagemann:
And what happens if two contacts have the same ID in the config file?

fhagemann:
Can you add tests where the contact IDs are not in order?
And maybe also a case in which two contacts have the same ID and running the code fails.

SebastianRuffert:
> Thank you. I saw that also changes in the I/O. Will this change allow to read in saved simulations (so no changes for the user)?

Partly yes. Before, the different weighting potentials were initialized inside a list, while now that is being done in a dictionary, so now changes for the user in the rea-in of a config file. At least not ones, that I can think of now. 
What changes is the final struct. Meaning, if the user wants to access sim.weighting potentials a dict will be returned instead of a list.

SebastianRuffert:
> And what happens if two contacts have the same ID in the config file?

No error is thrown right now, when reading in a config file with two contacts of id: 1 and using `simulate()`
```
julia> sim = Simulation{Float32}("Software/SolidStateDetectors.jl/examples/example_config_files/contact_id_same_id.yaml")
julia> simulate!(sim)
Simulation{Float32, Cartesian} - Coordinate system: Cartesian
  Environment Material: High Purity Germanium
  Detector: Contact ID test file
  Electric potential: (58, 34, 34)
  Charge density: (58, 34, 34)
  Impurity scale: (58, 34, 34)
  Fix Charge density: (58, 34, 34)
  Dielectric distribution: (59, 35, 35)
  Point types: (58, 34, 34)
  Electric field: (58, 34, 34)
  Weighting potentials: 
    Contact 1: (14, 12, 12)
    Contact 1: (14, 12, 12)

```
But when accessing the weighting potentials directly, only one is shown
```
julia> sim.weighting_potentials
OrderedCollections.OrderedDict{Int64, Union{Missing, WeightingPotential}} with 1 entry:
  1 => [0.999674 0.999675 … 0.999684 0.999682; 0.999672 0.999676 … 0.999684 0.999681; … ; 0.999682 0.999686 … 0.999696 0.999692; 0.999684 0.999687 ……

```
How do we want to handle that case? We can throw an error if there is ambiguity in the contact ids. Another option is changing the ambigous contact id internally, maybe by adding one to the id or multiplying by 10.

lmh91:
I get an error in
```julia
sim = Simulation(SSD_examples[:ContactIDTest])
simulate!(sim)
calculate_capacitance_matrix(sim)
```

lmh91:
Also,
```julia
sim = Simulation(SSD_examples[:ContactIDTest])
simulate!(sim)
calculate_capacitance_matrix(sim)
```
does actually calculate some field and stores it into the dict. 
This should throw an error.

SebastianRuffert:
> ```julia
> sim = Simulation(SSD_examples[:ContactIDTest])
> simulate!(sim)
> calculate_weighting_potential!(sim, 4)
> ```

All calls of this function already use contact.id

lmh91:
> ```julia
> ContactIDTest
> ```

But there is no contact with `id == 4`. That's the issue.

src/Simulation/Simulation.jl
```diff
@@ -72,12 +72,20 @@ function NamedTuple(sim::Simulation{T}) where {T <: SSDFloat}
...
 end
 Base.convert(T::Type{NamedTuple}, x::Simulation) = T(x)
 
+function _viable_contact_types(contacts)
+    for contact_i in contacts
+        for contact_j in contacts
+            
+        end
+    end
+end
```
fhagemann, line 87: What is this function for?

SebastianRuffert: Forgot to delete, gone now

src/Simulation/Simulation.jl
```diff
@@ -550,7 +550,12 @@ function apply_initial_state!(sim::Simulation{T}, ::Type{WeightingPotential}, co
         paint_contacts, 
         point_types = depletion_handling ? sim.point_types : missing
     );
-    sim.weighting_potentials[contact_id] = WeightingPotential(ElectricPotentialArray(pcs), grid)
+    contact_ids = [contact.id for contact in sim.detector.contacts]
+    if contact_id in contact_ids
+        sim.weighting_potentials[contact_id] = WeightingPotential(ElectricPotentialArray(pcs), grid)
+    else
+        @error "No such contact"
+    end
```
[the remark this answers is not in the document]
SebastianRuffert: Done

Issue #287, opened 2022-05-06
Type instability in Torus constructor


lmh91:
This came up in #285.

The construction of a Torus is not entirely type stable as some parametric types of `Torus`
depend on the values and not the types of the input arguments to the constructor:
```julia
@inferred CSG.Torus{Float64}(φ=π,θ=(π,2π)) # -> error
```

The parametric types which cause the issue are `TT1` and `TT2` of the Torus: https://github.com/SebastianRuffert/SolidStateDetectors.jl/blob/de2af27789cf8883f74ad0599acc60fc11aadd9a/src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl#L69.
These define the types of two surfaces of the Torus. If we just remove `TT1` and `TT2` from the `Torus` we would shift the type instability into the function `surfaces(t::Torus)`. 

I think the best solution would be if we would came up with a way to remove the `:inwards` and `:outwards` parametric types of (all) surface primitives. Maybe by defining some field to those structs which only hold `+1` or `-1` in order to define the direction of surfaces like for the `ConeMantle` or `TorusMantle`. 

However, we would still have the issue of the `:flat` case for the `Torus`.
Here, one (or two) of the surfaces of the `Torus` become not a `ConeMantle` but an
`EllipticalSurface`: see `TorusThetaSurface`.
We would need to come up with a way to define these "flat" surfaces still as `ConeMantle`. In its current implementation of `ConeMantle`, this is not possible as `hZ` would be `0`.

PR #292, opened 2022-05-19
Add convenient constructor for `TorusMantle`

SebastianRuffert:


test/ConstructiveSolidGeometry/CSG_primitives.jl
```diff
@@ -205,4 +205,14 @@ no_translations = (rotation = one(SMatrix{3, 3, T, 9}), translation = zero(Carte
         @test !in(CartesianPoint{Float64}(1,1,3+tol),box_closed_trafo)
         @test in(CartesianPoint{Float64}(1,1,3-tol),box_open_trafo)
     end
+    @testset "TorusMantle" begin
+        # torusmantle1 = @inferred CSG.TorusMantle(r_torus=1f0)
```
SebastianRuffert, line 209: inferred does not work here due to D which can be :inwards or :outwards. I think we need to add two types instead of using symbols for inferred to work. Should we do that?

src/ConstructiveSolidGeometry/SurfacePrimitives/TorusMantle.jl
```diff
@@ -40,6 +85,7 @@ get_φ_limits(tm::TorusMantle{T,Nothing}) where {T} = T(0), T(2π)
 
 get_θ_limits(tm::TorusMantle{T,<:Any,Tuple{T,T}}) where {T} = tm.θ[1], tm.θ[2]
 get_θ_limits(tm::TorusMantle{T,<:Any,Nothing}) where {T} = T(0), T(2π)
+get_θ_limits(tm::TorusMantle{T,<:Any,T}) where {T} = T(0), tm.θ
```
SebastianRuffert, line 88: The docs only allow for theta to be a tuple or nothing. The constructor does not specifically forbid using a real here, and the usage of a real might come naturally to the user, considering that this functionality exists for phi. Therefor I think it makes sense to add this function.

src/ConstructiveSolidGeometry/SurfacePrimitives/TorusMantle.jl
```diff
@@ -22,14 +22,59 @@ Surface primitive describing the mantle of a [`Torus`](@ref).
...
+end
+
+function TorusMantle(D::Symbol=:outwards;
+    # define default parameters as Int64 to not influence type promotion
+    r_torus = 1, 
+    r_tube = 1,
+    φ = nothing,
+    θ = nothing,
+    origin = zero(CartesianPoint{Int64}), 
+    rotation = one(SMatrix{3, 3, Int64, 9})
```
fhagemann, line 64: Use `Int` instead of `Int64`

src/ConstructiveSolidGeometry/SurfacePrimitives/TorusMantle.jl
fhagemann: When we introduced real `φ` back in #240, that was mostly changing how `φ` is saved in the `struct` during simulations. Config files are usually still written as `from`-`to` fields. I am not sure whether the user knows what would happen if they pass only a single value. But I also think it does not harm to have this functionality here.

test/ConstructiveSolidGeometry/CSG_primitives.jl
fhagemann: @lmh91 ?

lmh91: Using types instead of symbols would not solve the issue. 

I think the best would be to remove `D` and add a field `normal_direction::T` (or some better name) to the struct.
Which either is `T(+1)` or `T(-1)`. 
This is related to #287. 

Just leave the test out for now.

lmh91 (approved):

PR #294, opened 2022-05-19
Add convenient constructor for  `EllipsoidMantle`

SebastianRuffert:


test/ConstructiveSolidGeometry/CSG_primitives.jl
```diff
@@ -15,17 +15,17 @@ no_translations = (rotation = one(SMatrix{3, 3, T, 9}), translation = zero(Carte
         for bot in (2.0, Dict("from" => 1.0, "to" => 2.0)),
             top in (2.0, 1.0, Dict("from" => 1.0, "to" => 2.0), Dict("from" => 2.0, "to" => 4.0)),
             φ in ((30,180),(0,360))
-        
+
```
fhagemann, line 18: Please avoid whitespace changes (can be deactivated in Atom)

src/ConstructiveSolidGeometry/SurfacePrimitives/EllipsoidMantle.jl
```diff
@@ -20,13 +20,42 @@ Surface primitive describing the surface of an [`Ellipsoid`](@ref).
...
+    EllipsoidMantle{T}(D, r, origin, rotation)
+end
+
+function EllipsoidMantle(D::Symbol=:outwards;
+    # define default parameters as Int to not influence type promotion
+    r = 1, 
+    origin = zero(CartesianPoint{Int}), 
+    rotation = one(SMatrix{3, 3, Int, 9}))
+    EllipsoidMantle(D, r, origin, rotation)
+end
```
fhagemann, line 51: Even though we do not allow for  `φ` and `θ` other than `Nothing`, it would still be nice to already have the code structure to pass them down to the core definition of `EllipsoidMantle`.

src/ConstructiveSolidGeometry/SurfacePrimitives/EllipsoidMantle.jl
```diff
@@ -20,13 +20,46 @@ Surface primitive describing the surface of an [`Ellipsoid`](@ref).
 * `origin::CartesianPoint{T}`: Origin of the `Ellipsoid` which has this `EllipsoidMantle` as surface.
 * `rotation::SMatrix{3,3,T,9}`: Rotation matrix of the `Ellipsoid` which has this `EllipsoidMantle` as surface.
 """
-@with_kw struct EllipsoidMantle{T,TR,TP,TT,D} <: AbstractCurvedSurfacePrimitive{T}
-    r::TR = 1
-    φ::TP = nothing
-    θ::TT = nothing
+struct EllipsoidMantle{T,TR,TP,TT,D} <: AbstractCurvedSurfacePrimitive{T}
```
fhagemann, line 23: Should we restrict `TP` and `TT` to `Nothing` here (as this is the only thing defined so far)?
And what about `TR` and `D`?

lmh91: `TP` and `TT` we could restrict for now. 

I would not restrict `TR` and `D`. In general, in Julia, you want to be as unrestricted as possible
to allow for possible compatibility between packages and their types.
(We should actually remove some restrictions at some point.)

lmh91 (approved):

PR #299, opened 2022-05-31
Fix bug in end of charge drift

SebastianRuffert:
When using a specific geometry, an error occured where several drift paths would go on forever. Specifically, this happened, whenever a thin contact was within the semiconductor. The simulation checks after each time step, wether the charge carrier is still inside the semiconductor and keeps on going if that is true.
This leads to bugs, when the contact is inside the semiconductor, because the simulation only checks, if the charge carrier is inside the semiconductor, but does not check, wether simultaneously it is inside a contact.


src/ChargeDrift/ChargeDrift.jl
```diff
@@ -80,12 +80,10 @@ function modulate_driftvector(sv::CartesianVector{T}, pt::CartesianPoint{T}, vdv
...
-@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cylindrical})::Bool where {T <: SSDFloat}
-    pt_cyl::CylindricalPoint{T} = CylindricalPoint(pt)
-    pt_cyl in point_types || pt_cyl in det.semiconductor
-end
-@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cartesian})::Bool where {T <: SSDFloat}
-    pt in point_types || pt in det.semiconductor
+@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, S})::Bool where {T <: SSDFloat,S}
+    cpt = _convert_point(pt, S)
+    point_type::PointType = point_types[cpt]
+    (point_type & bulk_bit > 0)
```
fhagemann, line 86: These three lines is exactly what `pt in point_types` does, see [PointTypes.jl#L63](https://github.com/SebastianRuffert/SolidStateDetectors.jl/blob/b013406e4e047f3d2f5791f2f77e9e94a29e5a9e/src/ScalarPotentials/PointTypes.jl#L63).

Also, if the `pt` does not have the `bulk_bit` set, we still might want to check the geometries, i.e. that the point is in the semiconductor, but **not** in the contacts, something like  `pt in det.semiconductor && !(pt in det.contacts)`.
Maybe we can do something with `CSGDifference`:
 `pt in sum(broadcast(s -> s.geometry, det.semiconductors)) - sum(broadcast(c -> c.geometry, det.contacts))`

src/ScalarPotentials/PointTypes.jl
```diff
@@ -74,7 +74,7 @@ function mark_bulk_bits!(point_types::Array{PointType, 3})
         point_types[i1,i2,i3] += bulk_bit * begin
             in_bulk = true
             for j1 in max(i1-1,1):min(i1+1,i1max), j2 in max(i2-1,1):min(i2+1,i2max), j3 in max(i3-1,1):min(i3+1,i3max)
-                in_bulk &= point_types[j1,j2,j3] & pn_junction_bit > 0
+                in_bulk &= (point_types[j1,j2,j3] & pn_junction_bit > 0) &&  (point_types[j1,j2,j3] & update_bit) > 0
```
fhagemann, line 77: Can you set the brackets in a consistent way?
(put the last bracket after `> 0`)

fhagemann (changes requested):

SebastianRuffert:
> @SebastianRuffert Could you also please provide a minimal example script which demonstrates the issue on the current main and disappears with your branch?

```julia
using Revise
using Plots; pyplot();
using SolidStateDetectors
using Unitful

detector_config_filename = "detector.json"
T = Float32
sim = Simulation{T}(detector_config_filename)
calculate_electric_potential!( sim, refinement_limits = [0.2, 0.1, 0.05, 0.01])
calculate_electric_field!(sim, n_points_in_φ = 72)
charge_drift_model = ADLChargeDriftModel()
sim.detector = SolidStateDetector(sim.detector, charge_drift_model);

spawn_positions = CartesianPoint{T}[]
idx_spawn_positions = []
x_axis = sim.electric_field.grid.axes[1][1:2:end]
z_axis = sim.electric_field.grid.axes[3][1:2:end]
for (i,x) in enumerate(x_axis)
    for (k,z) in enumerate(z_axis)
        push!(spawn_positions, CartesianPoint([x,0,z]))
        push!(idx_spawn_positions, CartesianIndex(i,k))
    end
end
length(spawn_positions)
in_idx = findall(x -> x in sim.detector && !in(x, sim.detector.contacts), spawn_positions);
ev = Event(spawn_positions[in_idx]);
time_step = 0.8u"ns"
drift_charges!(ev, sim, Δt = time_step, verbose = false)

DT = fill(NaN,length(x_axis),length(z_axis))
for (i, idx) in enumerate(idx_spawn_positions[in_idx])
    DT[idx] = length(ev.drift_paths[i].h_path)*ustrip(time_step)
end

heatmap(x_axis, z_axis, transpose(DT), aspect_ratio = 1)
```

using the config file:
```json
{
    "name": "detector",
    "units": {
        "length": "mm",
        "angle": "deg",
        "potential": "V",
        "temperature": "K"
    },
    "grid": {
        "coordinates": "cylindrical",
        "axes": {
            "r": {
                "to": 40.0,
                "boundaries": "inf"
            },
            "phi": {
                "from": 0,
                "to": 0,
                "boundaries": "periodic"
            },
            "z": {
                "from": -10,
                "to": 50.0,
                "boundaries": {
                    "left": "inf",
                    "right": "inf"
                }
            }
        }
    },
    "medium": "vacuum",
    "detectors": [
        {
            "semiconductor": {
                "material": "HPGe",
                "bulk_type": "p",
                "temperature": 77.0,
                "geometry": {
                    "tube": {
                        "r": 35.0,
                        "h": 40.0,
                        "origin": [
                            0.0,
                            0.0,
                            20.0
                            ]
                    }
                }

            },
            "contacts": [
                {
                    "name": "p contact",
                    "material": "HPGe",
                    "potential": 0.0,
                    "geometry": {
                        "tube": {
                            "r": 1.5,
                            "h": 0.1,
                            "origin": [
                                0.0,
                                0.0,
                                0.05
                            ]
                        }
                    },
                    "id": 1
                },
                {
                    "name": "n contact",
                    "material": "HPGe",
                    "potential": 4000.0,
                    "geometry": {
                        "union": [
                            {
                                "cone": {
                                    "r": {
                                        "bottom": {
                                            "from": 35.0,
                                            "to": 35.0
                                        },
                                        "top": {
                                            "from": 35.0,
                                            "to": 35.0
                                        }
                                    },
                                    "h": 35.0,
                                    "origin": [
                                        0.0,
                                        0.0,
                                        22.5
                                    ]
                                }
                            },
                            {
                                "tube": {
                                    "r": 35.0,
                                    "h": 0.0,
                                    "origin": [
                                        0.0,
                                        0.0,
                                        40.0
                                    ]
                                }
                            },
                            {
                                "cone": {
                                    "r": {
                                        "bottom": {
                                            "from": 30.0,
                                            "to": 30.0
                                        },
                                        "top": {
                                            "from": 35.0,
                                            "to": 35.0
                                        }
                                    },
                                    "h": 5.0,
                                    "origin": [
                                        0.0,
                                        0.0,
                                        2.5
                                    ]
                                }
                            }
                        ]
                    },
                    "id": 2
                }
            ]
        }
    ]  
}
```

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -80,12 +80,8 @@ function modulate_driftvector(sv::CartesianVector{T}, pt::CartesianPoint{T}, vdv
...
 modulate_driftvector(sv::CartesianVector{T}, pt::CartesianPoint{T}, vdv::Missing) where {T} = sv
 
-@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cylindrical})::Bool where {T <: SSDFloat}
-    pt_cyl::CylindricalPoint{T} = CylindricalPoint(pt)
-    pt_cyl in point_types 
-end
-@inline function _is_next_point_in_det(pt::CartesianPoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, Cartesian})::Bool where {T <: SSDFloat}
-    pt in point_types 
+@inline function _is_next_point_in_det(pt::AbstractCoordinatePoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, S})::Bool where {T <: SSDFloat, S}
+    pt in point_types || (pt in det.semiconductor && !(pt in det.contacts))
```
fhagemann, line 84: You need to convert the point to the right coordinates first before you check if it's in `point_types`.
For checking geometries, this should be fine as it is.

Maybe something like:
```julia
@inline function _is_next_point_in_det(pt::AbstractCoordinatePoint{T}, det::SolidStateDetector{T}, point_types::PointTypes{T, 3, S})::Bool where {T <: SSDFloat, S}
    _convert_point(pt, S) in point_types || (pt in det.semiconductor && !(pt in det.contacts))
end

**Review by fhagemann — CHANGES_REQUESTED** [review]

File: test/runtests.jl [hunk 4]
```diff
@@ -212,6 +212,39 @@ end
...
+    in_idx = findall(x -> x in sim.detector && !in(x, sim.detector.contacts), spawn_positions);
+    ev = Event(spawn_positions[in_idx]);
+    time_step = 0.8u"ns"
+    drift_charges!(ev, sim, Δt = time_step, verbose = false)
+
+    DT = fill(NaN,length(x_axis),length(z_axis))
+    for (i, idx) in enumerate(idx_spawn_positions[in_idx])
+        DT[idx] = length(ev.drift_paths[i].h_path)*ustrip(time_step)
+    end
+    @test maximum(DT[35:50,60:80])<100
```
fhagemann, line 244: If you only check part of `DT`, then you should only drift charges for these points
(not the full `x_axis` and `z_axis` but only the part you check)

fhagemann (changes requested):
The tests fail for the diffusion and self-repulsion case, when multiple charges are drifted.
It looks like not all holes reach the point contact before the charge drift is ended. This should not be.
Diffusion_SR

examples/example_config_files/isochrone_test.json
```diff
@@ -0,0 +1,133 @@
+{
+    "name": "Isochrone Bug",
```
fhagemann, line 2: The `"Isochrone Bug"` detector also has a weird geometry, the contacts are reaching into the bulk of the detector at the bottom of the detector.. I'm wondering why the tests do not fail for points in the region shielded from the point contact by the mantle contact 

Isochrone Test Detector

fhagemann (changes requested):

examples/example_config_files/isochrone_test.json
fhagemann: Ok, I guess it's because you only select part of `DT` in the tests and only look at the hole drift paths, not the electrons.

```julia
using SolidStateDetectors
using Unitful
using Test
using Plots; gr(fmt = :png)
T = Float32

sim = Simulation{T}(SSD_examples[:IsochroneTest])
simulate!(sim)
calculate_electric_potential!( sim, refinement_limits = [0.2, 0.1, 0.05, 0.01])
calculate_electric_field!(sim, n_points_in_φ = 36)
charge_drift_model = ADLChargeDriftModel()

spawn_positions = CartesianPoint{T}[]
idx_spawn_positions = []
x_axis = 0:0.0005:0.04
z_axis = -0.005:0.0005:0.045
spawn_positions = CartesianPoint{T}[]
idx_spawn_positions = CartesianIndex[]
for (i,x) in enumerate(x_axis)
    for (k,z) in enumerate(z_axis)
        push!(spawn_positions, CartesianPoint(T[x,0,z]))
        push!(idx_spawn_positions, CartesianIndex(i,k))
    end
end

in_idx = findall(x -> x in sim.detector && !in(x, sim.detector.contacts), spawn_positions);
length(in_idx)

ev = Event(spawn_positions[in_idx]);
time_step = T(1)u"ns"
max_nsteps = 10000
drift_charges!(ev, sim, Δt = time_step, max_nsteps = max_nsteps, verbose = false)

DT_h = Array{typeof(time_step),2}(fill(typeof(time_step)(NaN),length(x_axis),length(z_axis)))
DT_e = Array{typeof(time_step),2}(fill(typeof(time_step)(NaN),length(x_axis),length(z_axis)))
for (i, idx) in enumerate(idx_spawn_positions[in_idx])
    DT_h[idx] = length(ev.drift_paths[i].h_path)*time_step
    DT_e[idx] = length(ev.drift_paths[i].e_path)*time_step
end
```

```julia
# Holes
contourf(vcat(-x_axis[end:-1:2], x_axis)*1000u"mm", z_axis*1000u"mm", DT_h[vcat(end:-1:2,1:end),:]', 
    ratio = 1, 
    size = (500,500), colorbar_title = "Driftime", unitformat = :slash, 
    xlabel = "x", ylabel = "z", clims = (0,Inf))
```
Holes


```julia
# Electrons
contourf(vcat(-x_axis[end:-1:2], x_axis)*1000u"mm", z_axis*1000u"mm", DT_e[vcat(end:-1:2,1:end),:]', 
    ratio = 1, 
    size = (500,500), colorbar_title = "Driftime", unitformat = :slash, 
    xlabel = "x", ylabel = "z", clims = (0,Inf))
```
Electrons

fhagemann:
With my last two commits, I have done the following:
- fa2f3a7: I have updated the isochrone config file:
  - I converted the JSON file to a YAML file to match the format of all other config files
  - I adapted the geometry of the mantle contact to the surface of the `Tube` which makes up the bulk of the detector.
  - I have added a impurity density, because otherwise drift times might be close to infinite
  - I have made the detector a bit smaller
  
image

- c03b622: I have updated the isochrone tests in `runtests.jl`:
  - Now, both the electron and hole drift times are calculated
  - We check if all drifts are shorted than the maximum allowed drift time (`max_nsteps * time_step`)
  - We check close to the point contact that the hole drift times are small (`< 200u"ns"`)
  - We check far away from the point contact that the hole drift times are large (`> 1000u"ns"`)

fhagemann:
The isochrone plots also look fine, I would propose to add them to the tutorial or documentation in a dedicated PR:

## Holes
image

## Electrons
image

## Both, holes and electrons
image

fhagemann (approved):

PR #302, opened 2022-06-14
Add convenient constructor for points

SebastianRuffert:


test/ConstructiveSolidGeometry/CSG_primitives.jl
```diff
@@ -278,4 +278,12 @@ no_translations = (rotation = one(SMatrix{3, 3, T, 9}), translation = zero(Carte
         CSG.ConeMantle()
         CSG.ConeMantle(φ=(1.,2.))    
     end
+    @testset "Point" begin
+        cart = @inferred CSG.CartesianPoint(x=2f0,z=1f0)
+        @inferred CSG.CartesianPoint{Float32}(x=2)
+        cyl = @inferred CSG.CylindricalPoint{Float32}(r=2.,z=1.)
+        cyl2 = @inferred CSG.CylindricalPoint(φ=3π)
+        @test CartesianPoint(cyl) == cart
+        @test cyl2.φ == T(π)
```
fhagemann, line 287: Here, you could also check if the type conversion of `CartesianPoint` to `CylindricalPoint` works, something like
```julia
@test CartesianPoint(x = 1) = CartesianPoint(CylindricalPoint(r = 1))
```
(not sure if this works right of the box. though.

fhagemann (approved):

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -24,6 +24,35 @@ struct CartesianPoint{T} <: AbstractCoordinatePoint{T, Cartesian}
...
+    eltypes = _csg_get_promoted_eltype.((TX,TY,TZ))
+    T = float(promote_type(eltypes...))
+    CartesianPoint{T}(T(x),T(y),T(z))
+end
+
+function CartesianPoint(x::TX, y::TY, z::TZ) where {TX<:Int,TY<:Int,TZ<:Int}
+    eltypes = _csg_get_promoted_eltype.((TX,TY,TZ))
+    T = float(promote_type(eltypes...))
+    CartesianPoint{T}(T(x),T(y),T(z))
+end
```
fhagemann, line 38: I do not see the difference between these two...
The upper one should do exactly the same thing as the lower one, or am I missing something?

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -45,10 +74,39 @@ struct CylindricalPoint{T} <: AbstractCoordinatePoint{T, Cylindrical}
...
+    eltypes = _csg_get_promoted_eltype.((TR,TP,TZ))
+    T = float(promote_type(eltypes...))
+    CylindricalPoint{T}(T(r),T(φ),T(z))
+end
+
+function CylindricalPoint(r::TR, φ::TP, z::TZ) where {TR<:Int,TP<:Int,TZ<:Int}
+    eltypes = _csg_get_promoted_eltype.((TR,TP,TZ))
+    T = float(promote_type(eltypes...))
+    CylindricalPoint{T}(T(r),T(φ),T(z))
+end
```
fhagemann, line 92: Same thing here: why do you need both?

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
SebastianRuffert: If the second on is missing, CartesianPoint() goes directly to the struct, which therefor will be a CartesianPoint{Int64}

fhagemann: Well, we introduced `CartesianPoint{Int}` to not be dominant in type promotion for other primitives.
With this method, `CartesianPoint{Int}` will always result in `CartesianPoint{Float64}` (because `float(Int64) = Float64`), which might cause volume primitives defined in `Float32` to become of type `Float64`.. I wonder why/if the tests fail.

lmh91: I think we only need the second method. Just use `<:Real` instead `<:Int`

lmh91:
Yes (that was also the case before but now with only one additional method).
`CartesianPoint(1, 2, 3)` will result in `CartesianPoint{Float64}` but
`CartesianPoint{Int}(1, 2, 3)` will result in `CartesianPoint{Int}`.

PR #303, opened 2022-06-14
Add convenient constructor for vector

SebastianRuffert:


src/ConstructiveSolidGeometry/PointsAndVectors/Vectors.jl
```diff
@@ -21,6 +21,35 @@ struct CartesianVector{T} <: AbstractCoordinateVector{T, Cartesian}
...
+    eltypes = _csg_get_promoted_eltype.((TX,TY,TZ))
+    T = float(promote_type(eltypes...))
+    CartesianVector{T}(T(x),T(y),T(z))
+end
+
+function CartesianVector(x::TX, y::TY, z::TZ) where {TX<:Int,TY<:Int,TZ<:Int}
+    eltypes = _csg_get_promoted_eltype.((TX,TY,TZ))
+    T = float(promote_type(eltypes...))
+    CartesianVector{T}(T(x),T(y),T(z))
+end
```
fhagemann, line 35: Why do you need two functions? Shouldn't the upper one do what the lower one does?

src/ConstructiveSolidGeometry/PointsAndVectors/Vectors.jl
```diff
@@ -60,4 +89,31 @@ struct CylindricalVector{T} <: AbstractCoordinateVector{T, Cylindrical}
...
+    eltypes = _csg_get_promoted_eltype.((TR,TP,TZ))
+    T = float(promote_type(eltypes...))
+    CylindricalVector{T}(T(r),T(φ),T(z))
+end
+
+function CylindricalVector(r::TR, φ::TP, z::TZ) where {TR<:Int,TP<:Int,TZ<:Int}
+    eltypes = _csg_get_promoted_eltype.((TR,TP,TZ))
+    T = float(promote_type(eltypes...))
+    CylindricalVector{T}(T(r),T(φ),T(z))
+end
```
fhagemann, line 103: Again

fhagemann (commented):
The name of your commits are a bit misleading, as you change the  `Vectors` here, not the `Points`.

src/ConstructiveSolidGeometry/PointsAndVectors/Vectors.jl
lmh91: Same as for the `CartesianPoint` I guess.

fhagemann:
I would also propose to delete `CylindricalVector` completely, as we only define and export it
but it is NEVER used anywhere.

PR #305, opened 2022-06-15
Constructor with units

SebastianRuffert:


SebastianRuffert:
This is a very straightforward way to implement units. Do we want to allow for units in both constructor functions or only the one I changed so far?

SebastianRuffert:
Also I am unsure if we should convert all default values to the user-specified unit, if only one keyword argument is given

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
```diff
@@ -70,10 +70,18 @@ function Box(CO, hX::TX, hY::TY, hZ::TZ, origin::PT, rotation::ROT) where {TX, T
...
+    Box(CO, _hX, _hY, _hZ, origin, rotation)
+end
+
 function Box(::Type{CO}=ClosedPrimitive;
-    hX = 1,
-    hY = 1,
-    hZ = 1,
+    hX = 1u"m",
+    hY = 1u"m",
+    hZ = 1u"m",
```
fhagemann, line 84: SSD does not know which function to call.
Try to specify in the already existing function above this one that `TX`, `TY` and `TZ` can only be a `Real`.

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
```diff
@@ -70,6 +70,15 @@ function Box(CO, hX::TX, hY::TY, hZ::TZ, origin::PT, rotation::ROT) where {TX, T
     Box{T}(CO,T(hX), T(hY), T(hZ), origin, rotation)
 end
 
+#Unit convesion
+Box(CO, hX::Quantity, hY, hZ, origin, rotation) = Box(CO, to_internal_units(hX), hY, hZ, origin, rotation)
```
SebastianRuffert, line 74: Is there a convenient way to write those permutations?

fhagemann: Use `RealQuantity`?

SebastianRuffert: Is `RealQuantity` part of Unitful? I can not find it

fhagemann: `RealQuantity` is part of SSD:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/9dabbf63f698fb1c73b03f8b1fd7218fdf1a1e23/src/Units.jl#L90

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
```diff
@@ -71,12 +71,13 @@ function Box(CO, hX::TX, hY::TY, hZ::TZ, origin::PT, rotation::ROT) where {TX, T
...
-    _hZ = to_internal_units(hZ)
-    Box(CO, _hX, _hY, _hZ, origin, rotation)
-end
+Box(CO, hX::Quantity, hY, hZ, origin, rotation) = Box(CO, to_internal_units(hX), hY, hZ, origin, rotation)
+Box(CO, hX, hY::Quantity, hZ, origin, rotation) = Box(CO, hX, to_internal_units(hY), hZ, origin, rotation)
+Box(CO, hX, hY, hZ::Quantity, origin, rotation) = Box(CO, hX, hY, to_internal_units(hZ), origin, rotation)
+Box(CO, hX::Quantity, hY::Quantity, hZ, origin, rotation) = Box(CO, to_internal_units(hX), to_internal_units(hY), hZ, origin, rotation)
+Box(CO, hX, hY::Quantity, hZ::Quantity, origin, rotation) = Box(CO, hX, to_internal_units(hY), to_internal_units(hZ), origin, rotation)
+Box(CO, hX::Quantity, hY, hZ::Quantity, origin, rotation) = Box(CO, to_internal_units(hX), hY, to_internal_units(hZ), origin, rotation)
+Box(CO, hX::Quantity, hY::Quantity, hZ::Quantity, origin, rotation) = Box(CO, to_internal_units(hX), to_internal_units(hY), to_internal_units(hZ), origin, rotation)
```
fhagemann, line 80: I am not a fan of these permutations...

What is the problem here and why doesn't this work without permutations?
`to_internal_units` also works for numbers without units and just returns the number:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/9dabbf63f698fb1c73b03f8b1fd7218fdf1a1e23/src/ConstructiveSolidGeometry/Units.jl#L8

This could be fixed by `RealQuantity` but I see that `RealQuantity` is only defined in SSD and not CSG and we are modifying CSG functions. Maybe it makes sense to define `RealQuantity` in CSG already and export it?

Then, the following might (?) work:
```julia
function Box(CO, hX::RealQuantity, hY::RealQuantity, hZ::RealQuantity, origin, rotation)
    _hX = to_internal_units(hX)
    _hY = to_internal_units(hY)
    _hZ = to_internal_units(hZ)
    Box(CO, _hX, _hY, _hZ, origin, rotation)
end

```

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -109,6 +109,8 @@ end
 @inline _convert_point(pt::AbstractCoordinatePoint, ::Type{Cylindrical}) = CylindricalPoint(pt)
 @inline _convert_point(pt::AbstractCoordinatePoint, ::Type{Cartesian}) = CartesianPoint(pt)
 
+scale(cart::CartesianPoint{T}, fact) where T = CartesianPoint{T}(cart.x*fact, cart.y*fact, cart.z*fact)
```
fhagemann, line 112: I would be careful with this!
If we look at the docstring, `CartesianPoint` will **always** be saved in units of `m`.
I would prefer to have a constructor with units for `CartesianPoint` and solve this via the constructor and not with `scale`..

src/ConstructiveSolidGeometry/VolumePrimitives/Box.jl
```diff
@@ -80,14 +89,16 @@ function Box(::Type{CO}=ClosedPrimitive;
...
     hX = 1.0,
     hY = 1.0,
     hZ = 1.0,
     origin = zero(CartesianPoint{Float64}), 
     rotation = one(SMatrix{3, 3, Float64, 9})
-) where {T, CO}
-    Box{T}(CO, hX, hY, hZ, origin, rotation)
+) where {T, CO, UN<:Unitful.Units}
+    unit_factor = ustrip(uconvert(unt,1u"m"))
+    Box{T}(CO, hX/unit_factor, hY/unit_factor, hZ/unit_factor, scale(CartesianPoint{T}(origin),1/unit_factor), rotation)
```
fhagemann, line 101: `(origin)` here does not need brackets, but as said above: maybe there is another way of implementing this.

lmh91:
I don't think we should do this like this.

Let's take just the `CartesianPoint` as an example (its the same for everything else including the primitives): 
```julia
CartesianPoint(1u"m", 2u"m", 3u"m")
```
should return `CartesianPoint{promote_type(typeof.(x, y, y))}` (pseudo code) and definitely not convert 
to `CartesianPoint{Int/Float}`. 

Instead we have to add methods to `Unitful.ustrip` and `Unitful.uconvert` for our types.
`uconvert` might be a bit difficult as our types have different fields. I am not sure if there is a general syntax for that in Unitful.jl.
Maybe there is also a function like `unify_units` in Unitful.jl for cases like that. Or this should be handled in the promotion.
E.g. `(1u"m", 2u"mm)"` will be promoted to `(1.0u"m", 0.002u"m")`. 
We have to look into Unitful.jl if there are already some conventions for this. 

If not, we could (in SSD but not in the submodule CSG), define some `_ssd_uconvert`-methods for our types and
then just `ustrip` them with the `ustrip` methods which we can define in the CSG submodule.

lmh91:
I just tested on the current main branch:
```julia
julia> CartesianPoint(1u"m", 2u"m", 3u"m")
3-element CartesianPoint{Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}} with indices SOneTo(3):
 1 m
 2 m
 3 m
```
That works already out of the box as we do not limit `T` for `AbstractGeometry`.
```julia 
julia> CartesianPoint(1u"m", 2u"m", 3u"m")
3-element CartesianPoint{Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}} with indices SOneTo(3):
 1 m
 2 m
 3 m
```
And there seems also already to be some promotion rules defined in Unitful.jl as the second case also works :)

lmh91 (commented):
Hmm, this topic is a bit tricky. 

In principle, we (Julia ecosystem) don't want to convert a Quantity by default to its underlying type. This is should be done by implementing a method for `ustrip` for `Box`. 
So a primitive with units could be constructed (and the information of the units 
are still attached to the primitive and not removed like it is implemented right now).

In order to do this we would need to change the structs of the primitives though.
We would need to separate the type of the `rotation` from the types of the other fields 
like `hX`:
```julia
struct Box{T,CO,RT} <: AbstractVolumePrimitive{T,CO}
    hX::T # so `T` could also be a Quantity here
    hY::T
    hZ::T
    origin::CartesianPoint{T}
    rotation::SMatrix{3,3,RT,9}
end
```

We also would need to change the definition of the `CylindricalPoint` by the way 
as we will need different types for `φ` and `r`&`z`.

This would be in principle the best (julianic) way in my opinion. 
However, it requires the most work as well and make things a bit more complicated (more parametric), though not impossible. 

Or, we force the `ustrip` in the construction and make it impossible to construct primitives with units still attached like it is the design in this PR right now. 
But, then we should also do this for the point types like `CartesianPoint`. For consistency. 
And for the points, I would really like to have them with units attached :thinking: 

Hmm...

src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
```diff
@@ -50,18 +50,21 @@ module ConstructiveSolidGeometry
 
     abstract type AbstractConstructiveGeometry{T} <: AbstractGeometry{T} end
     
-    _csg_convert_args(eltype::Type{T}, r::Real) where T = convert(T, r) 
+    _csg_convert_args(eltype::Type{T}, r::Real) where T = convert(T, r)
     _csg_convert_args(eltype::Type{T}, r::Tuple) where T = broadcast(x -> _csg_convert_args(T, x), r)
     _csg_convert_args(eltype::Type{T}, r::Nothing) where T = nothing
+    _csg_convert_args(eltype::Type{T}, r::Quantity) where T = convert(T,ustrip(uconvert(u"m",r)))
```
fhagemann, line 56: This only works for `LengthQuantity`, not for angles etc.

src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
```diff
@@ -64,10 +67,15 @@ module ConstructiveSolidGeometry
...
-    _csg_get_promoted_eltype(::Type{Quan}) where {Quan<:Quantity}= Quan.parameters[1]
+    _csg_get_promoted_eltype(::Type{Q}) where {Q <: Quantity}= Q
     
     _handle_phi(φ, rotation) = (φ, rotation)
     _handle_phi(φ::Tuple, rotation) = (abs(φ[2]-φ[1]), rotation*RotZ(φ[1]))
+    
+    _precision_type(::Type{T}) where {T <: Real} = T
+    _precision_type(::Type{Quantity{T}}) where {T <: Real} = T
+    
+    float(::Type{Quantity{T}}) where T = T
```
fhagemann, line 78: Careful! This is type piracy!
`Quantity` is a type that SSD does not own, so we should be **VERY** careful when extending the `Base.float` method

test/runtests.jl
```diff
@@ -13,13 +13,13 @@ using Unitful
...
-@testset "SOR GPU Backend" begin
-    include("SOR_GPU_Backend.jl")
-end
+# @testset "Comparison to analytic solutions" begin
+#     include("comparison_to_analytic_solutions.jl")
+# end
+# 
+# @testset "SOR GPU Backend" begin
+#     include("SOR_GPU_Backend.jl")
+# end
```
fhagemann, line 22: Please don't exclude tests without mentioning why you exclude them.

src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
```diff
@@ -38,7 +38,7 @@ module ConstructiveSolidGeometry
     const CartesianTicksTuple{T} = NamedTuple{(:x,:y,:z), NTuple{3,Vector{T}}}
     const CylindricalTicksTuple{T} = NamedTuple{(:r,:φ,:z), NTuple{3,Vector{T}}}
 
-    abstract type AbstractGeometry{T <: AbstractFloat} end
+    abstract type AbstractGeometry{T <: Number} end
```
fhagemann, line 41: Why `Number` and not `AbstractFloat`?

PR #308, opened 2022-06-21
Add convenient constructor for line

SebastianRuffert:


src/ConstructiveSolidGeometry/LinePrimitives/Line.jl
```diff
@@ -1,8 +1,29 @@
-struct Line{T} <: AbstractLinePrimitive{T}
+struct Line{T<:AbstractFloat} <: AbstractLinePrimitive{T}
```
lmh91, line 1: `<:AbstractFloat` is already in `abstract type AbstractGeometry{T <: AbstractFloat} end`.
Please remove it here.

src/ConstructiveSolidGeometry/LinePrimitives/Line.jl
```diff
@@ -1,8 +1,29 @@
...
     origin::CartesianPoint{T}
     direction::CartesianVector{T}
 end
 
+#Type promotion happens here
+function Line(origin::PT, direction::DIR) where {PT<:(CartesianPoint), DIR<:(CartesianVector)}
```
lmh91, line 7: I don't think we need brackets here around the types`PT <: (CartesianPoint)` -> `PT <: CartesianPoint, ...`

lmh91 (changes requested):

src/ConstructiveSolidGeometry/LinePrimitives/Line.jl
SebastianRuffert: That is true, but when calling Line() the compiler will skip the type promotion and call directly the struct. This leads to an error, since now Int is handed to the struct.
If I explicitly state T<:AbstractFloat, Line() will call the type promotion function

lmh91 (approved):

PR #309, opened 2022-06-21
Add convenient constructor for edge

SebastianRuffert:


src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
```diff
@@ -1,8 +1,29 @@
-struct Edge{T} <: AbstractLinePrimitive{T}
+struct Edge{T<:AbstractFloat} <: AbstractLinePrimitive{T}
```
lmh91, line 1: `<:AbstractFloat` is already in `abstract type AbstractGeometry{T <: AbstractFloat}` end.
Please remove it here.

src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
```diff
@@ -1,8 +1,29 @@
...
     a::CartesianPoint{T}
     b::CartesianPoint{T}
 end
 
+#Type promotion happens here
+function Edge(a::A, b::B) where {A<:(CartesianPoint), B<:(CartesianPoint)}
```
lmh91, line 7: I don't think we need brackets here around the types `PT <: (CartesianPoint), ...` -> `PT <: CartesianPoint, ...`

lmh91 (changes requested):

src/ConstructiveSolidGeometry/LinePrimitives/Edge.jl
SebastianRuffert: see discussion in #308

fhagemann: Well `T` should be at least a `RealQuantity`... 
I see no setting in which it would make sense for `T` to be a `String`, for example...

lmh91: Julia: Never restrict types if not necessary. 
This allows for different packages to work together without knowing each other. 
The only reason to restrict types is for dispatching in general. 

I could live with `Number` which is supposed to be the supertype of all types which have something to do with units. 
It's also the supertype of a Quantity. 
Another example would be if people use the CSG objects in automatic differentiation. 
Those packages have special types for Numbers, e.g. `ForwardDiff.Dual`. 

But since there is no reason to restrict, as we do not get any benefit from it, we should not restrict at all.

Issue #253, opened 2022-01-27
Expansion of grids to full `2π` for cylindrical grids with `φmin ≠ 0` is incorrect


fhagemann:
Related to #215.

The expansion of grids to full `2π` for cylindrical grids with `φmin ≠ 0` is incorrect.

MWE:
Using the example configuration file for the inverted coaxial detector and changing the `phi` range for the `grid`
```yaml
name: Public Inverted Coax
# ...
grid:
  coordinates: cylindrical
  axes:
    r: # ...
    phi:
      from: 30  # not zero !
      to: 90
      boundaries: periodic
    z: # ...
#...
```

```julia
using SolidStateDetectors
sim = Simulation(SSD_examples[:InvertedCoax]) # with the modified config file => phi: [30, 90]
calculate_electric_potential!(sim)
calculate_electric_field!(sim)
```

For the electric potential, the `φ` axis seems to be parsed correctly:
```julia
sim.electric_potential.grid[2]
0.5235988..1.5707964 (closed–open) - length = 4

sim.electric_potential.grid[2].ticks
4-element Vector{Float32}:
 0.5235988
 0.7853982
 1.0471976
 1.3089969
```

However, for the electric field, it is expanded to `2π` and it looks like the interval for the `φ` axis is always `[0, 2π)`, while the `ticks` seem to exceed `2π`. This, the interval and the ticks of the axis contradict each other.

```julia
sim.electric_field.grid[2]
0.0..6.2831855 (closed–open) - length = 24

sim.electric_field.grid[2].ticks
24-element Vector{Float32}:
 0.5235988
 0.7853982
 1.0471976
 1.3089969
 1.5707964
 1.8325958
 2.0943952
 2.3561945
 2.6179938
 2.8797934
 3.1415927
 3.403392
 3.6651917
 3.926991
 4.1887903
 4.4505897
 4.712389
 4.9741883
 5.2359877
 5.4977875
 5.7595863
 6.0213857
 6.283185
 6.544985
```

This results in an error when plotting (but I guess there are also errors when trying to interpolate the electric field for the charge drift.)
```julia
using Plots
plot(sim.electric_field, full_det = true)
plot_electric_fieldlines!(sim, full_det = true)
```

```
knot-vectors must be unique and sorted in increasing order

Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] check_gridded
[...]

PR #310, opened 2022-06-23
Fix tick bug when expanding electric potential to full 2pi

SebastianRuffert:
This should solve issue #253

### DIFF
--- a/examples/example_config_files/cone_config.yaml
+++ b/examples/example_config_files/cone_config.yaml
@@ -1,0 +1,71 @@
+name: "Cut Cone"
+units:
+  length: "mm"
+  angle: "deg"
+  potential: "V"
+  temperature: "K"
+grid:
+  coordinates: "cylindrical"
+  axes:
+    r:
+      to: 12.0
+      boundaries: "inf"
+    phi:
+      from: 13
+      to: 13
+      boundaries: "periodic"
+    z:
+      from: -5
+      to: 15
+      boundaries:
+        left: "inf"
+        right: "inf"
+medium: "vacuum"
+detectors:
+  - semiconductor:
+      material: "HPGe"
+      bulk_type: "p"
+      temperature: 77.0
+      impurity_density:
+        name: "constant"
+        value: 1e7cm^-3
+      geometry:
+        cone:
+          r:
+            bottom:
+              from: 1.0
+              to: 2.0
+            top:
+              from: 1.0
+              to: 10.0    
+          h: 10.
+          origin:
+            z: 5
+    contacts:
+      - name: "p contact"
+        id: 1
+        material: "HPGe"
+        potential: 4000.
+        geometry:
+          tube:
+            r: 1.
+            h: 10.
+            origin:
+              z: 5
+      - name: "Outer Contact 1"
+        id: 2
+        material: "HPGe"
+        potential: 0
+        geometry:
+          tube:
+            r:
+              bottom:
+                from: 2.0
+                to: 2.5
+              top:
+                from: 10.
+                to: 10.5
+            h: 10.
+            origin:
+              z: 5
+      
--- a/src/ScalarPotentials/ScalarPotential.jl
+++ b/src/ScalarPotentials/ScalarPotential.jl
@@ -41,7 +41,7 @@ function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::Discret
     @assert int.right != 0 "Right boundary of φ interval is not allowed to be 0"
     l::Int = length( axφ )
     Δφ::AT = width(int)
-    #new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(0, 1)
+    new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(0, 2π)
     n::Int = Int(round(T(2π) / Δφ, sigdigits = 6))
     new_ticks::Vector{AT} = Vector{AT}(undef, l * n)
     new_pot = Array{eltype(sp.data), 3}(undef, size(sp, 1), l * n, size(sp, 3))
@@ -53,7 +53,7 @@ function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::Discret
         end
     end
     P = sortperm(new_ticks)
-    new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(new_ticks[P][1], new_ticks[P][end])
+    #new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(new_ticks[P][1], new_ticks[P][end])
     new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks[P] )
     new_axes = (sp.grid[1], new_axφ, sp.grid[3])
     new_grid = Grid{AT, 3, Cylindrical, typeof(new_axes)}( new_axes )
@@ -100,7 +100,7 @@ end
 function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}; n_points_in_φ::Union{Missing, Int} = missing) where {T}
     axφ::DiscreteAxis{T} = sp.grid[2]
     int::Interval = axφ.interval
-    if length(axφ) == 1 # 2D
+    if int.right == 0 && length(axφ) == 1 # 2D
         if ismissing(n_points_in_φ)
             error(ArgumentError, ": First Argument is a 2D potential (only 1 point in φ). User has to set the keyword `n_points_in_φ::Int` (e.g. `n_points_in_φ = 18`) in order to get a 3D potential.")
         else
--- a/src/ScalarPotentials/ScalarPotential.jl
+++ b/src/ScalarPotentials/ScalarPotential.jl
@@ -41,7 +41,7 @@ function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::Discret
     @assert int.right != 0 "Right boundary of φ interval is not allowed to be 0"
     l::Int = length( axφ )
     Δφ::AT = width(int)
-    new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(0, 2π)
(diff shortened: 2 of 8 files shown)


lmh91:
The field calculations seems to work but I can't plot the electric potential and electric field in r-phi:
```julia
plot(sim.electric_potential, z= 0)
```
Could you please have a look?

fhagemann:
Yes, the ticks should be ordered, and ideally, the first tick of a `DiscreteAxis` should also be the left boundary of the respective `interval`, i.e.
```julia
g = Grid(sim)
g.φ.ticks[1] == g.φ.interval.left
```

src/ScalarPotentials/ScalarPotential.jl
```diff
@@ -48,14 +48,15 @@ function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::Discret
...
             new_pot[:, idx, :] = sp[:, il, :]
         end
     end
-    new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks )
+    P = sortperm(new_ticks)
+    new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks[P] )
     new_axes = (sp.grid[1], new_axφ, sp.grid[3])
     new_grid = Grid{AT, 3, Cylindrical, typeof(new_axes)}( new_axes )
-    ScalarPotential(sp, new_pot, new_grid)
+    ScalarPotential(sp, new_pot[:,P,:], new_grid)
```
fhagemann, line 59: Also here, `g.φ.ticks[1]` is not necessarily equal to `g.φ.interval.left`.
Please compose an example where `g.φ.ticks` does not include `0` or multiples of `2pi` (probably a 2D symmetric detector from `phi: [17°,17°]` or something like this and check whether the expansion of the potential to full `2pi` affects the calculation, plotting, etc.
It would also be good to add such a case with a non-zero left `phi` entry for the `grid` in the config file to the tests.

fhagemann (changes requested):

src/ScalarPotentials/ScalarPotential.jl
SebastianRuffert: I already added such a geometry and will write tests. Though when using a 2D geometry we only test `extend_2D_to_3D_by_n_points`. We are not testing `get_2π_potential` for periodic boundary conditions. What is not yet implemented is `get_2π_potential` for reflecting boundaries, which should also be tested.

src/ScalarPotentials/ScalarPotential.jl
```diff
@@ -81,24 +83,26 @@ end
...
     Δφ::AT = 2π / n_points_in_φ
     for i in 1:n_points_in_φ
-        new_ticks[i] = (i - 1) * Δφ
+        new_ticks[i] = mod((i - 1) * Δφ + sp.grid[2][1],T(2π))
         new_pot[:, i, :] = sp[:, 1, :]
     end
-    new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks )
+    P = sortperm(new_ticks)
+    new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(new_ticks[P][1], new_ticks[P][end])
+    new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks[P] )
```
fhagemann, line 96: Then, this is also not needed (or needs to be modified according to my first comment.

src/ScalarPotentials/ScalarPotential.jl
```diff
@@ -41,21 +41,23 @@ function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::Discret
...
             idx::Int = il + (idx_n - 1) * l
-            new_ticks[idx] = (idx_n - 1) * Δφ + axφ.ticks[il]
+            new_ticks[idx] = mod((idx_n - 1) * Δφ + axφ.ticks[il], T(2π))
             new_pot[:, idx, :] = sp[:, il, :]
         end
     end
-    new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks )
+    P = sortperm(new_ticks)
+    new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(new_ticks[P][1], new_ticks[P][end])
+    new_axφ = DiscreteAxis{AT, :periodic, :periodic}( new_int, new_ticks[P] )
```
fhagemann, line 57: I do not think that this is what we want...

The endpoints of the `Interval` should not necessarily be the first and last ticks of `new_ticks[P]`, especially when the `Interval` is `:open` on the right.

In this function (`get_2π_potential`), we want to end up with a `2π` potential,and in my personal opinion the `new_int` should also be a `2π`-`Interval`, i.e. where the `:closed`-`:open`-`Interval` ranges from `[0,2π[` (or `[x,x+2π[` but then again there might be values outside of the `[0,2π[` range...)

src/ScalarPotentials/ScalarPotential.jl
```diff
@@ -81,24 +83,26 @@ end
...
 function extend_2D_to_3D_by_n_points(sp::ScalarPotential{T, 3, Cylindrical}, axφ::DiscreteAxis{AT, :reflecting, :reflecting}, int::Interval{:closed, :closed, AT},
         n_points_in_φ::Int ) where {T, AT}
-    new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(0, 2π)
+    #new_int::Interval{:closed, :open, AT} = Interval{:closed, :open, AT}(0, 2π)
     new_ticks::Vector{AT} = Vector{AT}(undef, n_points_in_φ)
     new_pot = Array{eltype(sp.data), 3}(undef, size(sp, 1), n_points_in_φ, size(sp, 3))
     Δφ::AT = 2π / n_points_in_φ
     for i in 1:n_points_in_φ
-        new_ticks[i] = (i - 1) * Δφ
+        new_ticks[i] = mod((i - 1) * Δφ + sp.grid[2][1],T(2π))
```
fhagemann, line 91: If we start from a `2D` potential, I do not think that we need this modification. We can just use the previous code and have the electric potential range from `[0,2π[`. (Also, I do not see why we would need to include the original `2D` tick value in the `new_ticks` for the `3D` potential here, as everything will be `φ`-symmetric).

examples/example_config_files/cone_config.yaml
```diff
@@ -0,0 +1,71 @@
...
+              from: 1.0
+              to: 10.0    
+          h: 10.
+          origin:
+            z: 5
+    contacts:
+      - name: "p contact"
+        id: 1
+        material: "HPGe"
+        potential: 4000.
```
fhagemann, line 48: Minor comment: p+ contacts usually have a negative potential applied

examples/example_config_files/cone_config.yaml
```diff
@@ -0,0 +1,71 @@
...
+      from: -5
+      to: 15
+      boundaries:
+        left: "inf"
+        right: "inf"
+medium: "vacuum"
+detectors:
+  - semiconductor:
+      material: "HPGe"
+      bulk_type: "p"
```
fhagemann, line 27: `bulk_type`, at least to my knowledge, should be ignored (?)

src/ScalarPotentials/ScalarPotential.jl
SebastianRuffert: That is correct, I did this mainly for consistency reasons.

fhagemann:
Why was this closed?

lmh91:
@fhagemann What is the status of this? 
It looks good to me. Can this be merged?

fhagemann:
I did not yet look into this in detail, might get to this next week. What I see is that we added new config files without corresponding test. It might make sense to add tests for a phi interval not starting at 0 to make sure that everything runs smoothly before merging this into main.

fhagemann:
If I have a look at the new config file introduced in this PR (`ConeSym`), the point types look non-2π-symmetric for some `z`, e.g.

```julia
using SolidStateDetectors, Unitful, Plots
sim = Simulation(SSD_examples[:ConeSym])
plot(sim.point_types, z = 0.009)
```
image

Not sure exactly what the problem is here (maybe missing important ticks?) but this is something to look into before merging.

fhagemann:
> image
> 
> Not sure exactly what the problem is here (maybe missing important ticks?) but this is something to look into before merging.

This happens to be due to rounding errors 
(one tick at `0.00575` and one at `0.00625`, and the intersection is at `0.006` or `0.0059999`..)

src/ScalarPotentials/ScalarPotential.jl
```diff
@@ -48,14 +48,35 @@ function get_2π_potential(sp::ScalarPotential{T, 3, Cylindrical}, axφ::Discret
...
+        new_axes = (sp.grid[1], new_axφ, sp.grid[3])
+        new_grid = Grid{AT, 3, Cylindrical, typeof(new_axes)}( new_axes )
+        ScalarPotential(sp, new_pot[:,P,:], new_grid)
+    else
+        new_ticks_new_zero::Vector{AT} = Vector{AT}(undef, l * n + 1)
+        new_ticks_new_zero[1] = T(0)
+        new_ticks_new_zero[2:end] = new_ticks[P]
+        new_pot_new_zero = Array{eltype(sp.data), 3}(undef, size(sp, 1), l * n + 1, size(sp, 3))
+        new_pot_new_zero[:,1,:] = (eltype(sp.data) <: AbstractFloat) ? sp[:,1,:] + 
+                        (sp[:,end,:] - sp[:,1,:]) * (T(0) - new_ticks_new_zero[2]) / (T(2π) - T(new_ticks_new_zero[end]) * n + T(new_ticks_new_zero[2])) :
```
fhagemann, line 67: I feel like this formula is wrong and it should contain something like `Δφ * n` rather than `T(new_ticks_new_zero[end]) * n`.. I will do the calculation and see if this needs to be adjusted.

fhagemann: The permutation of the new potential based on the permutation of the new ticks is missing here, I will push some changes to fix this..

PR #311, opened 2022-06-23
Depletion voltage with local min max search

SebastianRuffert:


SebastianRuffert:
I did not add any tests yet. How do we want to test `get_depletion_voltage`? Should we test it for one of the analytic geometries or do we want to roughly compare it to, for example, the BEGe measurements

fhagemann:
> I did not add any tests yet. How do we want to test `get_depletion_voltage`? Should we test it for one of the analytic geometries or do we want to roughly compare it to, for example, the BEGe measurements

I would suggest comparing it to analytical solutions, and maybe also checking what the method yields for our standard IVC simulation and adding that as well (just to see if any code changes also influence the value of the depletion voltage).

I wouldn't include the BEGe measurements as we do not know the exact impurity profile of that detector.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,152 @@
...
+    The accuracy of the result depends on the precision of the initial simulation.
+    
+See also [`is_depleted`](@ref).
+"""
+function get_depletion_voltage(sim::Simulation{T}, contact_id::Int,
+            potential_range::AbstractRange = range(extrema(broadcast(c -> abs(c.potential), sim.detector.contacts))..., length = 1001);
+            verbose = true)::T where {T <: AbstractFloat}
+    
+    @assert is_depleted(sim.point_types) "This method only works for fully depleted simulations. Please increase the potentials in the configuration file to a greater value."
+    @assert sum(potential_range.>=0)==length(potential_range) "Please use a strictly positive potential range"
```
SebastianRuffert, line 60: I screwed up and did not put this into a seperate commit. When allowing negative ranges, the calls to range parameters would need to be flipped in that case (going form first entry to last is wrong if range goes from -150:0. We would need to start at 0 and go down). On the other side it should not really make a difference, when using a purely positive range. Therefor I decided to restrict the range to positive potentials.

fhagemann: But what if the potential needed to deplete the detector (e.g. at the p++ contact of a p-type detector) is negative? 

Can you maybe check if the values are negative and then modify the range with a step of -1 if the range should reach into the negative?

SebastianRuffert: Sure, I forgot about doping

fhagemann:
Can you briefly describe what this algorithm does now?

SebastianRuffert:
`get_depletion_voltage` works as follows: First we calculate the voltage, where the global minimum or maximum is on a contact, and not inside the semiconductor anymore. This was already implemented in the PR #250, where a description is provided. In the same PR, the correct remark was made, that this condition is not sufficient, due to the possibilities of local minima or maxima inside the semiconductor. In a fully depleted detector, those should not show up either.
### Local Min Max search
The voltage, which is produced by the global min max search is used as starting voltage for the local search. Now we loop over all points inside the semiconductor, and for each point gradually increase the voltage. This is done, until the specific point is neither a local minimum or maximum. We check that by comparing the center potential, to all nearest neighbour potentials. The maximum voltage of all such points will be our depletion voltage.

test/runtests.jl
```diff
@@ -43,6 +43,7 @@ T = Float32
         @test isapprox( signalsum, T(2), atol = 5e-3 )
         nt = NamedTuple(sim)
         @test sim == Simulation(nt)
+        #@test isapprox(get_depletion_voltage(sim, 1, 1:5000, verbose = false), T(3053), atol = 2.0)
```
SebastianRuffert, line 46: There is an effect especially prominent for 2D geometries, which results in varying depletion voltages, depending on the refinement parameters from simulate!. When choosing smaller convergence limits or refinement limits, there is fluctuation in the electric potential, which affects the local min max search. This test for example yields 3053V for default `simulate!` parameters, and 440533V for the parameters chosen above.
How should we deal with that? Do we expect the user to know about this effect or should we adapt the algorithm, in order to be less sensitive to noise?

fhagemann: These values seems way to high to me... I would expect that the depletion voltage would be of the order of 2000V and not fluctuate that much. 

What I think happens here is that the detector is not biased in reversed bias mode.
In the config file, we apply 0V to the core (id: 1) and 3500V to the mantle (id: 2) contact. 
What happens in this algorithm is that we set all contacts to 0V and then vary the potential that we apply to the core contact (id: 1). In this scenario, we would need to apply a **NEGATIVE** voltage to the core. 

Can you try if you choose the range from -1:-1:-5000 if the result is still unstable?

fhagemann: Otherwise, you would need to call
`get_depletion_voltage(sim, 2, 1:5000)` (for the mantle which has id: 2)

docs/src/tutorials/complete_simulation_chain_IVC_lit.jl
```diff
@@ -90,6 +90,11 @@ is_depleted(sim_undep.point_types)
 println("Depleted:   ", get_active_volume(sim.point_types))
 println("Undepleted: ", get_active_volume(sim_undep.point_types));
 
+# The depletion voltage can also be estimated from the simulation with the depleted detector.
+# For this, the user has to choose a contact to which to apply the voltage, as well as a range to probe:
+
+get_depletion_voltage(sim, 2, 1700:0.5:2000)
```
fhagemann, line 96: This is by the way also something that we do in the tutorial (with contact id: 2)

test/comparison_to_analytic_solutions.jl
```diff
@@ -24,6 +24,13 @@ struct DummyImpurityDensity{T} <: SolidStateDetectors.AbstractImpurityDensity{T}
     @testset "Capacity" begin
         @test all(isapprox.(C_ssd, C_Analytical, rtol = 0.001))
     end   
+    
+    @testset "Depletion Voltage" begin
+        dep_volt_1 = get_depletion_voltage(sim, 1, verbose = false)
+        dep_volt_2 = get_depletion_voltage(sim, 1, verbose = false)
+        @test isapprox(dep_volt_1, 0, atol = 0.2)
+        @test isapprox(dep_volt_2, 0, atol = 0.2)
+    end
```
fhagemann, line 33: The depletion voltage is given by the impurity density of the crystal.
In this example, the detector has no impurity density.

Is this test to show that you can instantly deplete a detector if it has no impurities in it (a trivial test)?

test/runtests.jl
```diff
@@ -95,6 +96,7 @@ T = Float32
         @test isapprox( signalsum, T(2), atol = 5e-3 )
         nt = NamedTuple(sim)
         @test sim == Simulation(nt)
+        @test isapprox(get_depletion_voltage(sim, 1, verbose = false), T(99), atol = 0.2) #13.5V depletion voltage was simulated before. This makes sure code changes do not affect the computation.
```
fhagemann, line 99: I am a bit surprised about this one because the `BEGe` has no impurity density defined.
I would expect the depletion voltage of this detector to be zero...

Two options:
- Easy fix for this: can you delete this test?
- More work: can you look into what happened here? Why isn't the depletion voltage zero?

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+!!! warn
+    This method only works if the initial `sim` was calculated for a case in which the detector is fully depleted.
+
+!!! note
+    The accuracy of the result depends on the precision of the initial simulation.
+    
+See also [`is_depleted`](@ref).
+"""
+function get_depletion_voltage(sim::Simulation{T}, contact_id::Int,
+            potential_range::AbstractRange = range(extrema(broadcast(c -> c.potential, sim.detector.contacts))..., length = 1001);
```
fhagemann, line 56: The problem with `potential_range` is the sign.
We have seen this in the case of the inverted coaxial point contact detector: you can achieve depletion by applying a negative potential of around -2000V to the core, or by applying a positive potential of around +2000V to the mantle.

In the config file, the potential applied to the core is 0V, whereas the potential applied to the mantle is +3500V.
Thus, `potential_range` is ranging from 0 to +3500V. However, when we look for the potential that we have to apply to the **CORE** to deplete the detector, we would expect the result to be negative and, thus, not in the range from 0 to +3500V. 

It might make sense to ensure that the user do not keep on increasing the range to unreasonably high voltages for cases in which the detector is not operated in reversed bias voltage. 

Any thought on how to deal with this?

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+            verbose = true)::T where {T <: AbstractFloat}
+    
+    @assert is_depleted(sim.point_types) "This method only works for fully depleted simulations. Please increase the potentials in the configuration file to a greater value."
+    @assert first(potential_range) * last(potential_range) >= 0 "Please search for depletion voltage either in a fully positive or a fully negative range. Currently you were looking in the range
```
fhagemann, line 60: The range should be negative for p++ contacts and positive for n++ contacts.
However, we do not force the user to decide what kind of contacts they have defined.

We could make a rough assignment by looking at the provided potential values and making all contacts with positive potentials n++ contacts and all contacts with negative potentials p++ (and for potentials of 0, we would have to check what is left, I guess).

Thoughts?

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+    
+    @showprogress for U in potential_range
+        ϕmin, ϕmax = extrema((ϕρ .+ T(U) * sim.weighting_potentials[contact_id].data)[inside])
+        depleted = ϕmax - ϕmin < abs(U)
+        if (initial_depletion && !depleted) || (!initial_depletion && depleted)
+            if T(U)>=0
+                start_local_search = T(U)
+            else
+                start_local_search = T(U)
+            end
```
fhagemann, line 100: This looks like an `if`-`else`-statement which is not needed as it does the same for all cases. 
Is this needed? Can we get rid of this?

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+        Ix = CartesianIndex(1,0,0)
+        Iy = CartesianIndex(0,1,0)
+        Iz = CartesianIndex(0,0,1)
+        neighbours = [Ix,-Ix,Iy,-Iy,Iz,-Iz]
+    end
+
+    scale = start_local_search
+    undepleted = 0
+    for (i,idx) in enumerate(inside)
+        if !(0 in [size_data[j] == 1 || idx[j] != 1 && idx[j] != size_data[j] for j in 1:length(idx)])#idx[1]!=1 && idx[1]!=size_data[1] && idx[2]!=1 && idx[2]!=size_data[2] && idx[3]!=1 && idx[3]!=size_data[3]
```
fhagemann, line 121: Are you sure that this statement is executed correctly?
You might want to add brackets around statements to ensure that `&&` and `||` are not mixed up.

Is it:
- `(size_data[j] == 1) || (idx[j] != 1 && idx[j] != size_data[j])` or
- `(size_data[j] == 1 || idx[j] != 1) && (idx[j] != size_data[j])` ?

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+                        local_range = range(scale, last(potential_range), step = step(potential_range))
+                    end         
+                    break
+                end
+                if scale_loc == local_range[end]
+                    undepleted+=1
+                end         
+            end
+        else
+            deleteat!(inside,i) #Delete indices which are in bulk bit, but still have non defined neighbour (Maybe we can get rid of this, I just kept it for safety)
```
fhagemann, line 147: Can you comment on this? When did you experience this?

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+        end
+    end
+    if initial_depletion
+        depletion_voltage = potential_range[1]
```
fhagemann, line 151: Let's assume the depletion voltage is +3000V and we test something above, let's say from +4000V to +5000V.
Then, this would return that the depletion voltage is +4000V.

I would be more in favour if the result would be `NaN` such that...

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,164 @@
...
+    elseif undepleted == 0
+        depletion_voltage = scale
+    end
+    
+    if verbose
+        if !isnan(depletion_voltage)
+            @info "The depletion voltage of the detector is ($(depletion_voltage) ± $(T(step(potential_range)))) V applied to contact $(contact_id)."
+        else
+            @warn "The depletion voltage is not in the specified range $(extrema(potential_range).*u"V")."
+        end
```
fhagemann, line 161: ... this message is printed out (that it is **NOT** in the range).

Then, I would also remove the tests for which detectors with no impurity density yield a depletion voltage of 0V.

src/Simulation/Simulation.jl
```diff
@@ -80,27 +80,32 @@ Base.convert(T::Type{NamedTuple}, x::Simulation) = T(x)
...
+                * depleted point -> imp_scale = 1;  
+            """
+            ImpurityScale(T.(.!is_undepleted_point_type.(sim.point_types.data)), sim.point_types.grid)
+        else
+            ImpurityScale(nt.imp_scale)
+        end
+        sim.electric_field = haskey(nt, :electric_field) && nt.electric_field !== missing_tuple ? ElectricField(nt.electric_field) : missing
     else
-        ImpurityScale(nt.imp_scale)
+        T = Float32
```
fhagemann, line 106: Unrelated to depletion voltages but should be make `T` a keyword argument of this function rather than a hard coded variable in the inside of the function? What if the user wants the `Simulation` in `Float64`.

fhagemann (changes requested):

src/Simulation/DepletionVoltage.jl
fhagemann: I checked, it does the right thing (if `size_data[j]` only has one entry of if we are at the beginning or end)

src/Simulation/DepletionVoltage.jl
fhagemann: Naively, to me it looks like this statements deletes all `i` if one of the axes has length 1.
There might also be a more sophisticated way how to treat `r = 0` in the cylindrical case.

test/runtests.jl
fhagemann: A good test would also be to check if the result is somewhat reasonable.

You can save the output of the depletion voltage and simulate the detector 10V above and below the depletion voltage with `depletion_handling = true` and check whether it matches the expected depletion behavior.

Something like
```julia
# Determine depletion voltage
id = 2
deplV = get_depletion_voltage(sim, contact_id)

# Simulate the detector 10V below depletion voltage => should be undepleted
sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = deplV - 10)
calculate_electric_potential!(sim, depletion_handling = true)
@test !is_depleted(sim.point_types)

# Simulate the detector 10V above the depletion voltage => should be depleted
sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = deplV + 10)
calculate_electric_potential!(sim, depletion_handling = true)
@test is_depleted(sim.point_types)
```
```

Comment (fhagemann) [reply test/runtests.jl:46]: So, for the inverted coaxial detector, this should be quite a good approximation:
```julia
simulate!(sim, depletion_handling = true, refinement_limits = [0.2,0.1,0.05,0.03,0.01])
U = get_depletion_voltage(sim, 2)
```
gives a depletion voltage of `U = 1855.0f0`.

If we use this depletion voltage for the simulation, i.e.
```julia
sim.detector = SolidStateDetector(sim.detector, contact_id = 2, contact_potential = U)
calculate_electric_potential!(sim, depletion_handling = true, refinement_limits = [0.2,0.1,0.05,0.03,0.01])
plot(sim.point_types, full_det = true)
```
image

whereas if we simulate the detector with 20V more voltage, we fully deplete it:
```julia
sim.detector = SolidStateDetector(sim.detector, contact_id = 2, contact_potential = U + 20)
calculate_electric_potential!(sim, depletion_handling = true, refinement_limits = [0.2,0.1,0.05,0.03,0.01])
plot(sim.point_types, full_det = true)
```
image

SebastianRuffert: You were right, for a negative voltage at contact 1, the depletion voltage is more reasonable. Still there is a rather big difference in the depletion voltage between the (more refined) default refinement parameters and the parameters used during testing. Default parameters yield -1827V and test parameters -1923V

test/comparison_to_analytic_solutions.jl
SebastianRuffert: Yes, that was the goal

test/runtests.jl
SebastianRuffert: I will look into this. I was wondering what was happening here too

src/Simulation/DepletionVoltage.jl
SebastianRuffert: We can get rid of this. This was just an artifact

src/Simulation/DepletionVoltage.jl
SebastianRuffert: No, if the phi axis has length 1, the neighbours are not even checked. This occurs for example in the InvertedCoax, where inside includes indices of the form [1,#,#]. Then a neighbour at [0,#,#] is checked, which throws an error. This is due to the fact that grid points at the boundary of the grid are marked as bulk bits. Therefor those are also inside

fhagemann: I don't understand.. If the phi axis would not have length 1, we would still have points inside with [1,#.#] or not?

SebastianRuffert: Yes, that is correct. Everything I wrote after the first sentence is completely uncorrelated to the phi axis length.
The issue is, that e.g. [1,#,#] is inside. Because when looking for local min or max, I compare this point to all its neighbours. Meaning also the non-existent neighbour at index [0,#,#]. This throws an error. Therefor I would like [1,#,#] not to be inside

fhagemann: And how about just checking all the neighbours except for [0,#,#]?
(So, checking if a neighbour exists and check all the existing neighbours)

SebastianRuffert: That is possible too. This would also detect local min or max at the world boundary, so maybe slightly better

test/runtests.jl
SebastianRuffert: Turns out, that the code could not figure out a depletion voltage of 0, since the criterium in the global search is that max-min < applied voltage. If the applied voltage is 0, the noisy left side will always be bigger. I changed the check to maximum(applied_voltage, eps) where eps=1e-3 to fix this

src/Simulation/DepletionVoltage.jl
```diff
@@ -114,12 +114,13 @@ function get_depletion_voltage(sim::Simulation{T}, contact_id::Int,
...
     undepleted = 0
+    eps = 1e-1
     length_idx = length(inside[1])
     for (i,idx) in enumerate(inside)
         for scale_loc in local_range
             center_pot = T(scale_loc) * sim.weighting_potentials[contact_id].data[idx] + ϕρ[idx]
-            min_pot = T(scale_loc) * sim.weighting_potentials[contact_id].data[idx-Ix] + ϕρ[idx-Ix]
-            max_pot = min_pot
+            min_pot = center_pot + eps
+            max_pot = center_pot - eps
```
fhagemann, line 123: Does this take into account that `center_pos` can be either positive or negative?

SebastianRuffert: nope, will fix that

test/runtests.jl
fhagemann: I guess the difference also depend on the refinement limits.
The default, I believe, stops at a refinement limit of 5%, whereas the test have a refinement limit of 10%.
Therefore, I would also assume a 5% uncertainty `(-1827 ± 91)V` on the depletion voltage for the default parameters and a 10% uncertainty `(-1923 ± 192)V` on the test parameters.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+                end         
+                break
+            end
+            if scale_loc == local_range[end] # If local scale is iterated all the way throughout the local range, this grid point is undepleted.
+                undepleted+=1
+            end         
+        end
+        if (undepleted > 0) # Break loop as soon as one grid point is undepleted
+            break
+        end
```
fhagemann, line 147: Why not already break when increasing `undepleted`?

src/Simulation/DepletionVoltage.jl
```diff
@@ -25,13 +25,12 @@ Returns the potential (in V) needed at the [`Contact`](@ref) with id `contact_id
...
 increased or decreased according to the `potential_range`. The depletion voltage
-is set to the potential for which a previously undepleted detector becomes depleted,
-or for which a previously depleted detector becomes undepleted.
+is set to the potential for which a previously undepleted detector becomes depleted.
 
 ## Arguments 
 * `sim::Simulation{T}`: [`Simulation`](@ref) for which the depletion voltage should be determined.
 * `contact_id::Int`: The `id` of the [`Contact`](@ref) at which the potential is applied.
-* `potential_range::AbstractRange`: Range of potentials to be tested.
+* `potential_range::AbstractRange`: Range of potentials to be tested. Must be strictly positive or negative.
```
fhagemann, line 33: Is `0` positive or negative for you?
Or is `0` not allowed at all?

src/Simulation/DepletionVoltage.jl
```diff
@@ -25,13 +25,12 @@ Returns the potential (in V) needed at the [`Contact`](@ref) with id `contact_id
 to fully deplete the detector in a given [`Simulation`](@ref). For this, all other
 contact potentials are set to `0` and the potential at the specified contact is
 increased or decreased according to the `potential_range`. The depletion voltage
-is set to the potential for which a previously undepleted detector becomes depleted,
-or for which a previously depleted detector becomes undepleted.
```
fhagemann: I would like to have this method work like explained in the previous docstring.
The potential range should be allowed to be increasing or decreasing, meaning that we can either start depleting an undepleted detector or undepleting a depleted detector.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+    ϕmin, ϕmax = extrema((ϕρ .+ potential_range[1] * sim.weighting_potentials[contact_id].data)[inside])
+    eps = 1e-3 * sign( potential_range[length(potential_range)÷2] )
+    initial_depletion::Bool = ϕmax - ϕmin < maximum((abs(potential_range[1]), eps))
+    depletion_voltage::T = NaN
+    start_local_search::T = 0
+    
+    @showprogress for U in potential_range
+        ϕmin, ϕmax = extrema((ϕρ .+ T(U) * sim.weighting_potentials[contact_id].data)[inside])
+        depleted = ϕmax - ϕmin < abs(U)
+        if (initial_depletion && !depleted) || (!initial_depletion && depleted)
```
fhagemann, line 95: It also somewhat looks like we would allow both cases of an increasing/decreasing potential range here.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+See also [`is_depleted`](@ref).
+"""
+function get_depletion_voltage(sim::Simulation{T}, contact_id::Int,
+            potential_range::AbstractRange = range(extrema(broadcast(c -> c.potential, sim.detector.contacts))..., length = 1001);
+            verbose = true)::T where {T <: AbstractFloat}
+    
+    @assert is_depleted(sim.point_types) "This method only works for fully depleted simulations. Please increase the potentials in the configuration file to a greater value."
+    @assert first(potential_range) * last(potential_range) >= 0 "Please search for depletion voltage either in a fully positive or a fully negative range. Currently you were looking in the range
+        from $(first(potential_range)) to $(last(potential_range))"
+    potential_range = (last(potential_range) > 0 && first(potential_range) >= 0) || (step(potential_range) < 0) ? potential_range : reverse(potential_range)
```
fhagemann, line 61: As commented before, I do not see why we should limit the `potential_range` to an increasing positive one or a decreasing negative one. We might wanna compare the execution speeds though and add a `@warn` or `@info` what is preferred (I guess from undepleted to depleted as the check for undepleted points is now ended once one point is undepleted). Also, it might make sense to check the depletion status at the beginning and at the end of `potential_range` and demand that it changes such that we can be sure that the expected depletion voltage is within that range.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+    ϕρ = deepcopy(sim.electric_potential.data)
+    for c in sim.detector.contacts
+        if c.potential == 0 continue end
+        ϕρ .-= c.potential * sim.weighting_potentials[c.id].data
+    end
+    
+    inside = findall(p -> p & bulk_bit > 0 && p & update_bit > 0, sim.point_types.data)
+    
+    ϕmin, ϕmax = extrema((ϕρ .+ potential_range[1] * sim.weighting_potentials[contact_id].data)[inside])
+    eps = 1e-3 * sign( potential_range[length(potential_range)÷2] )
```
fhagemann, line 87: Can you comment on this?
`potential_range[length(potential_range)÷2]` is just the mean value of the potential range?

This, by the way, throws an error if `length(potential_range) == 0` or `length(potential_range) == 1`, which might result if the step is defined in a wrong way, e.g. `-1500:1:-2000`. We might wanna take care that this does not happen or that the error thrown is more meaningful that this right now.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+            if scale_loc == local_range[end] # If local scale is iterated all the way throughout the local range, this grid point is undepleted.
+                undepleted+=1
+            end         
+        end
+        if (undepleted > 0) # Break loop as soon as one grid point is undepleted
+            break
+        end
+    end
+    if initial_depletion
+        @info "Detector is already depleted at the start of the specified voltage range ($(potential_range))! Please find a starting voltage where the detector is not yet depleted."
```
fhagemann, line 150: This can be removed if both increasing and decreasing ranges are supported

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+    elseif undepleted == 0
+        depletion_voltage = scale
+    end
+    
+    if verbose
+        if !isnan(depletion_voltage)
+            @info "The depletion voltage of the detector is ($(depletion_voltage) ± $(T(step(potential_range)))) V applied to contact $(contact_id)."
```
fhagemann, line 157: We might add to this `@info` that the error of the depletion voltage also depends on the accuracy (refinement and convergence limits) of the simulation.

src/Simulation/DepletionVoltage.jl
```diff
@@ -0,0 +1,163 @@
...
+        else
+            @warn "The depletion voltage is not in the specified range $(extrema(potential_range).*u"V")."
```
fhagemann, line 159: This is something that should already be checked at the beginning of the function where the depletion status at the edges of the `potential_range` are checked. If they are equal: print this warning and already return `NaN`.

test/runtests.jl
```diff
@@ -43,6 +43,18 @@ T = Float32
...
+        @info deplV
+        @test isapprox(deplV, T(1904), atol = 2.0) 
+        # Check wether detector is undepleted, 10V below the previously calculated depletion voltage
+        sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = deplV - 40)
+        calculate_electric_potential!(sim, depletion_handling = true)
+        @test !is_depleted(sim.point_types)
+        # Check wether detector is depleted, 10V above the previously calculated depletion voltage
+        sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = deplV + 40)
+        calculate_electric_potential!(sim, depletion_handling = true)
+        @test is_depleted(sim.point_types)
```
fhagemann, line 57: We might wanna use finer `refinement_limits` for this test to get closer to the "real" depletion voltage and maybe really only check 10V above and below (rather than the 40V used here).

src/Simulation/DepletionVoltage.jl
SebastianRuffert: In this case it would work, but at the moment the plan is to allow both going from depleted to undepleted and also from undepleted to depleted. Maybe this can be done there as well, but not sure yet

src/Simulation/DepletionVoltage.jl
SebastianRuffert: 0 is both negative and positive

src/Simulation/DepletionVoltage.jl
SebastianRuffert: I only want the sign of the potential range. If I would use first or last of potential range this might be sign(0), which does not differentiate between a negative and a positive range

test/runtests.jl
SebastianRuffert: I think this won`t solve the issue, because the underlying problem is, that the point types are fully depleted, when the global min max search is done. Still the local min max search returns something bigger than that. This delta will always determine how much below we will have to check, and this delta can actually be quite high (around 50V sometimes)

src/Simulation/DepletionVoltage.jl
```diff
@@ -21,155 +21,138 @@ end
...
-    
-    ϕmin, ϕmax = extrema((ϕρ .+ potential_range[1] * sim.weighting_potentials[contact_id].data)[inside])
-    eps = 1e-3 * sign( potential_range[length(potential_range)÷2] )
-    initial_depletion::Bool = ϕmax - ϕmin < maximum((abs(potential_range[1]), eps))
-    depletion_voltage::T = NaN
-    start_local_search::T = 0
+function estimate_depletion_voltage(
+        sim::Simulation{T}, 
+        contact_id::Int;
+        field_sim_settings::NamedTuple = (verbose = false,)
```
fhagemann, line 51: This doesn't seem to work well, if you wanna change verbose from the outside.
I guess this is done to set the default value of verbose to `false`?

Is this intended that one needs to know the name of the keyword argument and pass the settings as `NamedTuple`?

src/Simulation/DepletionVoltage.jl
```diff
@@ -21,155 +21,138 @@ end
...
+        grid_size, 
+        pts, 
+        imp_field, 
+        zero_imp_field
+    ) where {T}
+    @inbounds begin
+        for i3 in 1:grid_size[3]
+            for i2 in 1:grid_size[2]
+                for i1 in 1:grid_size[1]
+                    if SolidStateDetectors.is_pn_junction_point_type(pts[i1, i2, i3])
```
fhagemann, line 130: avoid prefix `SolidStateDetectors` in the package (not only here)

src/Simulation/DepletionVoltage.jl
lmh91: It's supposed to allow for changing any field simulation settings. 
Verbose was only chosen as default to restrict the argument to NamedTuples.
The `merge` call will overwrite preset settings defined in the function body.

I guess we could turn it into an argument with default value instead.

src/Simulation/DepletionVoltage.jl
lmh91: Ups, sorry. Remaining of Copy&Paste from testing script.

src/Simulation/Simulation.jl
lmh91: Ups, sorry. Remaining of Copy&Paste from testing script.

src/Simulation/Simulation.jl
```diff
@@ -953,7 +953,11 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
             # to switch it off (sor_const = 1)
             ref_limits = T.(_extend_refinement_limits(refinement_limits[iref]))
             if isEP
-                max_diffs = abs.(ref_limits .* bias_voltage)
+                max_diffs = if iszero(bias_voltage)
+                    abs.(ref_limits .* (extrema(sim.electric_potential.data) |> e -> e[2] - e[1]))
```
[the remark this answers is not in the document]
lmh91: It is called piping: https://docs.julialang.org/en/v1/manual/functions/#Function-composition-and-piping

lmh91:
I disabled the test for Cartesian case for `estimate_depletion_voltage` for now.
Will look into that later.

src/Simulation/Simulation.jl
```diff
@@ -845,6 +845,8 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
             sor_consts = (T(sor_consts), T(sor_consts))
         elseif length(sor_consts) > 1 && CS == Cartesian
             sor_consts = T(sor_consts[1])
+        else
+            sor_consts = T.(sor_consts)
```
fhagemann, line 849: Why is this needed?

test/runtests.jl
```diff
@@ -44,15 +44,14 @@ T = Float32
         nt = NamedTuple(sim)
         @test sim == Simulation(nt)
         id = 2
-        deplV = estimate_depletion_voltage(sim, id, verbose = false)
-        @info deplV
-        @test isapprox(deplV, T(1904), atol = 2.0) 
+        deplV = SolidStateDetectors.estimate_depletion_voltage(sim, id, (verbose = false,))
```
fhagemann, line 47: This function should be exported (but also doesn't harm to include this in the tests)

test/runtests.jl
```diff
@@ -44,15 +44,14 @@ T = Float32
...
+        @test isapprox(deplV, 1870*u"V", atol = 5.0*u"V") 
         # Check wether detector is undepleted, 10V below the previously calculated depletion voltage
-        sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = deplV - 40)
+        sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = ustrip(deplV - deplV*0.02))
         calculate_electric_potential!(sim, depletion_handling = true)
         @test !is_depleted(sim.point_types)
         # Check wether detector is depleted, 10V above the previously calculated depletion voltage
-        sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = deplV + 40)
+        sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = ustrip(deplV + deplV*0.02))
```
fhagemann, line 54: Is the error on this 2% or could we also use stricter limits to be more sensitive in the tests to possible future changes?

fhagemann (commented):
How long does it take to estimate the depletion voltage?
How does it compare memory-wise (if you copy a full simulation and calculate additional new potentials)? Might it be that this will produce out-of-memory errors for fine simulations?

PR #314, opened 2022-07-14
Enhance symmetry handling

SebastianRuffert:


examples/example_config_files/isochrone_test.yaml
```diff
@@ -46,6 +46,13 @@ detectors:
             h: 0.1
             translate:
               z: 0.05
+        symmetry:
+          phi:
+            left: reflecting
+            right: fixed
+          r:
+            left: fixed
+            right: fixed
```
lmh91, line 55: No, we want to switch to define the symmetries. Not the boundary conditions.

So something like: 
```yaml
symmetries:
     mirror_symmetry: 
           phi: 180
```
Then, the boundary conditions will be determined from the symmetries internally
for both kind of grids, cylindrical and Cartesian, automatically.

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,17 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
     return min(nextpow(2, max(cld(n+1, 25), 4)), max_nthreads)
 end
 
+function _get_grid_for_WP(sim::Simulation{T}, contact_id::Int) where T
```
lmh91, line 459: Why not call the function just `Grid`?

src/SolidStateDetector/Contacts.jl
```diff
@@ -46,8 +46,21 @@ struct Contact{T,G,MT} <: AbstractContact{T}
     id::Int
     name::String
     geometry::G
+    symmetry::Dict
```
lmh91, line 49: No. This will make the `struct Contact` type unstable. We don't need a `Dict` here.

We have to create structs for the symmetries. 

While I am writing this, I just realized that this actually does not belong to the contact itself. 
As the symmetries might depend on the environment of the detector. 
We need to add a vector of Symmetries (or a `Dict`) to the `Simulation` I think.

lmh91 (changes requested):
Thanks for starting this. 

I added some comments regarding the general design.

lmh91:
Alright, let's deal with the IO and how it is stored in `Simulation` later. 

Let's first get the following running:

```julia
using SolidStateDetectors
sim = Simulation(SSD_examples[:Coax]);

sym = MirrorSymmetry(30degree) # We need a constructor for this. 

Grid(sim)  # This should return a grid over the whole world. Thus, 360deg.
Grid(sim, symmetries = sym) # this should return a grid from 30deg to (30+180)deg with reflecting boundaries at both ends.
```

src/Simulation/Simulation.jl
```diff
@@ -458,6 +461,33 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
...
+        SSDInterval{T,SolidStateDetectors.get_boundary_types(sim.world.intervals[2])...}(sim.world.intervals[2].left, symmetry.symmetry_plane.origin[2])
+    z_interval = symmetry.symmetry_plane.normal[3] == 0 ? sim.world.intervals[3] : 
+        SSDInterval{T,SolidStateDetectors.get_boundary_types(sim.world.intervals[3])...}(sim.world.intervals[3].left, symmetry.symmetry_plane.origin[3])
+    world = World{world_types(sim.world)...}((x_interval, y_interval, z_interval))
+    Grid(sim, world = world)
+end
+
+function Grid(sim::Simulation{T, Cylindrical}, symmetry::MirrorSymmetry{T}) where {T}
+    r_interval = sim.world.intervals[1]
+    φ_interval = symmetry.symmetry_plane.normal[2] == 0 ? sim.world.intervals[2] :
```
SebastianRuffert, line 482: This is not yet mathematically correct

src/Symmetry/Mirror_Symmetry.jl
```diff
@@ -0,0 +1,33 @@
...
+
+function MirrorSymmetry{T}(origin::CartesianPoint{T}, normal::CartesianVector{T}) where {T}
+    MirrorSymmetry{T}(Plane{T}(origin, normal))
+end
+
+function MirrorSymmetry(axis, value::T) where {T<: SSDFloat}
+    println(axis)
+    if axis == "phi" || axis == "φ"
+        eps = 1e-5
+        MirrorSymmetry{T}(CylindricalPoint{T}(r=1, φ = value/360 * 2π), CylindricalVector{T}(0,eps,0))
```
SebastianRuffert, line 19: Input units not yet implemented

test/runtests.jl
```diff
@@ -258,6 +258,16 @@ end
...
 end
 
+@testset "Symmetry handling begin" begin
+    sim_no_symmetry = Simulation(SSD_examples[:InvertedCoax])
+    sim_with_symmetry = Simulation(SSD_examples[:InvertedCoax])
+    SolidStateDetectors.apply_initial_state!(sim_no_symmetry, WeightingPotential, 1) 
+    SolidStateDetectors.update_till_convergence!(sim_no_symmetry, WeightingPotential, 1, 1e-6)
+    SolidStateDetectors.apply_initial_state!(sim_with_symmetry, WeightingPotential, 1) 
+    SolidStateDetectors.update_till_convergence!(sim_with_symmetry, WeightingPotential, 1, 1e-6)
+    @test sim_no_symmetry.weighting_potentials[1][2,6,2] == sim_no_symmetry.weighting_potentials[1][2,6,2]
```
SebastianRuffert, line 268: Use several points, once this works

src/Simulation/Simulation.jl
```diff
@@ -228,9 +228,8 @@ function Grid(sim::Simulation{T, Cylindrical};
                 for_weighting_potential::Bool = false,
                 max_tick_distance::Union{Missing, LengthQuantity, Tuple{LengthQuantity, AngleQuantity, LengthQuantity}} = missing,
                 max_distance_ratio::Real = 5,
-                add_ticks_between_important_ticks::Bool = true)::CylindricalGrid{T} where {T}
+                add_ticks_between_important_ticks::Bool = true, world = sim.world )::CylindricalGrid{T} where {T}
```
lmh91, line 231: That is passed twice as `sim` is the first argument.

src/Simulation/Simulation.jl
```diff
@@ -356,9 +355,8 @@ function Grid(  sim::Simulation{T, Cartesian};
                 max_tick_distance::Union{Missing, LengthQuantity, Tuple{LengthQuantity, LengthQuantity, LengthQuantity}} = missing,
                 max_distance_ratio::Real = 5,
                 add_ticks_between_important_ticks::Bool = true,
-                for_weighting_potential::Bool = false)::CartesianGrid3D{T} where {T}
+                for_weighting_potential::Bool = false, world = sim.world )::CartesianGrid3D{T} where {T}
```
lmh91, line 358: Here again.

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,41 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
     return min(nextpow(2, max(cld(n+1, 25), 4)), max_nthreads)
 end
 
+function _get_grid_for_WP(sim::Simulation{T}, contact_id::Int) where T
```
lmh91, line 459: I guess we just wan't to call this function `Grid` and use dispatch.

`Grid(sim::Simulation)` for Electric Potential
`Grid(sim::Simulation, contact_id::Int)` for a Weighting Potential.

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,41 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
...
+        SSDInterval{T, :closed, :closed, :reflecting, :reflecting}(world.intervals[2].left,
+        CylindricalPoint(symmetry.symmetry_plane.origin)[2])
+    z_interval = symmetry.symmetry_plane.normal[3] == 0 ? world.intervals[3] : 
+        SSDInterval{T, SolidStateDetectors.get_boundary_types(world.intervals[3])[1], :closed,
+        SolidStateDetectors.get_boundary_types(world.intervals[3])[3], :reflecting}(world.intervals[3].left, 
+        symmetry.symmetry_plane.origin[3])
+    intervals = (r_interval, φ_interval, z_interval)
+end
+
+world_types(world::World{T,N,S}) where {T,N,S} = (T,N,S)
```
lmh91, line 493: This belongs in another file.

src/Symmetry/Mirror_Symmetry.jl
```diff
@@ -0,0 +1,35 @@
...
+    MirrorSymmetry{T}(Plane{T}(origin, normal))
+end
+
+function MirrorSymmetry(axis, value::T, units::NamedTuple = default_unit_tuple()) where {T<: SSDFloat}
+    length_unit = units.length
+    angle_unit = units.angle
+    if axis == "phi" || axis == "φ"
+        eps = 1e-7
+        MirrorSymmetry{T}(CylindricalPoint{T}(r=1, φ = to_internal_units(value * angle_unit)),
+            CylindricalVector{T}(0, to_internal_units(eps * length_unit), 0))
```
lmh91, line 21: Why this small value and not just `1`? 

Is this correct? A `Plane` is defined by a `CartesianPoint` and a `CartesianVector`.

If we find out that we get troubles with numerical conversions between Cartesian and cylindrical coordinates, we should define a dedicated struct like `PhiMirrorSymmetry`.

src/Simulation/Simulation.jl
SebastianRuffert: That is true, but this is the default only. When using a symmetry, another world will be passed, therefor I need this keyword

src/Simulation/Simulation.jl
SebastianRuffert: Sounds reasonable

src/Symmetry/Mirror_Symmetry.jl
SebastianRuffert: This was a convenience solution for the time being, because there is no function converting between cylindrical and cartesian vector. For the plane we also have the origin, meaning we can define the transformation. I used the analytical expression now. Still it contains a sin and cos, leading to numerical errors in the range of the machine precision. Are we okay with that or should we round?

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,57 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
...
+    z_interval = symmetry.symmetry_plane.normal[3] == 0 ? world.intervals[3] : 
+        SSDInterval{T, SolidStateDetectors.get_boundary_types(sim.world.intervals[3])[1], :closed,
+        SolidStateDetectors.get_boundary_types(world.intervals[3])[3], :reflecting}(world.intervals[3].left, 
+        symmetry.symmetry_plane.origin[3])
+    intervals = (x_interval, y_interval, z_interval)
+end
+
+function _get_grid_intervals(world::World{T, 3, Cylindrical}, symmetry::MirrorSymmetry{T}) where {T}
+    r_interval = world.intervals[1]
+    φ_interval = symmetry.symmetry_plane.normal[2] == 0 ? world.intervals[2] :
```
fhagemann, line 501: Same thing here: a `CylindricalVector` got converted to a `CartesianVector`.

I expect a bug here (`symmetry.symmetry_plane.normal[2] == 0`) even though I did not check it: 
you are checking if the `y` component of a `CartesianPoint` is 0. If we have a mirror symmetry at `φ = 30°`, the normal vector would point at `CartesianVector{T}(-0.5,0.866,0)`. Even though the second entry of `symmetry.symmetry_plane.normal` is not zero, we still want to adjust the limits and boundary types of the `φ_interval`.

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,57 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
...
+"""
+    Grid(sim::Simulation{T}, contact_id::Int) where T
+
+Determines the grid for a certain contact, constrained by its symmetry.
+    
+## Arguments
+* `sim::Simulation{T}`: [`Simulation`](@ref) for which the grid should be determined.
+* `contact_id::Int` : Contact for which the symmetry (specific for this contact) constrained grid should be determined
+"""
+function Grid(sim::Simulation{T}, contact_id::Int) where T
```
fhagemann, line 468: I am somewhat in favor of making `contact_id` a keyword argument...
From the outside, it is not easy to understand what `Grid(sim, 2)` might do..
Give you the second (?) `Grid` of the simulation?

fhagemann (changes requested):

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,57 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
...
+        SolidStateDetectors.get_boundary_types(world.intervals[3])[3], :reflecting}(world.intervals[3].left, 
+        symmetry.symmetry_plane.origin[3])
+    intervals = (x_interval, y_interval, z_interval)
+end
+
+function _get_grid_intervals(world::World{T, 3, Cylindrical}, symmetry::MirrorSymmetry{T}) where {T}
+    r_interval = world.intervals[1]
+    φ_interval = symmetry.symmetry_plane.normal[2] == 0 ? world.intervals[2] : 
+        SSDInterval{T, :closed, :closed, :reflecting, :reflecting}(world.intervals[2].left,
+        CylindricalPoint(symmetry.symmetry_plane.origin)[2])
```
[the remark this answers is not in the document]
SebastianRuffert: I get the concern. The alternative would be defining two separate structs for cartesian and cylindrical mirror symmetry

src/Simulation/Simulation.jl
SebastianRuffert: But then I can not use multiple dispatch anymore...

src/Simulation/Simulation.jl
```diff
@@ -458,6 +456,57 @@ function _guess_optimal_number_of_threads_for_SOR(gs::NTuple{3, Integer}, max_nt
...
+    y_interval = symmetry.symmetry_plane.normal[2] == 0 ? world.intervals[2] : 
+        SSDInterval{T, SolidStateDetectors.get_boundary_types(world.intervals[2])[1], :closed,
+        SolidStateDetectors.get_boundary_types(world.intervals[2])[3], :reflecting}(world.intervals[2].left, 
+        symmetry.symmetry_plane.origin[2])
+    z_interval = symmetry.symmetry_plane.normal[3] == 0 ? world.intervals[3] : 
+        SSDInterval{T, SolidStateDetectors.get_boundary_types(sim.world.intervals[3])[1], :closed,
+        SolidStateDetectors.get_boundary_types(world.intervals[3])[3], :reflecting}(world.intervals[3].left, 
+        symmetry.symmetry_plane.origin[3])
+    intervals = (x_interval, y_interval, z_interval)
+end
```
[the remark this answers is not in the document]
SebastianRuffert: Ok

PR #317, opened 2022-08-25
Fix bug when transforming `CartesianVector` to object coordinate systems

### LINKED ISSUE #316: Bug in normal(obj::(ConeMantle...)
The transformation of "obj_normal"  back to the global coordinate system should be a pure rotation and not a  full _transform_into_global_coordinate_system(...)  (i.e.  rotation&translation ) that can stretch and change the length of obj_normal.
Removing the spurious translation has fixed some charge-drift problems I had in simulating a coaxial detector with the passivation on a cylindrical surface.
I suppose that the same argument applies also to ElipsoidMantle and ThorusMantle, but, sorry, I have not checked it in details.

fhagemann:
Should fix #316.

As stated correctly by @bazzadino, `CartesianVectors` should not be translated when being transformed to the global or object coordinate system. I have introduced new methods for these functions for `CartesianVector` that do not translate but only rotate the vectors and have modified the source code accordingly.

### DIFF
--- a/src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
+++ b/src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
@@ -72,9 +72,11 @@ module ConstructiveSolidGeometry
     rotation(p::AbstractPrimitive) = p.rotation
     origin(p::AbstractPrimitive) = p.origin
     _transform_into_global_coordinate_system(pt::CartesianPoint, p::AbstractPrimitive) = (rotation(p) * pt) + origin(p)
+    _transform_into_global_coordinate_system(pt::CartesianVector, p::AbstractPrimitive) = rotation(p) * pt
     _transform_into_global_coordinate_system(pts::AbstractVector{<:CartesianPoint}, p::AbstractPrimitive) =
         broadcast(pt -> _transform_into_global_coordinate_system(pt, p), pts)
     _transform_into_object_coordinate_system(pt::CartesianPoint, p::AbstractPrimitive) = inv(rotation(p)) * (pt - origin(p)) 
+    _transform_into_object_coordinate_system(pt::CartesianVector, p::AbstractPrimitive) = inv(rotation(p)) * pt
     in(pt::CartesianPoint{T}, p::AbstractPrimitive{T}, csgtol::T = csg_default_tol(T)) where {T} = 
         _in(_transform_into_object_coordinate_system(pt, p), p; csgtol = csgtol)
     in(pt::CylindricalPoint{T}, p::AbstractPrimitive{T}, csgtol::T = csg_default_tol(T)) where {T} = 
--- a/src/ConstructiveSolidGeometry/LinePrimitives/Line.jl
+++ b/src/ConstructiveSolidGeometry/LinePrimitives/Line.jl
@@ -29,6 +29,6 @@ distance(pt::CartesianPoint, l::Line) = norm((pt - l.origin) × l.direction) / n
 
 function _transform_into_object_coordinate_system(l::Line{T}, p::AbstractPrimitive) where {T}
     origin = _transform_into_object_coordinate_system(l.origin, p) 
-    direction = inv(rotation(p)) * l.direction
-    Line( origin, CartesianVector(direction) )  
+    direction = _transform_into_object_coordinate_system(l.direction, p)
+    Line( origin, direction )  
 end
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
@@ -76,33 +76,33 @@ radius_at_z(cm::ConeMantle{T,Tuple{T,T}}, z::T) where {T} = radius_at_z(cm.hZ, c
 get_φ_limits(cm::ConeMantle{T,<:Any,T}) where {T} = T(0), cm.φ
 get_φ_limits(cm::ConeMantle{T,<:Any,Nothing}) where {T} = T(0), T(2π)
 
-function normal(cm::ConeMantle{T,T,<:Any,:inwards}, pt::CartesianPoint{T}) where {T}
+function normal(cm::ConeMantle{T,T,<:Any,:inwards}, pt::CartesianPoint{T})::CartesianVector{T} where {T}
     pto = _transform_into_object_coordinate_system(pt, cm)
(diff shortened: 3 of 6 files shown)


src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
```diff
@@ -72,9 +72,11 @@ module ConstructiveSolidGeometry
     rotation(p::AbstractPrimitive) = p.rotation
     origin(p::AbstractPrimitive) = p.origin
     _transform_into_global_coordinate_system(pt::CartesianPoint, p::AbstractPrimitive) = (rotation(p) * pt) + origin(p)
+    _transform_into_global_coordinate_system(pt::CartesianVector, p::AbstractPrimitive) = rotation(p) * pt
     _transform_into_global_coordinate_system(pts::AbstractVector{<:CartesianPoint}, p::AbstractPrimitive) =
         broadcast(pt -> _transform_into_global_coordinate_system(pt, p), pts)
     _transform_into_object_coordinate_system(pt::CartesianPoint, p::AbstractPrimitive) = inv(rotation(p)) * (pt - origin(p)) 
+    _transform_into_object_coordinate_system(pt::CartesianVector, p::AbstractPrimitive) = inv(rotation(p)) * pt
```
fhagemann, line 79: These are the new methods for transforming `CartesianVectors` (just without the translation)

src/ConstructiveSolidGeometry/LinePrimitives/Line.jl
```diff
@@ -29,6 +29,6 @@ distance(pt::CartesianPoint, l::Line) = norm((pt - l.origin) × l.direction) / n
 
 function _transform_into_object_coordinate_system(l::Line{T}, p::AbstractPrimitive) where {T}
     origin = _transform_into_object_coordinate_system(l.origin, p) 
-    direction = inv(rotation(p)) * l.direction
-    Line( origin, CartesianVector(direction) )  
+    direction = _transform_into_object_coordinate_system(l.direction, p)
+    Line( origin, direction )
```
fhagemann, line 33: This also allows us to replace some pieces of the old code with this new methods
(here: `l.direction` is already a `CartesianVector`)

src/ConstructiveSolidGeometry/SurfacePrimitives/EllipticalSurface.jl
```diff
@@ -69,9 +69,10 @@ const PartialCircularArea{T} = EllipticalSurface{T,T,T}
 const Annulus{T} = EllipticalSurface{T,Tuple{T,T},Nothing}
 const PartialAnnulus{T} = EllipticalSurface{T,Tuple{T,T},T}
 
-Plane(es::EllipticalSurface{T}) where {T} = Plane{T}(es.origin, es.rotation * CartesianVector{T}(zero(T),zero(T),one(T)))
+Plane(es::EllipticalSurface{T}) where {T} = Plane{T}(es.origin, normal(es))
 
-normal(es::EllipticalSurface{T}, ::CartesianPoint{T} = zero(CartesianPoint{T})) where {T} = es.rotation * CartesianVector{T}(zero(T), zero(T), one(T))
+normal(es::EllipticalSurface{T}, ::CartesianPoint{T} = zero(CartesianPoint{T})) where {T} = 
+    _transform_into_global_coordinate_system(CartesianVector{T}(zero(T), zero(T), one(T)), es)
```
fhagemann, line 75: Or here: we rotate `[0,0,1]` into the global coordinate system using the `rotation` of the primitive, in this case the `EllipticalSurface`

src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
```diff
@@ -76,33 +76,33 @@ radius_at_z(cm::ConeMantle{T,Tuple{T,T}}, z::T) where {T} = radius_at_z(cm.hZ, c
...
 get_φ_limits(cm::ConeMantle{T,<:Any,Nothing}) where {T} = T(0), T(2π)
 
-function normal(cm::ConeMantle{T,T,<:Any,:inwards}, pt::CartesianPoint{T}) where {T}
+function normal(cm::ConeMantle{T,T,<:Any,:inwards}, pt::CartesianPoint{T})::CartesianVector{T} where {T}
     pto = _transform_into_object_coordinate_system(pt, cm)
     cyl = CylindricalPoint(pto)
-    return CartesianVector(_transform_into_global_coordinate_system(
-            CartesianPoint(CylindricalPoint{T}(-cyl.r, cyl.φ, zero(T))), cm))
+    return _transform_into_global_coordinate_system(
+            CartesianVector(CartesianPoint(CylindricalPoint{T}(-cyl.r, cyl.φ, zero(T)))), cm)
```
fhagemann, line 83: See this example: In order for this to work, the `CartesianPoint` is **first** converted to a `CartesianVector` and **then** transformed.

bazzadino: > See this example: In order for this to work, the `CartesianPoint` is **first** converted to a `CartesianVector` and **then** transformed.

Maybe a silly question due to the fact that I am not very familiar with Julia but, wouldn't it be better to be explicit and define a new function _rotate_into_object_coordinate_system instead of adding a new method to _transform_into_object_coordinate_system where one depends on the order of parameters?

fhagemann: We are using multiple dispatch here, meaning that the method for the function will be chosen by the **type** of the parameters, not their **order** (one is `_transform...(::CartesianPoint, ::AbstractPrimitive)`, the other is `_transform...(::CartesianVector, ::AbstractPrimitive)`, so the primitive is always second).

In that case, the programmer can use the same function name (and not care about if the first argument is a `Point` or a `Vector`) but it does different things depending on the type of the argument passed to the function.

bazzadino: Thank you Felix,
it was indeed a silly question.
To attenuate (a bit) its silliness, I was referring to your starting replay
-    _transform_into_global_coordinate_system(es, CartesianVector{T}(zero(T), zero(T), one(T)))
+   _transform_into_global_coordinate_system(CartesianVector{T}(zero(T), zero(T), one(T)), es)
where the change of  order  implies the change of types that allows for multiple dispatch...
I'll stop interfering!
Dino

PR #318, opened 2022-09-02
Fix internal copy of `sim` in `estimate_depletion_voltage`

lmh91:


fhagemann:
What's the problem here?

lmh91:
If a Simulation was not created by a config file, but manually in a script file, there is no `sim.config_dict`.
Thus, the creation of the temporary Simulation object fails.

PR #327, opened 2023-01-18
Fix unused declared type variables in method definitions

fhagemann:
When precompiling, SSD throws warning that some type variables in method definition are declared but not used. I have looked through the list of warnings and updated the method definitions accordingly by either removing the unused type variables or using them.


oschulz:
Yes, getting rid of unused type variables is important, they can cause type inference failure, affecting performance.

PR #328, opened 2023-02-14
Fix conversion Geometry <-> Dictionary

fhagemann:


src/ConstructiveSolidGeometry/IO.jl
```diff
@@ -231,7 +232,7 @@ function get_origin(::Type{T}, dict::AbstractDict, unit::Unitful.Units) where {T
     haskeyssum = sum(haskeys)
     if haskeyssum == 1
         o = dict[origin_names[i]]
-        if o isa Vector
+        if o isa AbstractVector
```
fhagemann, line 235: When a `CSG.Geometry` is converted to a `Dict` using the method `CSG.Dictionary`, then the `origin` is saved as `CartesianPoint`. If one tries to create a `CSG.geometry` from that dictionary it ignores the `origin`, as it is not a `Vector` but `<: AbstractVector`. The method to parse the `origin` vector from a `Dict` was generalized to avoid this issue.

test/ConstructiveSolidGeometry/CSG_primitives.jl
```diff
@@ -299,4 +299,10 @@ no_translations = (rotation = one(SMatrix{3, 3, T, 9}), translation = zero(Carte
         @test CartesianPoint(cyl) == cart
         @test cyl2.φ == T(π)
     end
+end
+
+@testset "Test geometry parsing" begin
+    sim = Simulation{T}(SSD_examples[:InvertedCoax])
+    c = sim.detector.contacts[2].geometry
+    @test Geometry(T, Dictionary(c), default_units, no_translations) == c
 end
```
fhagemann, line 308: This test was added to check the conversion `Geometry` <-> `Dict`

PR #331, opened 2023-03-18
Update `_in` method `FullTorus` and `HollowTorus`

### LINKED ISSUE #330: Negative torus does not contain center
Negative volumes do not include their surfaces as intended
However in the case of torus the "inner surface" at ``r_tube = 0`` get omitted because of this. 
 See example tube with torus removed bellow

Screen Shot 2023-03-17 at 10 14 38 PM

```
"torus": {
    "r_torus": 5.0,
    "r_tube": {
        "from": 0.0,
        "to": 2.0
    },
    "theta": {
        "from": "0.0°",
        "to": "360.0°"
    },
    "origin": [
        0.0,
        0.0,
        70.0
    ]
}
```

fhagemann:
This PR should fix the issue reported in #330, where the a `FullTorus` would not include the center point of the tube if it was an`OpenPrimitive`.

I locally modified the geometry of the semiconductor in the [example config file of our toroidal detector](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/main/examples/example_config_files/public_coaxial_torus_config.yaml#L41-L46) to have a negative `Torus` starting at `0`:
```
    geometry:
      difference:
      - torus:
          r_torus: 10
          r_tube:
            from: 0
            to: 5
      - torus:
          r_torus: 10
          r_tube:
            from: 0
            to: 2
```

Even though SSD understands that this describes a `FullTorus` because of `from: 0` (see [here](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/47d2eefa3735b564122291a0216c84cdb30e50ef/src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl#L149)), the [`_in` method for `Torus`](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/47d2eefa3735b564122291a0216c84cdb30e50ef/src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl#L302-L315) would determine the radial endpoints check whether the point is in the interval, even if the lower endpoint was `0`.

I implemented different `_in` methods for `FullTorus` and `HollowTorus` to fix #330.

## Before this PR
InnerTorusOld

## After this PR
InnerTorusNew

### DIFF
--- a/src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
+++ b/src/ConstructiveSolidGeometry/VolumePrimitives/Torus.jl
@@ -12,6 +12,7 @@ radius around a given origin.
     or `OpenPrimitive`, i.e. the surface points do not belong to the primitive.
 * `TR`: Type of `r_tube`.
     * `TR == T`: Full tube without cutout (constant radius `r_tube`).
+    * `TR == Tuple{T,T}`: Hollow tube with cutout (inner radius `r_tube[1]`, outer radius `r_tube[2]`).
 * `TP`: Type of the azimuthial angle `φ`.
     * `TP == Nothing`: Full 2π in `φ`.
     * `TP == T`: Partial Torus ranging from `0` to `φ`.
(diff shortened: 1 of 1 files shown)


hervasa2 (approved):
Looks great Felix! Thank you for taking a look at this so quickly!

fhagemann:
Well, thanks for reporting! :)

PR #333, opened 2023-03-25
Update to KernelAbstractions v0.9

oschulz:
Keeps support for KernelAbstractions to v0.8.

Closes #329


Project.toml
```diff
@@ -48,7 +48,7 @@ GPUArrays = "8"
 Interpolations = "0.14"
 IntervalSets = "0.3, 0.4, 0.5, 0.6, 0.7"
 JSON = "0.21"
-KernelAbstractions = "0.8"
+KernelAbstractions = "0.8, 0.9"
```
fhagemann, line 51: Does this still work with 0.8?

oschulz:
Seems to run on CUDA with both KernelAbstractions v0.8 and v0.9.

src/ka_compat.jl
fhagemann, line 1: I see, this is where the support for both v0.8 and v0.9 comes from!

test/benchmarks/benchmark_SOR_update_method.jl
```diff
@@ -63,9 +63,9 @@ df_SOR_update = if !isfile(fn_SOR_update_benchmark_df) || recalculate
                 sim.detector, sim.electric_potential.grid, sim.medium, sim.electric_potential.data, sim.imp_scale.data, 
             ));
             ndrange = size(pcs.potential)[1:3] .- 2
-            dev = get_device(pcs.potential)
```
fhagemann, line 66: Here, we should also use `_get_kabackend` rather than `get_device`
Add the beginning of the file, it also says: `using KernelAbstractions: get_device`, I guess this will not work anymore..

Not sure if this benchmarking scripts are used for the actual tests or to be run manually if wanted though.

fhagemann (changes requested):

test/benchmarks/benchmark_SOR_update_method.jl
oschulz: Ooops, overlooked that one, changed now.

fhagemann:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/9e25ae6ce51f13fcf719fce1f55db3549f0b0b36/test/benchmarks/benchmark_SOR_update_method.jl#L14

Will this give an error?

oschulz:
>  using KernelAbstractions: get_device 

Slipped through the search somehow, gone now.

oschulz:
We should make our tests leaner/faster in the future - CI is taking a long time ... :-)

fhagemann (approved):

oschulz:
@fhagemann maybe tag a new release after CI on the current main branch has passed?

fhagemann:
Here we go:
https://github.com/JuliaRegistries/General/pull/80327

PR #335, opened 2023-04-17
Update the diffusion implementation

johannaluehrs:
The old diffusion implementation in SSD is not optimal and needed an update.

In this PR, the diffusion implementation in SSD was improved:
- The previous diffusion implementation favored diffusion in `z` direction and was anisotropic. The diffusion is isotropic now (see 634b839)
- SSD will now take the value of diffusion constants for electrons and holes as input (see 0cb43dd) and simulate diffusion such that it reproduces that diffusion constant (see 997c5b6)
-  Instead of using effective electric field vectors, diffusion is added as a fixed-step vector after simulating the drift step (resulting from external electric fields and self-repulsion), (see bc40119)
- Update the documentation to describe the new diffusion implementation (see 7a7cfe4)


docs/src/man/charge_drift.md
```diff
@@ -224,24 +224,24 @@ simulate!(evt, sim, diffusion = true)
 ```
 ![Diffusion](../assets/diffusion.gif)
 
-The magnitude of the diffusion field vector influences the diffusion constant, $D$, with which the charge carriers diffuse.
-In the abscence of an external electric field, the size (standard deviation $\sigma$) of the charge cloud is expected to evolve with $\sigma = \sqrt{2 D t}$, where $t$ is the time. Thus, the diffusion constant can be estimated from the time evolution of the charge cloud size.
+The diffusion constant, $D$, influences the magnitude of the diffusion field vector.
```
fhagemann, line 227: Shouldn't it be `diffusion step vector` instead of `diffusion field vector`?

fhagemann (approved):
Looks good to me!

PR #339, opened 2023-05-12
Allow Julia v1.6

oschulz:


fhagemann:
Did you test if this works or are we just seeing whether the tests pass/fail ?

fhagemann:
The number of threads with which the tests run seems to be 1. Is there any way to run the tests with more than 1 thread to allow for a massive speed-up?

oschulz:
We can try - apparently the [standard GitHub runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) has two CPUs. There is a [beta for larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners), but I don't know if that costs something.

fhagemann (approved):

fhagemann:
Ill try something out, but this PR should be good for now

oschulz:
Thanks!

oschulz:
> Is there any way to run the tests with more than 1 thread to allow for a massive speed-up?

Trying this out in #340

PR #340, opened 2023-05-21
Run CI with four Julia threads active

oschulz:


fhagemann:
What happens if we set this to 8 threads? ^^
And any idea why we are getting an OutOfMemoryError on x86?

oschulz:
> What happens if we set this to 8 threads?

Well, the default GitHub Actions runner only has two virtual CPUs, so that should work but provide no speed-up.

> And any idea why we are getting an OutOfMemoryError on x86

Nope ...

oschulz:
I'll re-run all CI jobs to time them again.

fhagemann:
Am I seeing it right that the ubuntu x64 tests on Julia 1 take almost 4 hours (?)

oschulz:
> Am I seeing it right that the ubuntu x64 tests on Julia 1 take almost 4 hours (?)

Yes, and it seems to be reproducible, and I don't understand why. Before (this PR) it was about 1.5h. But the tests on the other platforms, that also used to take around 1.5 h, do run in about 50 minutes now, which makes sense.

oschulz:
I've added test timers to this PR (should be useful in general), let's see where all the time is spent.

oschulz:
Tests on Julia v1.9 keep failing with out-of-memory, not sure why (and why only when using multithreading).

Docs failures should be fixed by #341.

fhagemann:
ubuntu x64 is still significantly slower than the rest..

oschulz:
> ubuntu x64 is still significantly slower than the rest..

Yes, massively - but I haven't the faintest intuition on why, it looks really strange.

oschulz:
How long does it run locally, on two core, x86-64, in comparison? Maybe the GitHub x86-64 runners are just in higher demand and more heavily oversubscribed?

fhagemann:
Let's see whether something has changed here ever since..

fhagemann:
@oschulz looks like the tests all speed up except for the one also testing the code coverage. Any idea why?

fhagemann:
I rebased this PR onto the current `main`

fhagemann:
The test also testing the code coverage seemed to take significantly longer than the rest.
Let's see whether using #434 solves this.

fhagemann:
Looks like the number of processors for public repositories was updated to four:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

fhagemann (approved):
I think this is the best we can do for now:
run all tests on julia-1.11 with 4 threads, and test the coverage on the (faster) julia-1.10 tests with 1 thread.

fhagemann:
The last force-push was a rebase onto the current `main`.

PR #341, opened 2023-05-28
Workaround for  JuliaPlots/Plots.jl#4758 and test timing

oschulz:
Workaround for JuliaPlots/Plots.jl#4758


oschulz:
I've added the test-timing code from #340, should be very useful in general.

PR #342, opened 2023-06-12
Fix plotting of electric field lines

fhagemann:
When plotting electric field lines, charges are drifting according to the electric field. The initial spawn positions are determined by sampling the contacts in steps of `sampling` and moving them into the bulk of the detector by an `offset`.

In `ConstructiveSolidGeometry`, the surfaces of a `VolumePrimitive` are defined such that their `normal` vector always points towards the center of the `VolumePrimitive`. In order to compute the correct spawn positions, the charges need to be spawned outside of the contact, i.e. in negative direction of the `normal` vector.

This behavior does not appear if the contacts have a thickness of 0. It does, however, when the contacts have non-zero width.

### DIFF
--- a/src/PlotRecipes/ElectricField.jl
+++ b/src/PlotRecipes/ElectricField.jl
@@ -149,7 +149,7 @@ end
                 pts = ConstructiveSolidGeometry.intersection(surf,l)
                 for pt in pts
                     if pt in c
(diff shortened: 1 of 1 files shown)


oschulz:
Please merge and release at your convenience @fhagemann .

PR #347, opened 2023-06-16
Improve handling of material properties

fhagemann:
Follow-up of #346:
Added a function `get_material_properties` to handle the breaking changes in defining copper as `Cu` instead of `Co`.
Added a test to check that the `ConfigFileError` is thrown when `Co` is still used.

### DIFF
--- a/src/MaterialProperties/MaterialProperties.jl
+++ b/src/MaterialProperties/MaterialProperties.jl
@@ -97,11 +97,7 @@ materials = Dict{String, Symbol}(
     "Al"  => :Al,
     "LAr" => :LAr,
     "CZT" => :CdZnTe,
-    "Si" => :Si,
-    "Co" => begin
-        @warn "In v0.9.0, the material 'Co' does not denote copper anymore. Please use 'Cu' for copper."
-        :Co
-    end
+    "Si" => :Si
 )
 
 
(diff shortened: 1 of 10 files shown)


src/MaterialProperties/MaterialProperties.jl
```diff
@@ -110,3 +106,10 @@ for key in keys(material_properties)
         push!(materials, string(key) => key)
     end
 end
+
+function get_material_properties(s::String, material_properties = material_properties, materials = materials)::NamedTuple
+    !haskey(materials, "Co") && s == "Co" && # users might have added 'Co' themselves
+        throw(ConfigFileError("In v0.9.0, the material 'Co' does not denote copper anymore.\n"*
+            "Please use 'Cu', 'Copper' or 'copper' to define copper in the configuration file."))
+    material_properties[materials[s]]
+end
```
fhagemann, line 115: @oschulz would this be the proper way of defining the function?
Not sure how to deal with the dictionaries `material_properties` and `materials` and if/how to pass them to the functions.

oschulz: As for `"Please use 'Cu', 'Copper' or 'copper'  ..."` - if we're making breaking changes, should we get stricter in general and always require the chemical symbol?

oschulz: I'm not sure about the order of the arguments ...

fhagemann: For germanium, we use `HPGe` right now and we also have composite materials and vacuum that don't allow for short chemical symbols.

The main reason to disallow `Co` for copper was that `Co` should be reserved for cobalt, if ever introduced.

oschulz: Good point, we should certainly allow for common names for composites. We may have to think about this some more in regard to Geant4 integration, people may need more complex and custom composites for holder structures and so on.

PR #348, opened 2023-07-05
Support setting the units on the axes when plotting 

fhagemann:
```julia
using SolidStateDetectors, Unitful, Plots
```

Somehow, the current implementation (stable v0.9.0) does not allow to change the units of the axes easily, meaning that the following snippets of code kill the kernel:
```julia
plot(sim.electric_potential, xunit = u"mm", yunit = u"mm")
```
```julia
plot(u"mm", u"mm", u"V")
plot!(sim.electric_potential)
```

This should be fixed now by explicitly setting the default units to the internal units and by giving `xlims` and `ylims` units.

```julia
plot(sim.electric_potential, xunit = u"mm", yunit = u"mm")
```
works like a charm:
grafik

```julia
plot(u"mm", u"mm", u"V")
plot!(sim.electric_potential)
```
looks weird somehow (but people should then just not use it for now):
grafik

I updated the electric field plot in the tutorial, so that it's to be seen somewhere.

Other plots:
```julia
plot(sim.point_types, xunit = u"cm", yunit = u"cm", full_det = true)
```
grafik

```julia
plot(sim.q_eff_imp, xunit = u"cm", yunit = u"cm", zunit = u"MV * m", full_det = true)
```
grafik


fhagemann:
This should be tested with all possible things there are to plot and with both coordinate systems, to be sure that nothing was messed up here.

PR #358, opened 2023-09-14
Update MaterialProperties.jl

lmanzanillas:
When using diffusion with Si material it gives error because De and Dh are missing in the material properties file.  Taking values from; 
https://sites.ecse.rpi.edu/~schubert/Educational-resources/Materials-Semiconductors-Si-and-Ge.pdf


fhagemann (approved):
Looks good to me, thanks!

PR #361, opened 2023-10-10
SquareRootModel: an  electric field dependent temperature model

hervasa2:
Added the temperature model implemented in _siggen_.

The parameterization for the longitudinal drift velocity temperature dependence, ``v_l``, as a function of the electric 
field strength, ``E``, was proposed by [M.Ali Omar, Lino Reggiani](https://www.sciencedirect.com/science/article/pii/0038110187900633):
```math
v_l(T) = v_l(T_0)f(T)/f(T_0) 
```
where v_l(T_0) is a known velocity at reference temperature ``T_0``. The temperature dependence, ``f(T)``,  is parametrized as:
```math
f(T) =  V_s(T)\frac{E/E_c(T)}{(1 + (E/E_c(T))^2)^{1/2}}
```
where 
```math
E_c(T) = V_s(T)/\mu_0(T)~, \quad V_s(T) = B\tanh^{1/2}(\theta/2T)~, \quad\mu_0(T) = A/T^P
```
The four parameters, ``A`` (``\mu_0`` at ``T=1``K), ``P``, ``B``, ``\theta``, are different
for electrons and holes. The work was confined to the fields along the <100> axis. Here the same values are assumed for the <111> axis. 

`println(sim.detector.semiconductor.charge_drift_model.temperaturemodel)`
```
________SquareRootModel________
V_s(T) = Btanh^{1/2}(theta/2T) and mu_0(T) = A/T^P

---Temperature settings---
Crystal temperature:  	95.0
Reference temperature:	78.0

---Fitting parameters---
   	e100      	h100
A 	5660.0   	163000.0
P 	-1.68   	-2.398
B 	130000.0   	120000.0
theta 	200.0   	200.0
```
`SSD.scale_to_given_temperature(T(1000), sim.detector.semiconductor.charge_drift_model.temperaturemodel)`
`(0.71818286f0, 0.71818286f0, 0.62357354f0, 0.62357354f0)`
drift_vel_temp_corr


fhagemann (commented):
Thanks for the initiative. 
I think it doesn't harm to implement the exact same temperature model as in _siggen_ to be able to compare results determined from both simulation packages.

One thought: the model you are quoting seems very similar to the parameterization of the longitudinal drift velocity used in the `ADLChargeDriftModel`. Can we maybe find a way to scale the parameters of `ADLChargeDriftModel` rather than the final drift velocity?
This might allow to speed up the code as we could avoid calling `scale_to_given_temperature` in every step of the charge drift simulation.

oschulz:
@hervasa2 is this still active?

hervasa2:
Lets keep it open since its something we definitely need it. I think we should proceed with @fhagemann suggestion. Will get to it soon.

hervasa2:
I'm not sure you understood what I am suggesting. It's exactly something like your `scale_to_temperature(cdm::ADL2016ChargeDriftModel{T,M,N}, Temp...}`. However, instead of using the parameters from the Omar and Reggiani paper directly in this function, I scale the velocities with the Omar and Reggiani parameters and fit them with mu, beta, and E0 completely free. Thus I predetermine mu(T), beta(T), and E0(T) numerically. This gives the best possible fit without adding additional parameters to the ADL2016ChargeDriftModel. Thus temperature scaling happens on the parameters of the model itself, not on EVERY step of charge drift. Note that in my fit at 100 K for holes on <100> I don't use beta = 2, but the best fit parameters beta = 1.05....

hervasa2:
Ideally we would not use the square root model at all. We would just fit ADL2016ChargeDriftModel to the data Omar and Reggiani used. But we don't have that. My intent here is to recreate their data with their parameters and provide the best possible fit, leaving all parameters free. I see no reason why to fix beta at its value at 77K.

hervasa2:
I will go ahead and create a new PR. We can discuss further there.

hervasa2:
For work in progress please see https://github.com/hervasa2/SolidStateDetectors.jl/commit/3a3f5d4b0e26afa4fe1956d764e582cf9d8c9073

fhagemann:
> Also note that my intent is to reproduce this original PR without using `scale_to_given_temperature`. This is for the sake of comparison to siggen. However if we are already deviating from the exact scaling used in siggen, we might as well just apply what is best: (as in your thesis) leaving beta fixed and using the mu and E0 scaling suggested by Omar and Reggiani (since this reproduces your data well).

Would it be bit work implementing both (beta fixed and beta T-dependent) and letting the user choose.
I'm just wondering: in case somewhere were to use the `ADL2016ChargeDriftModel` with other values for mu0, E0, and beta -- would they have to rerun your analysis to get the correct scaling according to the square-root-model, or would SSD do that for them?

fhagemann:
>  Screenshot 2025-11-12 at 18 47 46 

For this plot: you said you are using a cubic function to fit this.
Is this functions also well described using a power law: 
$$\dfrac{\mu_0(T)}{\mu{T_0}} = \left(\dfrac{T_0}{T}\right)^p$$ 
and, if so, is the parameter $p$ close to the value reported in the Omar/Reggiani paper?

hervasa2:
> For this plot: you said you are using a cubic function to fit this.
Is this functions also well described using a power law:

Here is the power law fit. The fit value doesn't match the P = 2.4 for holes. I am inclined to abandon the direct reproduction of Omar and Reggiani temperature scaling used in siggen (since we are getting rid of `scale_to_given_temperature`).

Screenshot 2025-11-13 at 18 30 23

PR #369, opened 2024-01-28
Add constructor for `ADLChargeDriftModel`

### LINKED ISSUE #252: Add constructor method for `ADLChargeDriftModel` (and others)
Right now the ADL charge drift model is usually constructed from a configuration file. 

I think it would be nice to also have a constructor function for it,
which takes all the parameters, `e100μ0` ..., as keyword arguments 
(defaults would be the values from the ADL paper(s) which we can update once we have better ones). 

The current method taking the configuration-file (dictionary) as input should than just pass the values to that keyword constructor method.

Same for the temperature model constructors.

fhagemann:
This PR improves the convenience of the `ADLChargeDriftModel` constructor and closes #252.

Previously, the `ADLChargeDriftModel` was always constructed from a config file. Now:
- it can also be constructed from a dictionary in a Julia session
- parameters in the config file or the config dict can have units
- the syntax of the config file/dict is checked before parsing it
- single parameter values can be overwritten using keyword arguments (with/without units), e.g.
```julia
ADLChargeDriftModel(e100μ0 = 40000u"cm^2/(V*s)", h100β = 2, h111E0 = 50u"V/cm")
```


src/ChargeDriftModels/ADLChargeDriftModel/ADLChargeDriftModel.jl
```diff
@@ -159,39 +159,88 @@ function ADLChargeDriftModel{T,M,N,TM}(chargedriftmodel::ADLChargeDriftModel{<:A
...
-    e111 = VelocityParameters{T}(e111mu0, e111beta, e111E0, e111mun)
-    h100 = VelocityParameters{T}(h100mu0, h100beta, h100E0, 0)
-    h111 = VelocityParameters{T}(h111mu0, h111beta, h111E0, 0)
+
+ADLChargeDriftModel{T}(args...; kwargs...) where {T <: SSDFloat} = ADLChargeDriftModel(args...; T=T, kwargs...)
+
+const defaultADLfile = joinpath(get_path_to_example_config_files(), "ADLChargeDriftModel/drift_velocity_config.yaml")
+function ADLChargeDriftModel(configfilename::AbstractString = defaultADLfile; kwargs...)
+    ADLChargeDriftModel(parse_config_file(configfilename); kwargs...)
+end
```
fhagemann, line 168: If no config file is provided, the default ADLChargeDriftModel parameters are loaded as basis.

src/ConstructiveSolidGeometry/IO.jl
```diff
@@ -58,8 +58,11 @@ end
 
 # parses dictionary entries of type Real or String to their value in internal units
 @inline _parse_value(::Type{T}, x::Real, unit::Unitful.Units) where {T} = T(to_internal_units(float(x) * unit))
-@inline _parse_value(::Type{T}, x::Quantity, unit::Unitful.Units) where {T} = T(ustrip(unit, x))
-@inline _parse_value(::Type{T}, s::String, ::Unitful.Units) where {T} = T(to_internal_units(uparse(s)))
+@inline _parse_value(::Type{T}, x::Quantity, unit::Unitful.Units) where {T} = begin
+    if dimension(x) != dimension(unit) throw(ConfigFileError("The quantity '$(x)' cannot be converted to unit '$(unit)'")) end
+    T(to_internal_units(float(x)))
+end
+@inline _parse_value(::Type{T}, s::String, unit::Unitful.Units) where {T} = _parse_value(T, uparse(s), unit)
```
fhagemann, line 65: `_parse_value` now also checks whether the passed value has the correct dimensions and throws a `ConfigFileError` if it doesn't

test/ADLChargeDriftModel.jl
```diff
@@ -124,3 +126,76 @@ for T in (Float32, Float64)
...
+
+end
+
+@testset "Modify mobility parameters using keyword arguments" begin
+    cdm = ADLChargeDriftModel{Float64}(e100μ0 = 40000u"cm^2/(V*s)", e111μn = 5u"cm^2/(V*s)", h100β = 2, h111E0 = 50u"V/cm")
+    @test cdm.electrons.axis100.mu0 == 4.0     # internal unit is m^2/(V*s)
+    @test cdm.electrons.axis111.mun == 0.0005  # internal unit is m^2/(V*s)
+    @test cdm.holes.axis100.beta == 2.0        # no unit conversion
+    @test cdm.holes.axis111.E0 == 5000.0       # internal unit is V/m
+end
```
fhagemann, line 159: Internally, the parameter values are stored in `internal_units`

PR #370, opened 2024-02-01
Update to Documenter v1

oschulz:


docs/make.jl
```diff
@@ -54,7 +54,7 @@ makedocs(
     doctest = ("fixdoctests" in ARGS) ? :fix : true,
     format = Documenter.HTML(canonical = "https://JuliaPhysics.github.io/SolidStateDetectors.jl/stable/", prettyurls = !("local" in ARGS)),
     linkcheck = ("linkcheck" in ARGS),
-    strict = !("local" in ARGS),
+    warnonly = ("nonstrict" in ARGS),
```
fhagemann, line 57: Why would you rename this...

docs/make.jl
```diff
@@ -54,7 +54,7 @@ makedocs(
     doctest = ("fixdoctests" in ARGS) ? :fix : true,
     format = Documenter.HTML(canonical = "https://JuliaPhysics.github.io/SolidStateDetectors.jl/stable/", prettyurls = !("local" in ARGS)),
```
fhagemann, line 55: ... but not this?

We use the `"local"` keyword whenever building docs on our local computers.
Not sure what you intend with `"nonstrict"` instead of `"local"` here?

oschulz: I want to keep this consistent between our packages, I made the switch in the other packages as well. I think `nonstrict` is semantically clearer than `local`.

oschulz: The `local` in regard to `prettyurls` actually serves a different function, it's to build static HTML docs that don't need a local HTTP server to function properly (though now that VS-Code has such functionality out-of-the-box, we probably will use `local` less often).

oschulz:
@fhagemann there are a bunch of errors "invalid local link/image: path pointing to a file outside of build directory in src/api.md" now with Documenter v1, not sure if Documenter has just become more picky with checks, of if this a new requirement in Documenter v1.

fhagemann:
The problem is that the docstrings include links to images (which anyhow did not seem to work previously).
I removed the lines from the respective docstrings, let's see whether this fixes the failing tests.

fhagemann:
I‘m a bit concerned about the fact that the docs now take longer than the tests (.) @oschulz 

it used to be something like 6min in the past, now 40 min

oschulz:
> I‘m a bit concerned about the fact that the docs now take longer than the tests

Yes, that's no good. I just don't know why ...

fhagemann:
Building the documentation in `julia-1.10` for now, because it takes somewhat longer on `julia-1.11`

PR #372, opened 2024-02-13
Add Downgrade CI

fhagemann:
This was a bit of PITA and took quite some time, but it's for a good cause! 

[Downgrade CI](https://discourse.julialang.org/t/psa-add-downgrade-ci-to-better-check-version-compatability/110063) can be used to check whether the versions bounds in `compat` are actually supported.

Why wouldn't that be? If a package introduces some functionality in its newest version and it is added to the project, the `compat` lower bound is not automatically updated because it might not be tested. Downgrade CI tests the package with the lowest allowed `compat` bounds and, then, runs the test. This way, `compat` lower bounds will also be tested and can be updated if necessary.

In SSD, three packages were listed with `compat` bounds too low:
- `ProgressMeter`: The `ProgressMeter.ProgressThresh` in [convergence.jl#L22](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/1677a035da0bd1dd07e0979eee79deff70432c95/src/PotentialCalculation/SuccessiveOverRelaxation/convergence.jl#L22) uses keyword arguments that were introduced in `ProgressMeter@1.5`. In the Project.toml, versions down to `ProgressMeter@1.2` were allowed.
- `Unitful`: The extensive use of `uparse` in `ConstructiveSolidGeometry` was only introduced in `Unitful@1.0` and the support for `range` with units in [MCEventsProcessing.jl#L149](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/1677a035da0bd1dd07e0979eee79deff70432c95/src/MCEventsProcessing/MCEventsProcessing.jl#L149) for julia-1.8 (or newer) requires at least `Unitful@1.11`. In the Project.toml, the minimum version is still `Unitful@0.18`.
- `IntervalSets`: The function `isclosedset` used in [ScalarPotential.jl#L124](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/1677a035da0bd1dd07e0979eee79deff70432c95/src/ScalarPotentials/ScalarPotential.jl#L124) was introduced in `IntervalSets@0.5`. The Project.toml allowed versions down to `IntervalSets@0.3`.

When running Downgrade, a series of other incompatibilities appeared:
- Some bug in `StaticArrays@0.12` 
--> set lower bound to `StaticArrays@1` 
--> requires lower bounds on `Rotations@1` and `Clustering@0.15`
- `Interpolations@0.14`
--> requires lower bound on `Requires@1.1`
--> requires lower bound on `ArraysOfArrays@0.5`
- `FillArrays@0.9`
--> requires lower bound on `Distributions@0.24.5`
- `LinearAlgebra`, `Random` and `Statistics` require `<0.0.1, 1` instead of `1` to support older julia versions.

While running the tests, some deprecation warnings for `range(start, stop)` appeared.
They are also fixed in this PR.


oschulz (approved):

oschulz:
Please merge at your convenience,  @fhagemann

PR #380, opened 2024-03-26
Improvements to logo

fhagemann:
Now that I switched to dark mode on GitHub, I realized how ugly the white stripe around the "SSD" gamma-ray in the logo is.
image

<br/>

I have updated the logos to plot only the cutouts in the right detector edge.
Kind of a downside is that the original Plots/PyPlot versions do not run anymore and the tip of the "SSD" gamma-ray is not round but sharp. In my opinion, that does well with the rest of the font.
logo
which looks a bit better in dark mode:
image
<br/>
However, I've also added logos specifically for dark mode.
GitHub will choose the correct logo automatically depending on the color scheme used.
image


oschulz:
Beautiful!

PR #384, opened 2024-06-24
Add SSD --> GDML

bene73:
Added functionality to the Geant4 extension for the `Geant4.G4JLDetector()` function, which creates a Geant4 detector object (`Geant4.G4JLDetectorGDML`) from SSD simulation object by translating the given CSG geometry to GDML.

There are two methods for the function `Geant4.G4JLDetector()`, which makes it possible to specify the `SolidStateDetectors.Simulation` object explicitly or through a corresponding YAML configuration file:

```Julia
function Geant4.G4JLDetector(sim::SolidStateDetectors.Simulation, output_filename::String = "tmp.gdml"; verbose::Bool = true)
function Geant4.G4JLDetector(input_filename::String, output_filename::String = "tmp.gdml"; verbose::Bool = true)
```

The following CSG geometry types are covered in the SSD to GDML translation:
- Primitives: Box, Cone, Sphere
- Boolean operators: Difference, Intersection, Union

Displacements in the $x, y, z$-axes work fine, but **rotations are currently not parsed correctly and are thus still considered an experimental feature**.

Calling the function on a `SolidStateDetectors.Simulation` object, the following volumes are created:
- **World** (where the simulation takes place, is specified in SSD by the grid)
- **Semiconductor**
- **Contacts**
- **Passives**

Geometrical objects that have zero volume are omitted in the GDML file, since the simulation algorithm of Geant4 requires all objects to have nonzero volume.

The predefined list of materials (with their physical characteristics) saved in the file `materials.xml` is internally referenced by the names given in the SSD simulation object and gives each physical volume its properties needed for the Geant4 simulation. Currently, the list only includes:
- High Purity Germanium
- Aluminium
- Vacuum

### Example usage

Creates a Geant4 detector object from the _InvertedCoax_ example file.

```Julia
using SolidStateDetectors
using Geant4

g4_detector = Geant4.G4JLDetector(SSD_examples[:InvertedCoax], output_filename = "inverted_coax.gdml", verbose = false)
typeof(g4_detector) # G4JLDetectorGDML
```

### Needs to be implemented
- [ ] Geometry parsing for **Torus** and **Prism**
- [ ] Correct translation of (composite) rotations
- [x] Extend `materials.xml`
  - [ ] Add _PEN_, _PTFE_, _CdZnTe_


Project.toml
```diff
@@ -38,10 +39,11 @@ YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [weakdeps]
 Geant4 = "559df036-b7a0-42fd-85df-7d5dd9d70f44"
-LegendHDF5IO="c9265ca6-b027-5446-b1a4-febfa8dd10b0"
+LegendHDF5IO = "c9265ca6-b027-5446-b1a4-febfa8dd10b0"
+LightXML = "9c8b4983-aa76-5018-a973-4c85ecc9e179"
 
 [extensions]
-SolidStateDetectorsGeant4Ext = "Geant4"
+SolidStateDetectorsGeant4Ext = ["Geant4", "LightXML"]
```
fhagemann, line 46: Does it make sense to make `LightXML` an actual dependency to avoid having to load both `Geant4` and `LightXML` to load the extension?

ext/SolidStateDetectorsGeant4Ext.jl
```diff
@@ -4,32 +4,92 @@ module SolidStateDetectorsGeant4Ext
...
-function _ssdsim2gdml(io::IO, sim::SolidStateDetectors.Simulation)
-    # Dummy implementation:
-
-    println(io, """<?xml version="1.0" encoding="UTF-8" standalone="no" ?>""")
-    println(io, """<gdml>""")
-    println(io, """</gdml>""")
-end
-
+# Given an SSD simulation object, create corresponding GDML file to desired location
+function Geant4.G4JLDetector(sim::SolidStateDetectors.Simulation, output_filename::String = "tmp.gdml"; verbose::Bool = true)
```
fhagemann, line 18: `G4JLDetector` is an abstract type in `Geant4.jl`.
Should we rename this to `G4JLDetectorSSD` to stick to the notation `G4JLDetectorGDML` when reading `GDML` files?

ext/io_gdml.jl
```diff
@@ -0,0 +1,358 @@
...
+
+# position relative to origin of a primitive is stored in its parameters
+@inline parse_origin(e::AbstractVolumePrimitive) = origin(e)
+
+# position relative to the origin of a boolean solid is defined as position of
+# the leftmost child in the tree structure
+@inline parse_origin(e::AbstractConstructiveGeometry) = parse_origin(e.a)
+
+# Returns euler angles of the rotation matrix
+function parse_rotation_matrix(e::AbstractVolumePrimitive, v::Bool)
```
fhagemann, line 20: You could add some `@inline` macros here (and elsewhere)

ext/materials.xml
```diff
@@ -0,0 +1,17 @@
...
+        <atom value="72.63"/>
+    </material>
+    <element name="G4_Galactic" Z="1" formula="Galactic">
+      <atom value="1.01" unit="g/mole"/>
+    </element>
+    <material name="Vacuum" state="gas">
+      <D value="1e-25" unit="g/cm3"/>
+      <fraction n="1" ref="G4_Galactic"/>
+    </material>
+</materials>
```
fhagemann, line 17: Are these values in accordance with the Geant4 materials library defaults?

Project.toml
oschulz: Yes, let's make it a hard dependency, then the GDML export can also be a standard feature. LightXML won't add noticeably to SSD's load time.

ext/materials.xml
oschulz: Let's not use spaces in material names.

PR #387, opened 2024-07-03
Bug fix in `has_volume` for `Cones` with height 0

julianhenzler:
The `has_volume` function, which checks whether a geometry has a non-zero volume, had a bug for `Cone`.
The height of a `Cone` was never checked, so `Cones` with height 0 were wrongfully labeled as having a volume.
This is fixed now.

### DIFF
--- a/ext/io_gdml.jl
+++ b/ext/io_gdml.jl
@@ -109,9 +109,12 @@ function has_volume(e::Cone{T,<:Any,TR}, ::XMLElement, ::XMLElement, id::Integer
     if rmin >= rmax
         v && @warn "Cone $(pf * string(id)): Outer radius must be strictly bigger than inner radius"
         return false
-    else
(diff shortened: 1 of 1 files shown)


fhagemann (approved):
Thanks for spotting this!

ext/io_gdml.jl
```diff
@@ -110,7 +110,7 @@ function has_volume(e::Cone{T,<:Any,TR}, ::XMLElement, ::XMLElement, id::Integer
         v && @warn "Cone $(pf * string(id)): Outer radius must be strictly bigger than inner radius"
         return false
     else
-        return true
+        return e.hZ > 0
```
bene73, line 113: Should we add a warning here?

fhagemann: I added a warning.

bene73 (approved):

PR #388, opened 2024-07-04
GDML geometry parsing for composite rotations and Torus + Prism, bugfix in IO.jl

bene73:
Implemented correct geometry parsing for more complicated objects with composite rotations and more exotic geometries involving Tori and Prisms. Also, a problem was fixed regarding the YAML file parsing.

In GDML, relative rotations between linked geometrical objects have to be defined via the Euler angles (as opposed to CSG, where ultimately, only the rotation matrix of each Volume Primitive is stored). These angles are computed by determining the relative rotation matrix of the two respective volumes.

Given the rotation matrices $R_1$ and $R_2$ (for volume $V_1$ and $V_2$, respectively), the rotation of the second volume relative to the rotation of the first volume is $R^{-1}_1R_2$. Since we are working with rotation matrices, inversion is always possible and the inverse of the matrix $R_1$ simply corresponds to the transposed matrix.

---
While testing the newly implemented features, a bug in the SSD code became present, regarding the geometry parsing of nested transformations. The problem became present whenever more than one transformation (translation/rotation) was consecutively used as a separate block. For example, consider the following segment that may appear in the geometry description:
```YAML
translate:      # Ignored
    rotate:     # Parsed correctly
        ...
    ...
```
When parsing a geometry of that kind, only the outer transformation was applied previously, resulting in a geometry without the rotation. With the changes to `IO.jl`, all nested transformations should now work fine.

### DIFF
--- a/ext/io_gdml.jl
+++ b/ext/io_gdml.jl
@@ -1,7 +1,7 @@
 using SolidStateDetectors
 using SolidStateDetectors: Cylindrical, Cartesian, World, Simulation
 using SolidStateDetectors.ConstructiveSolidGeometry: CSGUnion, CSGDifference, CSGIntersection, 
-    Box, Cone, Ellipsoid, Torus, show_CSG_tree, AbstractVolumePrimitive, AbstractGeometry,
+    Box, Cone, Ellipsoid, Torus, RegularPrism, show_CSG_tree, AbstractVolumePrimitive, AbstractGeometry,
     AbstractConstructiveGeometry, CylindricalPoint, origin, rotation
 using IntervalSets
 using DataStructures: OrderedDict
@@ -18,14 +18,14 @@ using LightXML
 @inline parse_origin(e::AbstractConstructiveGeometry) = parse_origin(e.a)
 
 # Returns euler angles of the rotation matrix
-@inline function parse_rotation_matrix(e::AbstractVolumePrimitive, v::Bool)
-    m = RotXYZ(RotMatrix{3}(rotation(e)))
-    return m.theta1, m.theta2, m.theta3
+@inline function parse_rotation(e::AbstractVolumePrimitive)
+    # m = RotXYZ(RotMatrix{3}(rotation(e)))
+    # return m.theta1, m.theta2, m.theta3
+    return rotation(e)
 end
 
-@inline function parse_rotation_matrix(e::AbstractConstructiveGeometry, v::Bool)
-    v && @warn "Combining multiple rotations is an experimental feature"
-    return parse_rotation_matrix(e.a, v) .- parse_rotation_matrix(e.b, v)
+@inline function parse_rotation(e::AbstractConstructiveGeometry)
+    return parse_rotation(e.a)
 end
 
 
@@ -72,14 +72,19 @@ end
 
 # Add <rotation> to <define> section, referenced in geometry definition (in <solids>) via the name
 function create_rotation(e::AbstractConstructiveGeometry, x_define::XMLElement, name::String, v::Bool)
-    # 
-    x, y, z = parse_rotation_matrix(e, v)
+    R1 = RotMatrix{3}(parse_rotation(e.a))
+    R2 = RotMatrix{3}(parse_rotation(e.b))
+
+    R_rel = transpose(R1) * R2
+
+    m = RotXYZ(R_rel)
+
     x_rot = new_child(x_define, "rotation")
     set_attributes(x_rot, OrderedDict(
         "name" => name,
-        "x" => x,
-        "y" => y,
-        "z" => z,
+        "x" => m.theta1,
+        "y" => m.theta2,
+        "z" => m.theta3,
         "unit" => "rad" 
     ))
 end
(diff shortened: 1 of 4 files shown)


ext/io_gdml.jl
```diff
@@ -135,11 +139,44 @@ function has_volume(e::Ellipsoid{T,<:Any,T}, ::XMLElement, ::XMLElement, id::Int
...
+            v && @warn "Torus $(pf * string(id)): Inner tube radius must be non-negative, Outer tube radius must be positive"
+            return false
+        end
+    elseif e.r_tube <= 0
+        v && @warn "Torus $(pf * string(id)): Tube radius must be a positive number"
+        return false
+    end
+    if e.r_torus <= 0
+        v && @warn "Torus $(pf * string(id)): Torus radius must be a positive number"
+        return true
```
fhagemann, line 157: This should be `return false`?

ext/io_gdml.jl
```diff
@@ -280,8 +317,71 @@ function parse_geometry(e::Ellipsoid{T,<:Any, TR, Nothing, Nothing}, x_solids::X
...
+    if has_volume(e, x_solids, x_define, id, pf, v)
+        y = new_child(x_solids, "polyhedra")
+        
+        r = e.r
+        hZ = e.hZ
+        set_attributes(y, OrderedDict(
+            "name" => pf * string(id),
+            "startphi" => 0,
+            "deltaphi" => 360,
+            "numsides" => 6,
```
fhagemann, line 361: `"numsides"` is a parametric type of `RegularPrism`:
Only `RegularPrism{T,CO,6}` is a `HexagonalPrism`.
You could extract `"numsides"` from the parametric types.

```julia
function parse_geometry(e::RegularPrism{T, <:Any, N},...) where {T, N}
    # ...
    "numsides" => N
    # ...
end
```

ext/io_gdml.jl
```diff
@@ -280,8 +317,71 @@ function parse_geometry(e::Ellipsoid{T,<:Any, TR, Nothing, Nothing}, x_solids::X
     nothing
 end
 
-function parse_geometry(e::World{T, 3, Cylindrical}, x_solids::XMLElement, x_define::XMLElement, id::Integer, pf::AbstractString, v::Bool)::Nothing where {T}
+@inline parse_θ(::Type{T}, θ::Nothing) where {T} = (T(0), T(360), "deg")
+@inline parse_θ(::Type{T}, θ::Tuple{T, T}) where {T} = throw(AssertionError("Torus: θ values are currently not supported"))
```
fhagemann, line 321: Where is this ever used?
(If not at all: maybe comment this out, so we can uncomment this later)

ext/io_gdml.jl
```diff
@@ -135,11 +139,44 @@ function has_volume(e::Ellipsoid{T,<:Any,T}, ::XMLElement, ::XMLElement, id::Int
...
+    return true
+end
+
+
 # If composite geometry contains children with zero volume, omit its geometry definition
 # Omit the construction of a Boolean solid if some children have zero volume
-function has_volume(e::AbstractConstructiveGeometry, x_solids::XMLElement, x_define::XMLElement, id::Integer, pf::AbstractString, v::Bool; parse::Bool = true)
-    hva = has_volume(e.a, x_solids, x_define, 2 * id, pf, v, parse = false)
-    hvb = has_volume(e.b, x_solids, x_define, 2 * id + 1, pf, v, parse = false)
+function has_volume(e::AbstractConstructiveGeometry, x_solids::XMLElement, x_define::XMLElement, id::Integer, pf::AbstractString, v::Bool; parse::Bool = false)
```
fhagemann, line 177: This function does not cover the case for `CSGDifference` if after the difference the volume is zero, e.g. when a bigger volume is subtracted from a smaller volume.
Same thing for a non-overlapping `CSGIntersection`
However, I think that this will never be the case and that we do not have to worry about this.

@bene73 Could you check what happens in these cases and if Geant4 is happy reading in the GDML file or not?

fhagemann (commented):
In general, I would like to split to split the volume check and the parsing part in `has_volume`. 
The volume check does not require the `XMLElements` and might even go into the actual SSD package and not just in the Geant4 extension.

ext/io_gdml.jl
bene73: Apparently, Geant4 has no problem with volumes that have no points lying inside of them at all (e.g. intersection where $A\cap B=\emptyset$), and only a problem with infinitesimally thin surfaces (e.g. sphere with radius $r=0$).

Thus, intersections could theoretically still cause problems. Two objects can overlap only in an infinitesimally thin surface.

fhagemann: Then, I would say we do not have to worry about this!

ext/io_gdml.jl
```diff
@@ -17,10 +17,8 @@ using LightXML
 # the leftmost child in the tree structure
 @inline parse_origin(e::AbstractConstructiveGeometry) = parse_origin(e.a)
 
-# Returns euler angles of the rotation matrix
+# Returns rotation matrix for the Primitive or the leftmost child in the tree
 @inline function parse_rotation(e::AbstractVolumePrimitive)
-    # m = RotXYZ(RotMatrix{3}(rotation(e)))
-    # return m.theta1, m.theta2, m.theta3
     return rotation(e)
 end
```
fhagemann, line 23: Style-wise, this could just become a one-liner:
```julia
@inline parse_rotation(e::AbstractVolumePrimitive) = rotation(e)
```

fhagemann (commented):
Something, I would like to note (even though it does not need to be included in this PR):
We could reduce the number of arguments of `has_volume` to just
```julia
function has_volume(e::VolumePrimitive, v::Bool)
```
without `pf` and `id`. 
Even though the `@warn` message might be less descriptive, this function would be way easier to use from an outside perspective (in particular if `v` is set to false) and it could go into the main SSD package.
If we do this, I would also suggest to set the default value of `v` to `v::Bool = false`, so that it can just be called as
```julia
function has_volume(e::VolumePrimitive)
```

fhagemann (approved):

PR #395, opened 2024-08-02
Implemented basic function to run Geant4 simulation

bene73:
Added the function `run_geant4_simulation`, which allows the user to perform Geant4 simulations based on a given `G4JLApplication` and the amount of events that should be generated (`number_of_events`).

### Example: Generating 1000 interaction events from an electron source
In the following example, an electron emitting source is placed next to some detector (defined by the file `"detector_configuration.yaml"`). 
```julia
using SolidStateDetectors
using Geant4
using Unitful
sim = Simulation("detector_configuration.yaml")
source = MonoenergeticSource("e-", 1000u"keV", CartesianPoint(0.1, 0, 0), CartesianVector(-1, 0, 0))
application = G4JLApplication(sim, source)
events = run_geant4_simulation(application, 1000)
```
The returned value from the function has the type `TypedTables.Table` and contains all the events as a `DetectorHit` object (see [RadiationDetectorSignals.jl](https://github.com/JuliaPhysics/RadiationDetectorSignals.jl/blob/c837ec29d67de08d7b0491550886ef3eb726d854/src/detector_hits.jl#L16) for reference).

### DIFF
--- a/ext/SolidStateDetectorsGeant4Ext.jl
+++ b/ext/SolidStateDetectorsGeant4Ext.jl
@@ -23,7 +23,7 @@ include(joinpath(@__DIR__, "Geant4", "io_gdml.jl"))
 include(joinpath(@__DIR__, "Geant4", "g4jl_application.jl"))
 
 # Given an SSD simulation object, create corresponding GDML file to desired location
-function Geant4.G4JLDetector(sim::SolidStateDetectors.Simulation, output_filename::String = "tmp.gdml"; verbose::Bool = true)
+function Geant4.G4JLDetector(sim::SolidStateDetectors.Simulation, output_filename::String = "tmp.gdml"; verbose::Bool = true, save_gdml::Bool = false)
     # Create basis for GDML file
     x_doc = XMLDocument()
     x_root = create_root(x_doc, "gdml")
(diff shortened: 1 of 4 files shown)


src/MCEventsProcessing/MCEventsProcessing.jl
```diff
@@ -153,3 +153,4 @@ function _generate_waveform( drift_paths::Vector{<:EHDriftPath{T}}, charges::Vec
 end
 
 
+function run_geant4_simulation end
```
fhagemann, line 156: Are you planning on adding a docstring in this PR or will that happen separately?

ext/SolidStateDetectorsGeant4Ext.jl
```diff
@@ -159,4 +160,24 @@ function Geant4.G4JLApplication(
 end
 
 
+function SolidStateDetectors.run_geant4_simulation(app::G4JLApplication, number_of_events::Int)
+    evts = DetectorHit[]
+    
+    evtno = 1
+    while evtno <= number_of_events
+        out = missing
```
fhagemann, line 168: For type stability: could you change this to `out = DetectorHit[]`…

ext/SolidStateDetectorsGeant4Ext.jl
```diff
@@ -159,4 +160,24 @@ function Geant4.G4JLApplication(
...
+        while ismissing(out) || isempty(out)
```
fhagemann, line 169: … and ditch the `ismissing(out)` here?

ext/SolidStateDetectorsGeant4Ext.jl
```diff
@@ -159,4 +160,24 @@ function Geant4.G4JLApplication(
...
+function SolidStateDetectors.run_geant4_simulation(app::G4JLApplication, number_of_events::Int)
+    evts = DetectorHit[]
+    
+    evtno = 1
+    while evtno <= number_of_events
+        out = missing
+        while ismissing(out) || isempty(out)      
+            @suppress_out beamOn(app,1)
+            out = app.sdetectors["SensitiveDetector"][1].data.detectorHits
+            out = out[findall(!iszero, out.edep)]
```
fhagemann, line 172: Maybe this line could also be optimized using `filter!`

PR #405, opened 2024-09-01
2D slice plotting

hervasa2:
A long awaited feature has finally arrived!
2D slice plotting is now available via `seriestype = :slice`
```julia
plot(sim.electric_potential, φ = 0)
plot!(sim.detector, show_passives = false, st = :slice, φ = 0, legend = :topleft)
```
Screenshot 2024-09-03 at 17 24 37

It is based on the sample surface method. Only a subset of these points are selected and projected on a 2D plane which can be set with the `x,y,z,φ` attributes. Default is set to `y = 0`. Sampling density can be set as usual with `n_samples`. Even though the default density is set 5 times higher than for `:samplesurface` plots can be generated without lag.
Example:
```julia
plot(
    plot(sim.detector, st = :slice, x = 2u"mm"),
    plot(sim.detector, st = :slice, z = 0u"mm"),
    plot(sim.detector, st = :slice, φ = 90u"°"),
    layout = (1,3), size = (1000,300)
    )
```
Screenshot 2024-09-04 at 17 16 39

Sampling density was not set correctly when multiple CSGs where plotted at once (eg. `:samplesurface` in the plot recipe for SolidStateDetectors). This was corrected and carried over to `:slice`. 

Default values for `n_samples` are 40 and 100 or 200 for `samplesurface` and `:slice` respectively. `:slice` checks if any surface is coplanar with the plotting plane. In such cases  `n_samples = 100`, else `n_samples = 200`.

Looping over values for the plotting planes gives:

| Scan along `y` | Scan along `z` | Scan along `φ` |
| --- | --- | --- |
| tmp (9) | tmp (10) | tmp (11) |
| tmp (6) | tmp (7) | tmp (8) |

### DIFF
--- a/src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
+++ b/src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
@@ -6,71 +6,62 @@
     )
 end
 
+function _get_csg_plot_objects(csg::AbstractConstructiveGeometry{T}, primitive, n_samples, CSG_scale, slice_val, plotattributes) where {T}
+    projections = [:x, :y, :z]
+    if plotattributes[:seriestype] in vcat(projections,:samplesurface)
+        spacing = T((ismissing(CSG_scale) ? get_scale(csg) : CSG_scale)/n_samples)
+        samples_csg = filter(p -> in(p,csg), sample(primitive, spacing))
+        if plotattributes[:seriestype] in projections
+            samples = filter(pt -> abs(getproperty(pt, plotattributes[:seriestype]) - slice_val) < spacing/2, samples_csg)
+            proj = filter(x -> x != plotattributes[:seriestype], projections)
+            internal_length_unit*getproperty.(samples, proj[1]), internal_length_unit*getproperty.(samples, proj[2])
+        else
+            samples_csg
+        end
+    else
+        primitive
+    end
+end
+
 @recipe function f(csg::AbstractConstructiveGeometry{T}; n_samples = 40, CSG_scale = missing, slice_val = T(0)) where {T}
     seriestype --> :csg
-    xunit --> internal_length_unit
-    yunit --> internal_length_unit
+    unitformat --> :slash
     projections = [:x, :y, :z]
     ps = primitives(csg)
-    if haskey(plotattributes, :seriestype) && plotattributes[:seriestype] in vcat(projections,:samplesurface)
-        spacing::T = T((ismissing(CSG_scale) ? get_scale(csg) : CSG_scale)/n_samples)
-    end
     @series begin
         label --> "CSG"
-        if haskey(plotattributes, :seriestype) 
-            if plotattributes[:seriestype] == :samplesurface
-                zunit --> internal_length_unit
-                filter(p -> in(p,csg), sample(ps[1], spacing))
-            elseif plotattributes[:seriestype] in projections
-                samples = filter(pt -> abs(getproperty(pt, plotattributes[:seriestype]) - slice_val) < spacing/2, filter(p -> in(p,csg), sample(ps[1], spacing)))
-                proj = filter(x -> x != plotattributes[:seriestype], projections)
-                xguide --> string(proj[1])
-                yguide --> string(proj[2])
-                unitformat --> :slash
-                internal_length_unit*getproperty.(samples, proj[1]), internal_length_unit*getproperty.(samples, proj[2])
-            else
-                zunit --> internal_length_unit
-                if !isClosedPrimitive(ps[1])
-                    fillcolor := :white
-                    fillalpha --> 0.2
-                    if haskey(plotattributes, :seriestype) 
-                        if plotattributes[:seriestype] in [:csg, :wireframe] 
-                            linewidth := 1
-                        end
-                    end
-                end
-                ps[1]
+        if plotattributes[:seriestype] in projections
+            proj = filter(x -> x != plotattributes[:seriestype], projections)
+            xunit --> internal_length_unit
+            yunit --> internal_length_unit
+            xguide --> string(proj[1])
+            yguide --> string(proj[2])
+        else
+            if !isClosedPrimitive(ps[1])
+                fillcolor := :white
+                fillalpha --> 0.2
+                linewidth := 1
             end
         end
+        _get_csg_plot_objects(csg, ps[1], n_samples, CSG_scale, slice_val, plotattributes)
     end
     for i in 2:length(ps)
         @series begin
             label := ""
-            if haskey(plotattributes, :seriestype) 
-                if plotattributes[:seriestype] == :samplesurface
-                    seriesalpha --> 0.2
-                    filter(p -> in(p,csg), sample(ps[i], spacing))
-                elseif plotattributes[:seriestype] in projections
-                    samples = filter(pt -> abs(getproperty(pt, plotattributes[:seriestype]) - slice_val) < spacing/2, filter(p -> in(p,csg), sample(ps[i], spacing)))
-                    proj = filter(x -> x != plotattributes[:seriestype], projections)
-                    xguide --> string(proj[1])
-                    yguide --> string(proj[2])
-                    unitformat --> :slash
-                    internal_length_unit*getproperty.(samples, proj[1]), internal_length_unit*getproperty.(samples, proj[2])
-                else
-                    if !isClosedPrimitive(ps[i])
-                        fillcolor := :white
-                        fillalpha --> 0.2
-                        if haskey(plotattributes, :seriestype) 
-                            if plotattributes[:seriestype] in [:csg, :wireframe] 
-                                linewidth := 1
-                            end
-                        end
-                    end
-                    ps[i]
+            if plotattributes[:seriestype] in projections
+                proj = filter(x -> x != plotattributes[:seriestype], projections)
+                xunit --> internal_length_unit
+                yunit --> internal_length_unit
+                xguide --> string(proj[1])
+                yguide --> string(proj[2])
+            else
+                if !isClosedPrimitive(ps[i])
+                    fillcolor := :white
+                    fillalpha --> 0.2
+                    linewidth := 1
                 end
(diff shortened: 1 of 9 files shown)


fhagemann:
At first glance, this looks very promising to me. Are you planning to write some text for the plotting documentation?
https://juliaphysics.github.io/SolidStateDetectors.jl/stable/man/plotting/

hervasa2:
Yes. But first I would like to consider if this is the most  intuitive front end  `slice_plane` selection. What do you think of it now versus having  `x,y,...`  as keyword arguments?

fhagemann:
I think that having `x`, `y`,.. as keyword argument might indeed be more intuitive, especially keeping in mind that most users would use this together with potential slice plot.

fhagemann (changes requested):
I haven't tried out the code yet but these are my first comments from my subway ride back home.

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -6,59 +6,71 @@
     )
 end
 
-@recipe function f(csg::AbstractConstructiveGeometry{T}; n_samples = 40, CSG_scale = missing) where {T}
+@recipe function f(csg::AbstractConstructiveGeometry{T}, primitive::AbstractPrimitive{T}, axis::Symbol, slice::T, st::Symbol; n_samples = missing, CSG_scale = missing) where {T}
+    if st == :samplesurface
+        CSG_scale = ismissing(CSG_scale) ? get_scale(csg) : CSG_scale
+        spacing = T(CSG_scale/n_samples)
```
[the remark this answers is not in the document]
hervasa2: yes, but this is never meant to be called directly. It isn internal plot recipe only called by 
```julia 
@recipe function f(csg::AbstractConstructiveGeometry{T})
```

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -6,59 +6,71 @@
...
+        filter(p -> in(p,csg), sample(primitive, spacing))
+    elseif st == :slice
+        CSG_scale = ismissing(CSG_scale) ? get_scale(csg) : CSG_scale
+        pointtype, axisrot = CartesianPoint, axis
+        spacing = T(CSG_scale/n_samples)
```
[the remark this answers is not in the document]
hervasa2: See above about it being an internal plot recipe

src/ConstructiveSolidGeometry/plotting/SurfacePrimitives/SurfacePrimitives.jl
```diff
@@ -9,49 +9,65 @@
...
-    seriesalpha --> 0.2
+    seriesalpha --> 0.1
+    markerstrokewidth --> 0
+    markersize --> 3
+    if occursin("GRBackend", string(typeof(plotattributes[:plot_object].backend)))
+        aspect_ratio --> 1.0
+    end 
+    seriestype := :scatter
+    ()
+end
```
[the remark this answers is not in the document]
hervasa2: Yes, I noticed that with the new plot versions they were getting too dark. This looks better overall.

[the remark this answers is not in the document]
fhagemann: I see, thanks!

fhagemann:
I tried out the code, works very smoothly. Some questions I would have:
- [ ] The plot recipe is based on sampling points and plotting them as `scatter` seriestype.
   What would be the recommended syntax to draw bigger/thinner lines?
- [ ] Regarding the point above: Would it make sense to use the `linewidth` series attribute to also be able to modify the thickness of the contacts in the slice plot? The slice plot looks basically like a line and might have people thinking that it is actually a `path` plot, where the line thickness could be adjusted via `linewidth`/`lw`.
- [ ] When making slices in `ϕ`, is there any convenient way to let the method also sample points in "negative" `ϕ` direction, similar to the `full_det` keyword in the electric potential plotting?
  image
- [ ] Can we add simple examples explaining how to use the `slice` seriestype to the documentation?

src/ConstructiveSolidGeometry/plotting/SurfacePrimitives/SurfacePrimitives.jl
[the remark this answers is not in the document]
fhagemann: Just to have it documented somewhere: This is how the plots evolve based on this change.

### Before
image
### After
image

hervasa2:
All the requested changes have been pushed. `linewidth/lw` is reinterpreted as `markersize` in the internal plot recipes for `st = :slice`. `full_det` is now also suported. Additionally when a `z` slice is chosen the `projection/proj = :polar` keyword is supported.
```julia
  plot(sim.electric_potential, z = 30u"mm")
  plot!(sim.detector, st = :slice, z = 30u"mm", proj = :polar, lw = 2)
```
Screenshot 2024-09-05 at 15 42 21

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -6,33 +6,44 @@
...
         xguide --> string(proj[1])
         yguide --> string(proj[2])
-        internal_length_unit*u, internal_length_unit*v
+        if linewidth != :auto
+            markersize := linewidth
+        end
+        if projection == :polar && axis == :z
+            if isgr aspect_ratio := :none end
+            xunit --> internal_angle_unit
+            internal_angle_unit*atan.(v,u), internal_length_unit*sqrt.(u.^2 + v.^2)
```
fhagemann, line 41: I feel like we shouldn't multiply `internal_angle_unit` here, because `atan` will always return the angle `rad` (likewise you could use `atand` to return it in degrees). I know the internal angle unit is rad, but if it should change to degree, this would break.

What you could do would be:
```julia
# Instead of internal_angle_unit*atan.(v,u)
convert.(internal_angle_unit, atan.(v,u))
# or just
atan.(v,u)
```

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -6,33 +6,44 @@
...
-        xunit --> internal_length_unit
         yunit --> internal_length_unit
         xguide --> string(proj[1])
         yguide --> string(proj[2])
-        internal_length_unit*u, internal_length_unit*v
+        if linewidth != :auto
+            markersize := linewidth
+        end
+        if projection == :polar && axis == :z
+            if isgr aspect_ratio := :none end
```
fhagemann, line 39: Is this really needed?

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -6,33 +6,44 @@
...
         end
         proj = filter(x -> x != axis, fieldnames(pointtype))
-        xunit --> internal_length_unit
         yunit --> internal_length_unit
         xguide --> string(proj[1])
         yguide --> string(proj[2])
-        internal_length_unit*u, internal_length_unit*v
+        if linewidth != :auto
+            markersize := linewidth
+        end
```
fhagemann, line 37: Didn't try this out, but does this also work with `lw`?

I would naively assume, that one would need something like `linewidth = seriesattribute[:linewidth]` in case the user uses an alias.

src/PlotRecipes/SolidStateDetector.jl
```diff
@@ -66,6 +66,7 @@ end
     seriestype --> :csg
     show_normal --> false
     st = plotattributes[:seriestype]
+    isgr = occursin("GRBackend", string(typeof(plotattributes[:plot_object].backend)))
```
fhagemann, line 69: Seems to me like it is not used --> can be removed.

src/ConstructiveSolidGeometry/plotting/VolumePrimitives/VolumePrimitives.jl
```diff
@@ -50,17 +47,27 @@ end
...
+        end
+        if projection == :polar && axis == :z
+            if isgr aspect_ratio := :none end
+            xunit --> internal_angle_unit
+            internal_angle_unit*atan.(v,u), internal_length_unit*sqrt.(u.^2 + v.^2)
+        else
+            if isgr aspect_ratio --> 1.0 end
+            xunit --> internal_length_unit
+            internal_length_unit*u, internal_length_unit*v
+        end
```
fhagemann, line 70: See comments above

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
hervasa2: Yes, `lw` and `proj` are supported by having `linewidth` and `projection` in the keywords of the plot recipe.

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
hervasa2: The only option would be 
```julia
uconvert.(internal_angle_unit, atan.(v,u))
```
since `atan` by itself does not return a unitful quantity

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
fhagemann: `aspect_ratio = :none`

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
fhagemann: Yes sorry, `uconvert` instead of `convert`

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
hervasa2: Yeah not needed in the end

fhagemann (approved):

PR #408, opened 2024-09-11
Replace `RealQuantity` with `Unitful.RealOrRealQuantity`

### LINKED ISSUE #404: Where/how to define `RealQuantity`
In SSD, we define `MaybeWithUnits` (never really used) and `RealQuantity` as follows:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/d0b4d57685b8305e8a18ca8741a094650200e938/src/Units.jl#L98-L99

This is an exact replica of what is defined in `RadiationDetectorSignals`:
https://github.com/JuliaPhysics/RadiationDetectorSignals.jl/blob/main/src/numeric_types.jl

Also, there is `RealOrRealQuantity` in `Unitful` which does the same thing:
https://github.com/PainterQubits/Unitful.jl/blob/317ba73cd8efb98868b36afd5a518134e120fb76/src/types.jl#L238

Could we clean up where/how we define `RealQuantity` (same applies to `RadiationDetectorSignals`)?

```julia
using RadiationDetectorSignals: RealQuantity # ?
using Unitful: RealOrRealQuantity as RealQuantity # ?
```

fhagemann:
Closes #404 .

Side note: `SolidStateDetectors.RealQuantity` still works, so this is not a breaking change (except for getting rid of `MaybeWithUnits{T}` which we were not really using).

### DIFF
--- a/src/Event/Event.jl
+++ b/src/Event/Event.jl
@@ -56,12 +56,12 @@ end
 
 function Event(locations::Vector{<:AbstractCoordinatePoint{T}}, energies::Vector{<:RealQuantity}, N::Int; 
                particle_type::Type{PT} = Gamma, number_of_shells::Int = 2,
-               radius::Vector{<:RealQuantity{T}} = radius_guess.(T.(to_internal_units.(energies)), particle_type)
+               radius::Vector{<:RealQuantity} = radius_guess.(T.(to_internal_units.(energies)), particle_type)
               )::Event{T} where {T <: SSDFloat, PT <: ParticleType}
     
     return Event(broadcast(i -> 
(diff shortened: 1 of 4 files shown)


fhagemann:
@oschulz: I also added PRs in all other packages that used a custom defintion of `RealQuantity`:
- [ ] https://github.com/JuliaPhysics/RadiationDetectorSignals.jl/pull/31
- [ ] https://github.com/JuliaPhysics/RadiationDetectorDSP.jl/pull/44
- [ ] https://github.com/legend-exp/LegendDSP.jl/pull/40
- [ ] https://github.com/legend-exp/LegendEventAnalysis.jl/pull/7

oschulz:
Thanks, I've merged the other PRs.

PR #411, opened 2024-09-12
Reintroduce `Requires` dependency for `LegendHDF5IO`

fhagemann:
As `LegendHDF5IO` is not registered in General or stdlib, we cannot release v0.10.0 with `LegendHDF5IO` in the `[weakdeps]`. A quick workaround is to reintroduce `Requires` to deal with this.

Long term plans could then be:
- register `LegendHDF5IO` or a part of it
- move the extension to `LegendHDF5IO`
- simply keeping the `Requires` dependency

### DIFF
--- a/ext/SolidStateDetectorsLegendHDF5IOExt.jl
+++ b/ext/SolidStateDetectorsLegendHDF5IOExt.jl
@@ -2,7 +2,7 @@
 
 module SolidStateDetectorsLegendHDF5IOExt
 
-import LegendHDF5IO
+import ..LegendHDF5IO
 
 using SolidStateDetectors
 using SolidStateDetectors: RealQuantity, SSDFloat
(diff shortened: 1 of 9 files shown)


ext/SolidStateDetectorsLegendHDF5IOExt.jl
```diff
@@ -2,7 +2,7 @@
 
 module SolidStateDetectorsLegendHDF5IOExt
 
-import LegendHDF5IO
+import ..LegendHDF5IO
```
fhagemann, line 5: @oschulz In the old code, `LegendHDF5IO` had to be imported with `..` before the package name if it was loaded via `Requires`.
Can you confirm that this is correct and needed?

oschulz: `import ..LegendHDF5IO` seems to be the recommended way.

PR #412, opened 2024-09-17
Relax type constraints of config dictionaries

fhagemann:
When reading in config files, the `dicttype` can be given. This allows to keep the order of entries, e.g. when using `OrderedDict`.

Also: `OrderedDict` in `DataStructures` is just imported and reexported from `OrderedCollections`.
I've changes the dependencies to only depend on `OrderedCollections` instead of `DataStructures`.


fhagemann:
Just for the record: the change of `[compat]` is needed because of https://github.com/JuliaData/YAML.jl/pull/87/

PR #414, opened 2024-09-22
Implement charge trapping models

fhagemann:
In this PR, the `Semiconductor` is extended by another field (`charge_trapping_model`), which (for now) is passed to `add_signal!` to include the effects of charge trapping.

So far, there are
* `NoChargeTrappingModel`: ignore charge trapping --> same results as before.
* `BoggsChargeTrappingModel`: Charge trapping model based on [Boggs _et al._ (2023)](https://doi.org/10.1016/j.nima.2023.168756).

In the future, `ChargeTrappingModels` might also be passed to `drift_charges!`, in case they should influence the charge drift itself and not just the signal development afterwards.

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -3,11 +3,13 @@ on:
   pull_request:
     branches:
       - main
+      - dev
     paths-ignore:
       - 'docs/**'
   push:
     branches:
       - main
+      - dev
     paths-ignore:
       - 'docs/**'
 jobs:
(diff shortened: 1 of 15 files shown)


fhagemann:
Still missing:
- [ ] How to define a charge trapping model in the config file
- [ ] Parse functions to read charge trapping models (and parameters) in.
- [ ] Choose reasonable default values.

fhagemann:
The documentation preview can be found here: https://juliaphysics.github.io/SolidStateDetectors.jl/previews/PR414/

oschulz (approved):

PR #428, opened 2024-11-05
Bug fix in `LegendHDF5IO` extension and restructure tests

fhagemann:
The current implementation of `simulate_waveforms` does not work because of some missing packages that have to be explicitly loaded in the `LegendHDF5IO` extension.

Also, I restructured the test directory and added tests for `simulate_waveforms` using `LegendHDF5IO`.


test/test_io.jl
```diff
@@ -0,0 +1,112 @@
...
+            number_of_shells = 2,
+            verbose = false
+        )
+        @info isfile("generated_waveforms_evts_1-1.h5")
+        @test contact_charge_signals == LegendHDF5IO.lh5open("generated_waveforms_evts_1-1.h5") do h5f
+            @test haskey(h5f, "generated_waveforms")
+            LegendHDF5IO.readdata(h5f.data_store, "generated_waveforms")
+        end
+        rm("generated_waveforms_evts_1-1.h5")
+    end
```
fhagemann, line 111: This is the new test, which might get lost in the number of code changes here.

PR #431, opened 2024-11-11
Add IsotropicChargeDriftModel and add more materials

Padniuk:
This model was called as IsotropicMobilityChargeDriftModel because it uses identity-like mobility tensor:

$$v_{i}^{e,h}=\mu_{ij}^{e,h}\cdot E_{j}^{e,h} = \mu^{e,h}\cdot E_{i}^{e,h}$$

The semiconductor should have material properties μ_e and μ_h. Like in example:

``` Julia
    μ_e = 1000u"cm^2/V/s",
    μ_h = 1000u"cm^2/V/s"
```

The usage examples are the following:

``` Julia
    sim.detector = SolidStateDetector(sim.detector, IsotropicChargeDriftModel{T}(1, 1)) # defaul units are m^2/V/s
    sim.detector = SolidStateDetector(sim.detector, IsotropicChargeDriftModel{T}(100u"cm^2/V/s", 1400u"cm^2/V/s"))
    sim.detector = SolidStateDetector(sim.detector, IsotropicChargeDriftModel("<path_to_config_file>"))
    sim.detector = SolidStateDetector(sim.detector, IsotropicChargeDriftModel()) # use default config in examples/example_config_files/IsotropicChargeDriftModel/
```
Additionally two new materials were added: $CsPbBr_{3}$ and lead:
``` julia
abstract type CsPbBr3 <: AbstractDriftMaterial end
Symbol(::Type{CsPbBr3}) = :CsPbBr3
material_properties[:CsPbBr3] = (
    name = "CsPbBr3",
    E_ionisation = 6.9u"eV",
    ϵ_r = 16.46,
    ρ = 4.73u"g*cm^-3"
)

material_properties[:Pb] = (
    name = "Lead",
    ϵ_r = 1,
    ρ = 11.35u"g*cm^-3"
)
```


fhagemann (commented):
This, in principle, looks good to me. I will have a detailed look st this later.
Could you maybe separate your first commit into two (one introducing the model, the other one adding new materials)?

src/MaterialProperties/MaterialProperties.jl
```diff
@@ -89,6 +89,21 @@ material_properties[:CdZnTe] = (
...
+material_properties[:CsPbBr3] = (
+    name = "CsPbBr3",
+    E_ionisation = 6.9u"eV",
+    ϵ_r = 16.46,
+    ρ = 4.73u"g*cm^-3"
+)
+
+material_properties[:Pb] = (
+    name = "Lead",
+    ϵ_r = 1,
```
fhagemann, line 104: Is this vale for lead correct? I would have expected a higher number

Padniuk: As I know relative permitive of metals is equal to 1 for all metals. I know that for the you have an aluminum foil with 10.8 value, but for the clear metals is one as I [see](https://en.wikipedia.org/wiki/Relative_permittivity#:~:text=r%20remains%20unitless\).-,Metals,relative%20permittivity%20equal%20to%20one.) . Aluminum foil is not a clear aluminum, so probably that's why we see that.

Padniuk: In general I add lead just as material for collimator. Electric field should not be applied to it.

fhagemann: This way, „lead“ will be treated as if it was vacuum in the field calculation. This is fine if the detector is shielded from the collimator by some cryostat. But the presence of materials in the vicinity of the detector can have a significant impact on the field (also inside of the detector if it has a lot of uncontacted surface)

Padniuk: I agree with you. It is better to put some huge value. I have done it, and have split request into two commits.

fhagemann (approved):
Looks good to me.
I will spend some time to write tests for this, but afteerwards, this should be good to go.

fhagemann:
Sorry for all this force pushing, but the tests I added should also be good to go now.
I would wait until the tests pass and then merge ASAP.

fhagemann (approved):

oschulz (approved):

fhagemann:
@Padniuk Thanks for the PR!

We will probably leave this on `main` for some time before releasing it for some test period. If you already want to use it now, you could load SolidStateDetectors using 
```julia
]add SolidStateDetectors#main
```

PR #438, opened 2024-11-26
Fix constructor of `BoggsChargeTrappingModel`

fhagemann:
Previously, the constructor of `BoggsChargeDriftModel` would just completely ignore the section with the `parameters`.
Now, they are properly parsed.

The constructor can now take two sorts of dictionaries:
```
model: Boggs
parameters:
  # ...
```
and just
```
# ...
```
The latter might be convenient when defining the `BoggsChargeDriftModel` in code, e.g. `BoggsChargeDriftModel{T}(Dict("nσe-1" => "1000cm"))` and ensures that this PR is non-breaking (this is what the previous method did).


fhagemann:
With the last push, the constructor now performs as explained in the documentation:
https://juliaphysics.github.io/SolidStateDetectors.jl/stable/man/charge_drift/#BoggsChargeTrappingModel

PR #440, opened 2024-12-10
Update documentation on charge densities

fhagemann:
Some parts of the documentation on charge densities were wrong or not up-to-date.
This is fixed in this PR.


fhagemann:
The latest commit removes obsolete entries in `unit_conversion` and adds `e` to be interpreted as `elementary_charge`.
`construct_unit` also allows interpreting strings from the module `UnitfulAtomic`.

PR #441, opened 2024-12-12
Minor update in painting algorithm for `CylindricalGrid`

fhagemann:
The `csgtol` in the painting algorithm takes into account that grid points can be unevenly spaced in different dimensions. However, for 2D grids, this results in very weird results. Also, here we are comparing quantities in units of length with quantities in units of radians. I have removed the angle quantities from the `csgtol` check.

### DIFF
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
@@ -230,6 +230,9 @@ function intersection(cm::ConeMantle{T,Tuple{T,T}}, l::Line{T}) where {T}
     end
     ints1 = obj_l.origin + λ1 * obj_l.direction 
     ints2 = obj_l.origin + λ2 * obj_l.direction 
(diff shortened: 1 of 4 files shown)


fhagemann:
The choice of `Δw_max_factor` influences how well the `TorusMantle` is painted when using `Float32`:

With default `Δw_max_factor = 1e-5`:
image

With increased `Δw_max_factor = 1e-4`:
image

Difference:
image

I am going to investigate whether this is the result of a poor choice of `Δw_max_factor` (and whether it should be increased) or some rounding errors in the implementation of `intersection` with `TorusMantle`.

### Side note

The painting looks fine with `Float64` and `Δw_max_factor = 1e-5`:
image

Difference:
image

Therefore, I believe that this might be caused by some rounding errors in `intersection` with `TorusMantle`, where we have to find the roots of a fourth-order polynomial.

fhagemann:
After increasing the precision in `intersection` of `TorusMantle`, 
this is what it looks like now with `Float32` and the default `Δw_max_factor = 1e-5`:

image

Difference:
image

PR #444, opened 2024-12-19
Add alias `deg` for `°` in `ConstructiveSolidGeometry`

fhagemann:
This will define `u”deg“` when explicitly calling `using SolidStateDetectors.ConstructiveSolidGeometry` and otherwise allow users to use ”deg“ as string in config dicts and files when just defining geometries (just in the CSG cosmos, without the need of any SSD functionality).
This could become interesting if we might wanna export CSG to a separate module.

### DIFF
--- a/src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
+++ b/src/ConstructiveSolidGeometry/ConstructiveSolidGeometry.jl
@@ -118,4 +118,15 @@ module ConstructiveSolidGeometry
     # Plotting
     include("plotting/plotting.jl")
 
+    # define the alias "deg" for ° --> this will also allow to use deg in config files
+    # u"deg" will be defined when explicitly calling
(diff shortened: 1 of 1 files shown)


oschulz:
Unitful does provide `u"°"`.

(I'm worried that it's also bit too "name grabby" for a high-level package like SolidStateDetectors to define `u"deg"`).

fhagemann:
Then I‘m closing this, but we should either deprecate the use of `deg` in the config file then or find a way to parse `deg` within the CSG module without the `construct_unit` workaround which is defined in SSD and not in CSG.

oschulz:
> deprecate the use of deg

We could deprecate `deg` in favor of `°`.

fhagemann:
--> `u"deg"` now exists in Unitful.jl, even though it's not yet released
https://github.com/PainterQubits/Unitful.jl/pull/764

PR #451, opened 2025-01-22
Add `Polycone`

### LINKED ISSUE #449: Implementation of Polycone geometries
In Geant4, there are objects called [G4GenericPolycone](https://geant4.kek.jp/Reference/11.1.1/classG4GenericPolycone.html), which define volumes based on a set of points that are then rotated around the z-axis. In `Geant4.jl@0.2.1`, they are also supported.

Image
https://slideplayer.com/slide/9412278/

We should have a look how easy it would be to implement such a geometry also in SSD, including all functionalities like `in`, plotting, ...

fhagemann:
I added `polycone` to the geometries, which can be defined as a set of points in `r` and `z`:
```yaml
polycone:
  r: [0, 35, 35, 24.42, 5, 5, 0]
  z: [0, 0, 20, 80, 80, 25, 25]
```
It follows the same logic as the [G4GenericPolycone](https://geant4.kek.jp/Reference/11.1.1/classG4GenericPolycone.html).

Note: I also updated the example inverted coaxial detector geometry to use `polycone` instead of a difference of `tubes` and `cones`. The results seem to be identical (all tests still pass). However, we can discuss to revert this and to keep the original file as is (especially using `difference`, which might be good guidance for people wanting to do more complicated CSG stuff.

This should close #449.


test/test_geant4.jl
```diff
@@ -56,7 +56,7 @@ end
 @testset "Run Geant4 simulations" begin
 
     # Create G4JLApplication
-    sim = Simulation{T}(SSD_examples[:InvertedCoax])
+    sim = Simulation{T}(SSD_examples[:InvertedCoaxInCryostat])
```
fhagemann, line 59: This way, we will also try out parsing another material and more geometries.

test/test_geant4.jl
```diff
@@ -8,7 +8,7 @@ using RadiationDetectorSignals
 using TypedTables
 using Unitful
 
-T = Float32
+T = Float64
```
fhagemann, line 11: Currently, `Polycone` has no implementation of `csgtol` which will cause some points that are "inside" the `Polycone` in `Float64` to be "outside" in `Float32`

fhagemann: --> hopefully resolved in 1978adb

PR #468, opened 2025-03-10
Update is_depleted

hervasa2:
`is_depleted` currently does not check if a grid point is bulk type. This was updated to correctly determine if the detector is depleted based only on points in the bulk. 

This was causing `estimate_depletion_voltage` to fail for many clearly fully depleted examples. The checking of depletion in this function was also made optional if the user makes their own checks beforehand.


fhagemann (approved):
This looks reasonable to me

PR #469, opened 2025-03-11
Allow user to refine an already simulated potential

hervasa2:
If user wants to refine an existing simulation, they can do so by using the `initialize = false` keyword. See example below.
```julia
using SolidStateDetectors
using BenchmarkTools

T = Float32
calculate_electric_potential!(sim, refinement_limits = [0.2], depletion_handling = true)
```
Simulation: Public Inverted Coax
Electric Potential Calculation
Bias voltage: 3500.0 V
φ symmetry: Detector is φ-symmetric -> 2D computation.
Precision: Float32
Device: CPU
Max. CPU Threads: 1
Coordinate system: Cylindrical
N Refinements: -> 1
Convergence limit: 1.0e-7  => 0.00035 V
Initial grid size: (24, 1, 20)

Grid size: (30, 1, 30) - using 1 threads now
```julia
calculate_electric_potential!(sim, refinement_limits = [0.1], depletion_handling = true, initialize = false)
```
Simulation: Public Inverted Coax
Electric Potential Calculation
Bias voltage: 3500.0 V
φ symmetry: Detector is φ-symmetric -> 2D computation.
Precision: Float32
Device: CPU
Max. CPU Threads: 1
Coordinate system: Cylindrical
N Refinements: -> 1
Convergence limit: 1.0e-7  => 0.00035 V
Initial grid size: (30, 1, 30)

Grid size: (40, 1, 50) - using 1 threads now
Clearly the initial grid for the second refinement is the result of the first calculation. The computation time also spreads out over both computations. See computation time below for simulating in one step vs 2 steps
```julia
@btime calculate_electric_potential!(sim, refinement_limits = [0.2, 0.1], depletion_handling = true, verbose = false)
```
128.599 ms (1239795 allocations: 291.36 MiB)
```julia
@btime calculate_electric_potential!(sim, refinement_limits = [0.2], depletion_handling = true, verbose = false)
```
71.188 ms (756028 allocations: 136.88 MiB)
```julia
@btime calculate_electric_potential!(sim, refinement_limits = [0.1], depletion_handling = true, verbose = false, initialize = false)
```
22.563 ms (180071 allocations: 61.19 MiB)

### DIFF
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -829,6 +829,7 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         paint_contacts::Bool = true,
         verbose::Bool = true,
         device_array_type::Type{<:AbstractArray} = Array,
+        initialize::Bool = true
     )::Nothing where {T <: SSDFloat, CS <: AbstractCoordinateSystem}
 
     begin # preperations
@@ -836,6 +837,9 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         convergence_limit::T = T(convergence_limit)
         isEP::Bool = potential_type == ElectricPotential
         isWP::Bool = !isEP
+        if !initialize
+            grid = isEP ? sim.electric_potential.grid : sim.weighting_potentials[contact_id].grid
+        end
         if ismissing(grid)
             grid = Grid(sim, for_weighting_potential = isWP, max_tick_distance = max_tick_distance, max_distance_ratio = max_distance_ratio)
         end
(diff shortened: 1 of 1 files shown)


src/Simulation/Simulation.jl
```diff
@@ -829,13 +829,17 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
     )::Nothing where {T <: SSDFloat, CS <: AbstractCoordinateSystem}
 
     begin # preperations
         onCPU = !(device_array_type <: GPUArrays.AnyGPUArray)
         convergence_limit::T = T(convergence_limit)
         isEP::Bool = potential_type == ElectricPotential
         isWP::Bool = !isEP
+        if !initialize
+            grid = isEP ? sim.electric_potential.grid : sim.weighting_potentials[contact_id].grid
+        end
```
fhagemann, line 842: How about you put this as default in line 817? Instead of
```julia
grid::Union{Missing, Grid{T}} = missing
```
put something like
```julia
grid::Union{Missing, Grid{T}} = initialize ? missing : (isEP ? sim.electric_potential.grid : sim.weighting_potential[contact_id].grid)
```

fhagemann (commented):
Could you add something to the tests to ensure that it works?

test/test_refine_existing_potential.jl
```diff
@@ -0,0 +1,21 @@
...
+    calculate_electric_potential!(sim2, refinement_limits = [0.2], depletion_handling = true)
+    calculate_electric_potential!(sim2, refinement_limits = [0.1], depletion_handling = true, initialize = false)
+
+    @test size(sim.electric_potential.data) == size(sim2.electric_potential.data)
+
+    calculate_weighting_potential!(sim, 1, refinement_limits = [0.2, 0.1], depletion_handling = true)
+    calculate_weighting_potential!(sim2, 1, refinement_limits = [0.2], depletion_handling = true)
+    calculate_weighting_potential!(sim2, 1, refinement_limits = [0.1], depletion_handling = true, initialize = false)
+
+    @test size(sim.weighting_potentials[1].data) == size(sim2.weighting_potentials[1].data)
```
fhagemann, line 20: Just a question out of curiosity: is the `grid` also equal (same ticks) or just the `size` of the data?

fhagemann (commented):
Thanks!

fhagemann:
Tests were failing because `Geant4` exports `size` and it conflicted with `Base.size`.
This should be (hopefully) fixed now.

hervasa2:
Could we please merge? 🥺

test/test_refine_existing_potential.jl
fhagemann: --> 20644726c98984a129a5010971304bcf19a8223c

PR #476, opened 2025-04-15
Add support for simulation for dead layer-Step2: DiffusionModelTiedWithDrifting

zpdg:
Changes:
1. Comment out the `done[n] = (step_vectors[n] == CartesianVector{T}(0,0,0))` in the `_add_fieldvector_drift!`function to test the diffusion, and for future dead layer simulation
2. Add diffusion model tied with drifting through Einstein's relation
> Method to use this new model: use custom drifting model and define a `calculate_mobility` function

Test:
Construct a z-linear mobility model (z=0-10 mm, mu=0-4.2 m^2/V/s) and kill the drifting. Put charge at r = 5 mm, z = 10-90% of max z (=10mm). The size of the carrier group will be positively correlated with z:
diffusion_DifferentCoef_GroupSigma

Test code:
```julia
using StatsBase
using Plots
using Pkg
Pkg.develop(path="./SolidStateDetectors.jl")
using SolidStateDetectors
using Revise
using Unitful

T = Float32
sim = Simulation{T}(SSD_examples[:TrueCoaxial])
det_z=T(10/1000) # m
det_r=T(10/1000) # m
sim.detector = SolidStateDetector(sim.detector, contact_id = 1, contact_potential = 0)
calculate_electric_field!(sim)


# define the drifting function
# stop the drifting to demonstrate the diffusion 
# use constant zero for drifting, while using z-linear mu for diffusion
using SolidStateDetectors: SSDFloat, AbstractChargeDriftModel
using SolidStateDetectors: Electron, Hole, ChargeCarrier
struct CustomChargeDriftModel{T <: SSDFloat} <: AbstractChargeDriftModel{T} 
    det_z::T
    calculate_mobility::Function
end
using StaticArrays
function SolidStateDetectors.getVe(fv::SVector{3, T}, cdm::CustomChargeDriftModel, current_pos::CartesianPoint{T})::SVector{3, T} where {T <: SSDFloat}
    mu0=0
    return -fv*mu0
end
function SolidStateDetectors.getVh(fv::SVector{3, T}, cdm::CustomChargeDriftModel, current_pos::CartesianPoint{T})::SVector{3, T} where {T <: SSDFloat}
    mu0=0
    return fv*mu0
end
function calculate_mobility(current_pos::CartesianPoint{T}, CC::Type{Hole}, det_z::T = det_z)
    current_pos[3]/det_z*4.2 # in m2/V/s
end
function calculate_mobility(current_pos::CartesianPoint{T}, CC::Type{Electron}, det_z::T = det_z)
    current_pos[3]/det_z*4.2 # in m2/V/s
end
charge_drift_model = CustomChargeDriftModel{T}(det_z, calculate_mobility)


sigma_list = []
startz_list = (0.1:0.1:0.9)*det_z
for startz in startz_list
    start= [det_r/2,0,T(startz)]
    nbcc = NBodyChargeCloud(CartesianPoint{T}(start), 100u"keV", 100, radius = T(0))
    evt = Event(nbcc)
    simulate!(evt, sim, diffusion = true, max_nsteps = 100)
    end_position_list=[]
    for path in evt.drift_paths
        push!(end_position_list, path.h_path[end])
    end
    end_position_list = permutedims(reduce(hcat, end_position_list))
    push!(sigma_list, std(end_position_list[:,1]))
end
p=plot(startz_list*1000, sigma_list*1000)
plot!(xlabel = "z / mm", legend=false, ylabel = "\$\\sigma\$ / mm", title="Group size (100 carriers), after 100 ns", size = (700,500))
```

### DIFF
--- a/src/ChargeDrift/ChargeDrift.jl
+++ b/src/ChargeDrift/ChargeDrift.jl
@@ -99,7 +99,7 @@ function _set_to_zero_vector!(v::Vector{CartesianVector{T}})::Nothing where {T <
 end
 
 function _add_fieldvector_drift!(step_vectors::Vector{CartesianVector{T}}, current_pos::Vector{CartesianPoint{T}}, done::Vector{Bool}, 
-    electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, det::SolidStateDetector{T}, ::Type{S}, end_drift_when_no_field::Bool)::Nothing where {T, S}
+    electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, det::SolidStateDetector{T}, ::Type{S}, end_drift_when_no_field::Bool = true)::Nothing where {T, S}
     for n in eachindex(step_vectors)
        if !done[n]
            step_vectors[n] += get_velocity_vector(electric_field, _convert_point(current_pos[n], S))
--- a/src/ChargeDrift/ChargeDrift.jl
+++ b/src/ChargeDrift/ChargeDrift.jl
@@ -36,7 +36,7 @@ end
 function _drift_charges(det::SolidStateDetector{T}, grid::Grid{T, 3}, point_types::PointTypes{T, 3},
                         starting_points::VectorOfArrays{CartesianPoint{T}}, energies::VectorOfArrays{T},
                         electric_field::Interpolations.Extrapolation{<:SVector{3}, 3},
-                        Δt::RQ; max_nsteps::Int = 2000, diffusion::Bool = false, self_repulsion::Bool = false, verbose::Bool = true)::Vector{EHDriftPath{T}} where {T <: SSDFloat, RQ <: RealQuantity}
+                        Δt::RQ; max_nsteps::Int = 2000, diffusion::Bool = false, self_repulsion::Bool = false, verbose::Bool = true, killdrift_while_nofield::Bool = true)::Vector{EHDriftPath{T}} where {T <: SSDFloat, RQ <: RealQuantity}
 
     drift_paths::Vector{EHDriftPath{T}} = Vector{EHDriftPath{T}}(undef, length(flatview(starting_points)))
     dt::T = T(to_internal_units(Δt))
@@ -53,8 +53,8 @@ function _drift_charges(det::SolidStateDetector{T}, grid::Grid{T, 3}, point_type
         timestamps_e::Vector{T} = Vector{T}(undef, max_nsteps)
         timestamps_h::Vector{T} = Vector{T}(undef, max_nsteps)
         
-        n_e::Int = _drift_charge!( drift_path_e, timestamps_e, det, point_types, grid, start_points, -charges, dt, electric_field, Electron, diffusion = diffusion, self_repulsion = self_repulsion, verbose = verbose )
-        n_h::Int = _drift_charge!( drift_path_h, timestamps_h, det, point_types, grid, start_points,  charges, dt, electric_field, Hole, diffusion = diffusion, self_repulsion = self_repulsion, verbose = verbose )
+        n_e::Int = _drift_charge!( drift_path_e, timestamps_e, det, point_types, grid, start_points, -charges, dt, electric_field, Electron, diffusion = diffusion, self_repulsion = self_repulsion, verbose = verbose, killdrift_while_nofield = killdrift_while_nofield )
+        n_h::Int = _drift_charge!( drift_path_h, timestamps_h, det, point_types, grid, start_points,  charges, dt, electric_field, Hole, diffusion = diffusion, self_repulsion = self_repulsion, verbose = verbose, killdrift_while_nofield = killdrift_while_nofield )
     
         for i in eachindex(start_points)
             drift_paths[drift_path_counter + i] = EHDriftPath{T}( drift_path_e[i,1:n_e], drift_path_h[i,1:n_h], timestamps_e[1:n_e], timestamps_h[1:n_h] )
@@ -103,7 +103,16 @@ function _add_fieldvector_drift!(step_vectors::Vector{CartesianVector{T}}, curre
     for n in eachindex(step_vectors)
        if !done[n]
            step_vectors[n] += get_velocity_vector(electric_field, _convert_point(current_pos[n], S))
-        #    done[n] = (step_vectors[n] == CartesianVector{T}(0,0,0))
+           done[n] = (step_vectors[n] == CartesianVector{T}(0,0,0))
+       end
+    end
+    nothing
+end
+function _add_fieldvector_drift_nokill_when_nofield!(step_vectors::Vector{CartesianVector{T}}, current_pos::Vector{CartesianPoint{T}}, done::Vector{Bool}, 
+    electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, det::SolidStateDetector{T}, ::Type{S})::Nothing where {T, S}
+    for n in eachindex(step_vectors)
+       if !done[n]
+           step_vectors[n] += get_velocity_vector(electric_field, _convert_point(current_pos[n], S))
        end
     end
     nothing
@@ -118,26 +127,12 @@ function _add_fieldvector_diffusion!(step_vectors::Vector{CartesianVector{T}}, d
     end
     nothing 
 end
-function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, CC::Type{Electron})::Nothing where {T <: SSDFloat}
+function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, ::Type{CC})::Nothing where {T <: SSDFloat, CC <: ChargeCarrier}
     for n in eachindex(step_vectors)
         if done[n] continue end
 
-        mu=calculate_mobility(current_pos[n], CC) # mobility_mu in m^2/V/s
-        D=mu*crystal_temperature*8.617e-5 # D = mobility_mu * k0*T_Deadlayer/e # D in m^2/s # k0=8.617e-5 eV/K
-        length=sqrt(6*D*Δt)
-
-        sinθ::T, cosθ::T = sincos(acos(T(2*rand() - 1)))
-        sinφ::T, cosφ::T = sincos(T(rand()*2π))
-        step_vectors[n] += CartesianVector{T}( length * cosφ * sinθ, length * sinφ * sinθ, length * cosθ )
-    end
-    nothing 
-end
-function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, CC::Type{Hole})::Nothing where {T <: SSDFloat}
-    for n in eachindex(step_vectors)
-        if done[n] continue end
-
-        mu=calculate_mobility(current_pos[n], CC) # mobility_mu in m^2/V/s
-        D=mu*crystal_temperature*8.617e-5 # D = mobility_mu * k0*T_Deadlayer/e # D in m^2/s # k0=8.617e-5 eV/K
+        mu=calculate_mobility(current_pos[n], CC) # mu in m^2/V/s
+        D=mu*crystal_temperature*kB/elementary_charge # D in m^2/s
         length=sqrt(6*D*Δt)
 
         sinθ::T, cosθ::T = sincos(acos(T(2*rand() - 1)))
@@ -277,7 +272,8 @@ function _drift_charge!(
                             ::Type{CC};
                             diffusion::Bool = false,
                             self_repulsion::Bool = false,
-                            verbose::Bool = true
+                            verbose::Bool = true,
+                            killdrift_while_nofield::Bool = true
                         )::Int where {T <: SSDFloat, S, CC <: ChargeCarrier}
                         
     n_hits::Int, max_nsteps::Int = size(drift_path)
@@ -315,17 +311,7 @@ function _drift_charge!(
     else
         zero(T)
     end
-    crystal_temperature::T = if diffusion
-        if haskey(det.semiconductor.material, :DT)
-            _parse_value(T, det.semiconductor.material.DT, u"K")
-        else 
-            @info "Ignoring diffusion for now"
-            diffusion = false
-            T(77)
-        end
-    else
-        T(77)
-    end
+    crystal_temperature::T = det.semiconductor.temperature
 
     last_real_step_index::Int = 1
     current_pos::Vector{CartesianPoint{T}} = deepcopy(startpos)
@@ -335,12 +321,13 @@ function _drift_charge!(
 
     # if the drifting model has a calculate_mobility function (scalar mobility), then we use the mobility-tied diffusion model
     # TODO: adapt the code to 3D-vector mobility
-    use_mobility_tied_diffusion = :calculate_mobility in propertynames(det.semiconductor.charge_drift_model) 
+    use_mobility_tied_diffusion = hasproperty(det.semiconductor.charge_drift_model, :calculate_mobility)
     
     @inbounds for istep in 2:max_nsteps
         last_real_step_index += 1
         _set_to_zero_vector!(step_vectors)
-        _add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S)
+        killdrift_while_nofield && _add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S)
+        !killdrift_while_nofield && _add_fieldvector_drift_nokill_when_nofield!(step_vectors, current_pos, done, electric_field, det, S)
         self_repulsion && _add_fieldvector_selfrepulsion!(step_vectors, current_pos, done, charges, ϵ_r)
         _get_driftvectors!(step_vectors, done, Δt, det.semiconductor.charge_drift_model, current_pos, CC)
         diffusion && !use_mobility_tied_diffusion && _add_fieldvector_diffusion!(step_vectors, done, diffusion_length)
--- a/src/ChargeDrift/ChargeDrift.jl
+++ b/src/ChargeDrift/ChargeDrift.jl
@@ -36,7 +36,7 @@ end
 function _drift_charges(det::SolidStateDetector{T}, grid::Grid{T, 3}, point_types::PointTypes{T, 3},
                         starting_points::VectorOfArrays{CartesianPoint{T}}, energies::VectorOfArrays{T},
                         electric_field::Interpolations.Extrapolation{<:SVector{3}, 3},
-                        Δt::RQ; max_nsteps::Int = 2000, diffusion::Bool = false, self_repulsion::Bool = false, verbose::Bool = true, killdrift_while_nofield::Bool = true)::Vector{EHDriftPath{T}} where {T <: SSDFloat, RQ <: RealQuantity}
+                        Δt::RQ; max_nsteps::Int = 2000, diffusion::Bool = false, self_repulsion::Bool = false, verbose::Bool = true, end_drift_when_no_field::Bool = true)::Vector{EHDriftPath{T}} where {T <: SSDFloat, RQ <: RealQuantity}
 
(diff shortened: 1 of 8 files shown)


src/ChargeDrift/ChargeDrift.jl
```diff
@@ -287,20 +315,36 @@ function _drift_charge!(
...
+        if haskey(det.semiconductor.material, :DT)
+            _parse_value(T, det.semiconductor.material.DT, u"K")
+        else 
+            @info "Ignoring diffusion for now"
+            diffusion = false
+            T(77)
+        end
+    else
+        T(77)
+    end
```
fhagemann, line 328: There is a field in `det.semiconductor`, which is `temperature`.
I would suggest to not put the `temperature` in the material properties, but to pass it to the `Simulation` via the config file entry `temperature` in `semiconductor`.

The default temperature should then also be chosen there.

```suggestion
    crystal_temperature::T = det.semiconductor.temperature
```

The default for germanium, right now, is `78K`, whereas for everything else it is `293K`:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/32851dcd3d458879dfcebd07c2352dc681dddf06/src/SolidStateDetector/Semiconductor.jl#L76-L82

src/MaterialProperties/MaterialProperties.jl
```diff
@@ -31,7 +31,8 @@ material_properties[:HPGe] = (
     ml = 1.64,
     mt = 0.0819,
     De = 101u"cm^2/s",
-    Dh = 49u"cm^2/s"
+    Dh = 49u"cm^2/s",
+    DT = 78u"K" # Temperature in Einstein's relation for diffusion calculation
```
fhagemann, line 35: As mentioned above, I would remove `DT` that here and pass the temperature via the detector config file.

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -103,7 +103,7 @@ function _add_fieldvector_drift!(step_vectors::Vector{CartesianVector{T}}, curre
     for n in eachindex(step_vectors)
        if !done[n]
            step_vectors[n] += get_velocity_vector(electric_field, _convert_point(current_pos[n], S))
-           done[n] = (step_vectors[n] == CartesianVector{T}(0,0,0))
+        #    done[n] = (step_vectors[n] == CartesianVector{T}(0,0,0))
```
fhagemann, line 106: We introduced this to kill charge drifts in case they reach a region with no electric field. It might make sense to still allow for this using a keyword argument to avoid unnecessary long charge drifts if the charges enter a region with an electric field of exactly zero.

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -118,6 +118,34 @@ function _add_fieldvector_diffusion!(step_vectors::Vector{CartesianVector{T}}, d
     end
     nothing 
 end
+function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, CC::Type{Electron})::Nothing where {T <: SSDFloat}
+    for n in eachindex(step_vectors)
+        if done[n] continue end
+
+        mu=calculate_mobility(current_pos[n], CC) # mobility_mu in m^2/V/s
+        D=mu*crystal_temperature*8.617e-5 # D = mobility_mu * k0*T_Deadlayer/e # D in m^2/s # k0=8.617e-5 eV/K
```
fhagemann, line 126: Can you use the Boltzmann constant defined in `MaterialProperties.jl`?
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/32851dcd3d458879dfcebd07c2352dc681dddf06/src/MaterialProperties/MaterialProperties.jl#L3-L7

You might need to adjust for the units by multiplying/dividing by the `elementary_charge`, which is also defined there:
```suggestion
        D=mu*kB*crystal_temperature/elementary_charge # D = mobility_mu * k0*T_Deadlayer/e # D in m^2/s
```

For reference:
```julia
using SolidStateDetectors
SolidStateDetectors.kB / SolidStateDetectors.elementary_charge
```
```
8.617333262145179e-5
```

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -118,6 +118,34 @@ function _add_fieldvector_diffusion!(step_vectors::Vector{CartesianVector{T}}, d
...
+        step_vectors[n] += CartesianVector{T}( length * cosφ * sinθ, length * sinφ * sinθ, length * cosθ )
+    end
+    nothing 
+end
+function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, CC::Type{Hole})::Nothing where {T <: SSDFloat}
+    for n in eachindex(step_vectors)
+        if done[n] continue end
+
+        mu=calculate_mobility(current_pos[n], CC) # mobility_mu in m^2/V/s
+        D=mu*crystal_temperature*8.617e-5 # D = mobility_mu * k0*T_Deadlayer/e # D in m^2/s # k0=8.617e-5 eV/K
```
fhagemann, line 140: same thing here.

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -118,6 +118,34 @@ function _add_fieldvector_diffusion!(step_vectors::Vector{CartesianVector{T}}, d
...
+        length=sqrt(6*D*Δt)
+
+        sinθ::T, cosθ::T = sincos(acos(T(2*rand() - 1)))
+        sinφ::T, cosφ::T = sincos(T(rand()*2π))
+        step_vectors[n] += CartesianVector{T}( length * cosφ * sinθ, length * sinφ * sinθ, length * cosθ )
+    end
+    nothing 
+end
+function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, CC::Type{Hole})::Nothing where {T <: SSDFloat}
```
fhagemann, line 135: I don't see an obvious difference between the methods for `Electrons` and `Holes`, just that you pass `CC` further to `calculate_mobility`. Can you combined the two methods into one?

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -287,20 +315,36 @@ function _drift_charge!(
...
 
     last_real_step_index::Int = 1
     current_pos::Vector{CartesianPoint{T}} = deepcopy(startpos)
     step_vectors::Vector{CartesianVector{T}} = Vector{CartesianVector{T}}(undef, n_hits)
     done::Vector{Bool} = broadcast(pt -> !_is_next_point_in_det(pt, det, point_types), startpos)
     normal::Vector{Bool} = deepcopy(done)
+
+    # if the drifting model has a calculate_mobility function (scalar mobility), then we use the mobility-tied diffusion model
+    # TODO: adapt the code to 3D-vector mobility
+    use_mobility_tied_diffusion = :calculate_mobility in propertynames(det.semiconductor.charge_drift_model)
```
fhagemann, line 338:
```suggestion
    use_mobility_tied_diffusion = hasproperty(det.semiconductor.charge_drift_model, :calculate_mobility) 

```

fhagemann (changes requested):

fhagemann:
I rebased your branch onto the current `JuliaPhysics#transition`, you might wanna do that as well before incorporating changes.

fhagemann:
And it would also be great to add tests for the current methods.
Do you need some guidance on this?

src/ChargeDrift/ChargeDrift.jl
zpdg: I agree. Actually I have considered about this. I followed the practice in the temperature setting of the Boggs trapping model. 

So, I think we should unify all the temperatures in SSD.jl by reading from the config entry: `det.semiconductor.temperature`

src/ChargeDrift/ChargeDrift.jl
zpdg: Yes,  but if there is a zero-field area, the long charge drifts is  actually a fact. Besides, the long charge drifts could be limited by the `max_steps` manually. 

I think we could introduce this termination only when we don't simulate diffusion by adding an if-else.

src/ChargeDrift/ChargeDrift.jl
zpdg: Sure!

src/ChargeDrift/ChargeDrift.jl
zpdg: Seems my understanding of the code is wrong. I thought I have to do this.
Yes, I will try.

zpdg:
> And it would also be great to add tests for the current methods. Do you need some guidance on this?

Yes, that would be nice. I don't know the rule of setting a test.

src/ChargeDrift/ChargeDrift.jl
zpdg: Ah, I see. I totally agree.

src/ChargeDrift/ChargeDrift.jl
fhagemann: So, if both methods agree except for what you expect `CC` to be (but that is passed on further), so could combine both methods into one using:
```suggestion
function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, ::Type{CC})::Nothing where {T <: SSDFloat, CC <: ChargeCarrier}
```
see L. 249 and L.253 in https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/32851dcd3d458879dfcebd07c2352dc681dddf06/src/ChargeDrift/ChargeDrift.jl#L239-L253

fhagemann: But check that your code still runs afterwards

fhagemann:
> > And it would also be great to add tests for the current methods. Do you need some guidance on this?
> 
> Yes, that would be nice. I don't know the rule of setting a test.

Basically, you want to add some code that runs the newly added functions and methods to see whether they perform as expected. So that, in the future, whenever someone changes your code, it still produces the expected results.

An example `@testset` using the julia package `Test` could look like this:
```julia
using SolidStateDetectors
using SolidStateDetectors: Electron, Hole
using LinearAlgebra
using Test
T = Float32

@testset "Test position-dependent diffusion" begin
    
    N = 10
    done = fill(false, N)
    Δt = T(1e-9)
    zs = 1:10
    current_pos = CartesianPoint{T}.(0, 0, zs)
    crystal_temperature = T(78)
    
    step_vectors = zeros(CartesianVector{T}, N)
    @test all(iszero.(step_vectors))

    # if calculate_mobility gives a zero mobility, nothing should happen
    calculate_mobility(pt::CartesianPoint{T}, ::Type{Hole}) where {T} = zero(T)
    SolidStateDetectors._add_fieldvector_diffusion_TiedWithMobility!(step_vectors, done, Δt, calculate_mobility, current_pos, crystal_temperature, Hole)
    @test all(iszero.(step_vectors))  
    calculate_mobility(pt::CartesianPoint{T}, ::Type{Electron}) where {T} = zero(T)
    SolidStateDetectors._add_fieldvector_diffusion_TiedWithMobility!(step_vectors, done, Δt, calculate_mobility, current_pos, crystal_temperature, Electron)
    @test all(iszero.(step_vectors))
    
    # if the mobility is set to some position-independent value, then we know what the length of the expected step_vectors should be
    μ0 = T(5.0 + rand() * 5.0)
    calculate_mobility_constant(pt::CartesianPoint{T}, ::Type{Hole}) where {T} = μ0
    SolidStateDetectors._add_fieldvector_diffusion_TiedWithMobility!(step_vectors, done, Δt, calculate_mobility_constant, current_pos, crystal_temperature, Hole)
    D0 = μ0 * crystal_temperature * SolidStateDetectors.kB / SolidStateDetectors.elementary_charge
    @test all(isapprox.(norm.(step_vectors), sqrt(6 * D0 * Δt), rtol = 0.01))
    step_vectors = zeros(CartesianVector{T}, N) # reset step vectors
    
    # use a position-dependent mobility
    calculate_mobility_pos(pt::CartesianPoint{T}, ::Type{Hole}) where {T} = pt.z
    SolidStateDetectors._add_fieldvector_diffusion_TiedWithMobility!(step_vectors, done, Δt, calculate_mobility_pos, current_pos, crystal_temperature, Hole)
    @test all(isapprox.(norm.(step_vectors).^2 ./ (6 * Δt) * SolidStateDetectors.elementary_charge / (crystal_temperature * SolidStateDetectors.kB), zs, rtol = 0.01))
end
```

Ideally, everything you added should somewhat be covered by tests and added to the test code in `test`.

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -335,12 +321,13 @@ function _drift_charge!(
...
     # TODO: adapt the code to 3D-vector mobility
-    use_mobility_tied_diffusion = :calculate_mobility in propertynames(det.semiconductor.charge_drift_model) 
+    use_mobility_tied_diffusion = hasproperty(det.semiconductor.charge_drift_model, :calculate_mobility)
     
     @inbounds for istep in 2:max_nsteps
         last_real_step_index += 1
         _set_to_zero_vector!(step_vectors)
-        _add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S)
+        killdrift_while_nofield && _add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S)
+        !killdrift_while_nofield && _add_fieldvector_drift_nokill_when_nofield!(step_vectors, current_pos, done, electric_field, det, S)
```
fhagemann, line 330: Then, you could just pass this keyword argument to the function, e.g. using
```suggestion
_add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S, killdrift_while_nofield)
```

fhagemann (commented):
Julia has this nice thing where you can have several methods for a function with the same name, depending on the keyword arguments you pass.
You could modify the function `_add_fieldvector_drift!` without the need to define a completely new method.

fhagemann:
Maybe also rename the keyword argument.
Something like `end_drift_when_no_field` or `terminate_drift_when_no_field` instead of `kill`.

src/ChargeDrift/ChargeDrift.jl
zpdg: I've thought this. This way is simpler, but N times slower then defining a new function (N is the carrier number). 

But I guess the if-else takes almost no time. I will change this.

zpdg:
> Maybe also rename the keyword argument. Something like `end_drift_when_no_field` or `terminate_drift_when_no_field` instead of `kill`.

Sure!

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -99,20 +99,11 @@ function _set_to_zero_vector!(v::Vector{CartesianVector{T}})::Nothing where {T <
...
-       if !done[n]
-           step_vectors[n] += get_velocity_vector(electric_field, _convert_point(current_pos[n], S))
-           done[n] = (step_vectors[n] == CartesianVector{T}(0,0,0))
-       end
-    end
-    nothing
-end
-function _add_fieldvector_drift_nokill_when_nofield!(step_vectors::Vector{CartesianVector{T}}, current_pos::Vector{CartesianPoint{T}}, done::Vector{Bool}, 
-    electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, det::SolidStateDetector{T}, ::Type{S})::Nothing where {T, S}
+    electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, det::SolidStateDetector{T}, ::Type{S}, end_drift_when_no_field::Bool)::Nothing where {T, S}
```
fhagemann, line 102: If you give a default value for `end_drift_when_no_field` here (`true`), the new code would still behave like the old code when explicitly calling `_add_fieldvector_drift!` and would, thus, be non-breaking
```suggestion
    electric_field::Interpolations.Extrapolation{<:StaticVector{3}, 3}, det::SolidStateDetector{T}, ::Type{S}, end_drift_when_no_field::Bool = true)::Nothing where {T, S}
```

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -287,20 +302,26 @@ function _drift_charge!(
...
     @inbounds for istep in 2:max_nsteps
         last_real_step_index += 1
         _set_to_zero_vector!(step_vectors)
-        _add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S)
+        _add_fieldvector_drift!(step_vectors, current_pos, done, electric_field, det, S, end_drift_when_no_field)
         self_repulsion && _add_fieldvector_selfrepulsion!(step_vectors, current_pos, done, charges, ϵ_r)
         _get_driftvectors!(step_vectors, done, Δt, det.semiconductor.charge_drift_model, current_pos, CC)
-        diffusion && _add_fieldvector_diffusion!(step_vectors, done, diffusion_length)
+        diffusion && !use_mobility_tied_diffusion && _add_fieldvector_diffusion!(step_vectors, done, diffusion_length)
+        diffusion && use_mobility_tied_diffusion && _add_fieldvector_diffusion_TiedWithMobility!(step_vectors, done, Δt, det.semiconductor.charge_drift_model.calculate_mobility, current_pos, crystal_temperature, CC)
```
fhagemann, line 324: can you also combine these methods into one?
One could think about making `DiffusionModel` a struct that can become part of the `Semiconductor` or `ChargeDriftModel`.

src/ChargeDrift/ChargeDrift.jl
zpdg: Yes, the default behavior doesn't change. When we want to calculate pulses in the dead layer, we need to set `end_drift_when_no_field` to `false`.

fhagemann: Right now, the function is modified such that it needs an additional argument `end_drift_when_no_field`.
You might wanna add that its default value is `true` here, so that people can still use their old code.

src/ChargeDrift/ChargeDrift.jl
zpdg: I have also thought about making a new struct for diffusion. But I am not sure about the goal. This depends on how we see "diffusion". Is it very clear physics process we should standardize? Or we provide options?

I would see "diffusion" and the Einstein's relation is quite clear. Maybe we should just standardize it and offer only one method `_add_fieldvector_diffusion_TiedWithMobility!`. This way we don't really need a struct. (We could upgrade it to 3D-vector diffusion length later)

Otherwise, we would need a struct, maybe it's better to build a new `ChargeDiffusionModel`

src/ChargeDrift/ChargeDrift.jl
zpdg: Ah, you mean the direct call of `_add_fieldvector_drift!`. Sure!

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -118,6 +118,20 @@ function _add_fieldvector_diffusion!(step_vectors::Vector{CartesianVector{T}}, d
     end
     nothing 
 end
+function _add_fieldvector_diffusion_TiedWithMobility!(step_vectors::Vector{CartesianVector{T}}, done::Vector{Bool}, Δt::T, calculate_mobility::Function, current_pos::Vector{CartesianPoint{T}}, crystal_temperature::T, ::Type{CC})::Nothing where {T <: SSDFloat, CC <: ChargeCarrier}
```
fhagemann, line 121: This is a very unusual naming (mixing camel case and snake case).
Can you just call this `_add_fieldvector_diffusion!` as well?
Julia will know which method to call by the arguments passed.

src/ChargeDrift/ChargeDrift.jl
fhagemann: That would also be an idea, but in general, the naming `_add_fieldvector_diffusion_TiedWithMobility!` is a bit weird...
Can you remove `_TiedWithMobility` from the name and then do something using multiple dispatch, for example:
```suggestion
        diffusion && (use_mobility_tied_diffusion ? _add_fieldvector_diffusion_TiedWithMobility!(step_vectors, done, Δt, det.semiconductor.charge_drift_model.calculate_mobility, current_pos, crystal_temperature, CC)  : _add_fieldvector_diffusion!(step_vectors, done, diffusion_length))
```

src/ChargeDrift/ChargeDrift.jl
zpdg: Sure!

fhagemann:
@zpdg I again squashed all commits into one, please make sure to rebase before continuing adding stuff

PR #473, opened 2025-04-02
Boule impurity models

hervasa2:
Added Impurity models `LinBouleImpurityDensity` and `LinExpBouleImpurityDensity` for cases when user whats to define an impurity model for the boule and then place the detector in a certain location `det_z0`. These are placed in a separate folder to distinguish from those that use models based on detector coordinates.
See documentation:
```
"""
Boule impurity densities are meant to be used when the impurity density is defined in the boule coordinates, where the z-axis is aligned with the boule growth direction. 

Different models are provided. In each the field `det_z0` is the z-coordinate of the detector origin in boule coordinates. The z-direction of the detector is opposite to the z-direction of the boule coordinates.

In this matter the detector impurities are automatically determined from those of the boule, depending on `det_z0`.
"""
```
```julia
det_z0 = 0.12
boule_ρ0 = -1e16
idm = LinBouleImpurityDensity{T}(boule_ρ0, -1e17, det_z0)

sim = Simulation{T}(SSD_examples[:InvertedCoax])
sim.detector = SolidStateDetector(sim.detector, idm)

points = broadcast(z -> SolidStateDetectors.CylindricalPoint{T}(0,0,z/1000), 0:80)
ρ = broadcast(p -> SolidStateDetectors.get_impurity_density(sim.detector.semiconductor.impurity_density_model, p), points);

plot(getproperty.(points, :z)*u"m", ρ*u"m^-3", xunit = u"mm", yunit = u"cm^-3", legend = false)
```
Screenshot 2025-04-02 at 14 34 58

PR #477, opened 2025-04-17
refactor boule impurity densities

hervasa2:
A a follow up to #473 additional `BouleImpurityDensity` models were added. To ensure consistency of the parameter names, these were slightly modified since last PR. Additionally, now these models can be included in the config files. Additional runtests were added.

### DIFF
--- a/src/ImpurityDensities/BouleImpurityDensities/LinExpBouleImpurityDensity.jl
+++ b/src/ImpurityDensities/BouleImpurityDensities/LinExpBouleImpurityDensity.jl
@@ -27,7 +27,7 @@ function ImpurityDensity(T::DataType, t::Val{:linear_exponential_boule}, dict::A
     b::T = haskey(dict, "b") ? _parse_value(T, dict["b"], input_units.length^(-4)) : T(0)
     n::T = haskey(dict, "n") ? _parse_value(T, dict["n"], input_units.length^(-3)) : T(0)
     l::T = haskey(dict, "l") ? _parse_value(T, dict["l"], input_units.length) : T(0)
-    m::T = haskey(dict, "m") ? _parse_value(T, dict["m"], input_units.length) : T(0)
+    m::T = haskey(dict, "m") ? _parse_value(T, dict["m"], input_units.length) : T(1)
     det_z0::T = haskey(dict, "det_z0") ? _parse_value(T, dict["det_z0"], input_units.length) : T(0)
     LinExpBouleImpurityDensity{T}(a, b, n, l, m, det_z0)
 end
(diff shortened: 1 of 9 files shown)


hervasa2:
I see. Then the idea would be to keep `BouleImpurityDensities` but remove `det_z0` in the future.  `det_z0` and any rotation would then be parsed from the config files to create a transformed `BouleImpurityDensity`

src/ImpurityDensities/BouleImpurityDensities/LinExpBouleImpurityDensity.jl
```diff
@@ -1,29 +1,39 @@
...
+    m::T
     det_z0::T
 end
 
+function ImpurityDensity(T::DataType, t::Val{:linear_exponential_boule}, dict::AbstractDict, input_units::NamedTuple)
+    a::T = haskey(dict, "a") ? _parse_value(T, dict["a"], input_units.length^(-3)) : T(0)
+    b::T = haskey(dict, "b") ? _parse_value(T, dict["b"], input_units.length^(-4)) : T(0)
+    n::T = haskey(dict, "n") ? _parse_value(T, dict["n"], input_units.length^(-3)) : T(0)
+    l::T = haskey(dict, "l") ? _parse_value(T, dict["l"], input_units.length) : T(0)
+    m::T = haskey(dict, "m") ? _parse_value(T, dict["m"], input_units.length) : T(0)
```
fhagemann, line 30: I am not sure whether you want the default value for `m` to be `T(0)` because, you will be dividing by zero then and always get `NaN`

src/ImpurityDensities/BouleImpurityDensities/ParExpBouleImpurityDensity.jl
```diff
@@ -0,0 +1,42 @@
...
+    det_z0::T
+end
+
+function ImpurityDensity(T::DataType, t::Val{:parabolic_exponential_boule}, dict::AbstractDict, input_units::NamedTuple)
+    a::T = haskey(dict, "a") ? _parse_value(T, dict["a"], input_units.length^(-3)) : T(0)
+    b::T = haskey(dict, "b") ? _parse_value(T, dict["b"], input_units.length^(-4)) : T(0)
+    c::T = haskey(dict, "c") ? _parse_value(T, dict["c"], input_units.length^(-5)) : T(0)
+    n::T = haskey(dict, "n") ? _parse_value(T, dict["n"], input_units.length^(-3)) : T(0)
+    l::T = haskey(dict, "l") ? _parse_value(T, dict["l"], input_units.length) : T(0)
+    m::T = haskey(dict, "m") ? _parse_value(T, dict["m"], input_units.length) : T(0)
```
fhagemann, line 33: Same thing here.

PR #484, opened 2025-05-09
Add contractor method and scale function to BouleImpurityDensity

hervasa2:
User can now construct BouleImpurityDensities from a unitfull vector and value like so:
```julia
idm = LinBouleImpurityDensity{T}(pars::Vector{<:Number}, det_z0::Number)
````
Then units are automatically converted to SSD internal units. 
Scaling of BouleImpurityDensities is also now possible with:

```julia
scale*idm
```
Which will scale the correct parameters in the `BouleImpurityDensity` such the numerical value calculated by get_impurity_density would be scaled accordingly.

I also added a tolerance for the warning when using depletion handling:
`""Detector seems not to be not fully depleted at a bias voltage of""`...
This warning appears almost every time and should be dropped unless it falls out of the given tolerance

### DIFF
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -992,17 +992,18 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         end
     end
     if verbose && depletion_handling && isEP
+        depletion_tol = 0.1
         maximum_applied_potential = maximum(broadcast(c -> c.potential, sim.detector.contacts))
         minimum_applied_potential = minimum(broadcast(c -> c.potential, sim.detector.contacts))
         @inbounds for i in eachindex(sim.electric_potential.data)
-            if sim.electric_potential.data[i] < minimum_applied_potential # p-type
+            if sim.electric_potential.data[i] < minimum_applied_potential - depletion_tol# p-type
                 @warn """Detector seems not to be fully depleted at a bias voltage of $(bias_voltage) V.
                     At least one grid point has a smaller potential value ($(sim.electric_potential.data[i]) V)
                     than the minimum applied potential ($(minimum_applied_potential) V). This should not be.
                     However, small overshoots could be due to numerical precision."""
                 break
             end
-            if sim.electric_potential.data[i] > maximum_applied_potential # n-type
+            if sim.electric_potential.data[i] > maximum_applied_potential + depletion_tol # n-type
                 @warn """Detector seems not to be not fully depleted at a bias voltage of $(bias_voltage) V.
                     At least one grid point has a higher potential value ($(sim.electric_potential.data[i]) V)
                     than the maximum applied potential ($(maximum_applied_potential) V). This should not be.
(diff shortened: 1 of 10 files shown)


src/Simulation/Simulation.jl
```diff
@@ -992,17 +992,18 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         end
     end
     if verbose && depletion_handling && isEP
+        depletion_tol = 0.1
```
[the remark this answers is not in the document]
hervasa2: I think the latter would require quite some investigation. I added it as a keyword for now.

src/Simulation/Simulation.jl
```diff
@@ -820,6 +820,7 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         max_tick_distance::Union{Missing, LengthQuantity, Tuple{LengthQuantity, <:Union{LengthQuantity, AngleQuantity}, LengthQuantity}} = missing,
         max_distance_ratio::Real = 5,
         depletion_handling::Bool = false,
+        depletion_tol = 0.1,
```
fhagemann, line 823: Maybe final comment: can you allow for this to also have units, so that you can pass `depletion_tol = 0.1u“V“`, using to_internal_units in the function below?
And add a short description to the docstring what it does.

src/Simulation/Simulation.jl
```diff
@@ -995,14 +996,14 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         maximum_applied_potential = maximum(broadcast(c -> c.potential, sim.detector.contacts))
         minimum_applied_potential = minimum(broadcast(c -> c.potential, sim.detector.contacts))
         @inbounds for i in eachindex(sim.electric_potential.data)
-            if sim.electric_potential.data[i] < minimum_applied_potential # p-type
+            if sim.electric_potential.data[i] < minimum_applied_potential - depletion_tol# p-type
```
fhagemann, line 999: I think you can use something like `to_internal_units` or `_parse_value` here.
(To internal units probably only makes sense if you force the type of `depletion_tol` to be eather a `Real` or a `Quantity` of dimension voltage)

fhagemann: We could do both: add a depletion_tol and adding a keyword to avoid this check

fhagemann: This check was initially implemented to check for obvious overshoots.

If the detector is not fully depleted but the potential calculation is run with `depletion_handling = false`, this warning should be more useful. However, it looks like we're only checking for this with `depletion_handling = true`.

I just ran an example where we expect the detector to NOT be depleted and where such a warning might be useful:
```julia
using SolidStateDetectors

# create inverted coax with 0V at point contact and 500V at outer contact
sim.detector = SolidStateDetector(sim.detector, contact_id = 2, contact_potential = 500)

# calculate electric potential with explicit depletion_handling = false
calculate_electric_potential!(sim, depletion_handling = false)

# check for extreme values in the electric potential
extrema(sim.electric_potential.data)
```
```
(-1192.5739f0, 500.00302f0)
```

fhagemann:
To speed this PR up: can you split this into two PRs? This one JUST having the scaling of `BouleImpurityDensity` and the other one dealing with the `depletion_handling` and overshoot warning?

These two are code-wise unrelated.

fhagemann (approved):
Are you planning on adding tests on this?

hervasa2:
I don't think we need to add tests or documentations just yet. I would like for it to be fully stable first

hervasa2:
Yes, I already removed the changes to `_calculate_potential!`. This is PR is thus now only  about `BouleImpurityDensity` constructors and `scale methods`

PR #486, opened 2025-05-12
fix for wrong radius charge cloud

### LINKED ISSUE #485: Possible issue in charge cloud radius
After switching from v.0.10.10 to v0.10.14 I noticed strange behaviour in the waveforms, as if the charges are getting moved very far.. This simulation was for 2 MeV electrons in the bulk

Image

I also get an almost constant print out of this warning:
```julia
Warning: 6 charges of the charge cloud at [-0.016007724609, -0.001966646963, 0.011837773379] are outside. Moving them inside...
│ The new charge center is at [-0.014480657772, -0.004455300255, 0.015663129164].
└ @ SolidStateDetectors [~/.julia/packages/SolidStateDetectors/6t4KL/src/Event/Event.jl:165](http://localhost:8888/home/tdixon/.julia/packages/SolidStateDetectors/6t4KL/src/Event/Event.jl#line=164)
```

I noticed the following line was added in [[link](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blame/abe40bd73c6883e0903955a69a80e33d796d0a4a/src/MCEventsProcessing/MCEventsProcessing.jl#L132) passing the charge cloud radius.

This confuses me since it seems the energy is passed not the radius guess...

tdixon97:
This seems to fix #485 , although there are still a lot of warnings about charges outside. But this is likely just due to some events being close to the surface.
For example this is now a waveform with diffusion and varying `n`
image

### DIFF
--- a/src/MCEventsProcessing/MCEventsProcessing.jl
+++ b/src/MCEventsProcessing/MCEventsProcessing.jl
@@ -129,7 +129,7 @@ function _convertEnergyDepsToChargeDeps(
                     T.(to_internal_units(edep[iEdep_indep][i_together])),
                     number_of_carriers,
                     number_of_shells = number_of_shells,
-                    radius = T(to_internal_units(edep[iEdep_indep][i_together]))
+                    radius = T(to_internal_units(radius[iEdep_indep][i_together]))
                 )
                 move_charges_inside_semiconductor!([nbcc.locations], [nbcc.energies], det)
                 nbcc


fhagemann:
Thanks!

fhagemann:
Im releasing a new version 0.10.15 which should be out soon
https://github.com/JuliaRegistries/General/pull/130828

PR #493, opened 2025-05-24
Add support for simulation for dead layer-Step4: impurity model based on the lithium thermal diffusion process

zpdg:
A PN junction impurity model based on lithium thermal diffusion and constant bulk impurity density.

Test:
Use the example yaml-TrueCoaxial and apply this impurity model:
image

Code for test:
```julia
using Revise
using Pkg
using Plots
using Unitful
Pkg.develop(path = "./SolidStateDetectors.jl")
using SolidStateDetectors

T = Float64
mm = T(1 / 1000)
det_z = det_r = 10 * mm
pn_r = 8.957282 * mm
sim = Simulation{T}(SSD_examples[:TrueCoaxial])

calculate_electric_potential!(sim, max_n_iterations = 10, grid = Grid(sim), verbose = false, depletion_handling = true)
ax1, ax2, ax3 = g.axes
bulk_tick_dis = 0.05 * mm
dl_tick_dis = 0.01 * mm
user_additional_ticks_ax1 = sort(vcat(ax1.interval.left:bulk_tick_dis:pn_r, pn_r:dl_tick_dis:ax1.interval.right))
user_g = typeof(g)((user_ax1, ax2, ax3))
calculate_electric_potential!(sim, refinement_limits = 0.1, use_nthreads = 8, grid = user_g, depletion_handling = true)

z_draw = det_z / 2
plot(sim.imp_scale, y = z_draw, xunit = u"mm", yunit = u"mm", title = "")
vline!([pn_r / mm], lw = 2, ls = :dash, color = :darkred, label = "PN junction boundary")
```

### DIFF
--- a/docs/src/man/electric_potential.md
+++ b/docs/src/man/electric_potential.md
@@ -105,6 +105,25 @@ impurity_density:
 Here, the impurity density at the origin is $10^{10}$cm$^{-3}$ and it increases radially with the gradient $10^{10}$cm$^{-4}$.
 If no units are given, `init` is parsed in units of `units.length`$^{-3}$ and `gradient` in units of `units.length`$^{-4}$.
 
+### P-type PN junction Impurity Density
+A PN junctin impurity model based on lithium thermal diffusion and constant bulk impurity density. The surface lithium density is saturated.
+ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
+
+This density model could be defined in the code: 
+```julia
+T = Float64
+sim = Simulation{T}(SSD_examples[:TrueCoaxial])
+
+calculate_depth2surface = pt -> 0.01-hypot(pt[1], pt[2])
+Li_annealing_temperature::T=623
+Li_annealing_time::T=18*60
+Bulk_Impurity = p_type_density = -1e10 * 1e6
+
+pn_junction_impurity_density = PtypePNJunctionImpurityDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface, p_type_density)
+sim.detector = SolidStateDetector(sim.detector, pn_junction_impurity_density)
+```
+
+This model couldn't be defined in the configuration because a depth2surface function is needed, except when using the public_TrueCoaxial.yaml.
 
 ### Custom Impurity Density
 
--- a/examples/example_config_files/public_TrueCoaxial.yaml
+++ b/examples/example_config_files/public_TrueCoaxial.yaml
@@ -26,10 +26,12 @@ medium: vacuum
 detectors:
 - semiconductor:
     material: HPGe
-    temperature: 77
+    temperature: 90
     impurity_density:
-      name: constant
-      value: 1e8cm^-3
+      name: PtypePNjunction
+      Li_annealing_temperature: 623
+      Li_annealing_time: 1080
+      bulk_imp: -1e10cm^-3
     charge_drift_model:
       include: ADLChargeDriftModel/drift_velocity_config.yaml
     charge_trapping_model:
@@ -37,12 +39,12 @@ detectors:
       parameters:
         τh: 1ms
         τe: 1ms
-        τh_inactive: 80ns
-        τe_inactive: 80ns
+        τh_inactive: 1μs
+        τe_inactive: 1μs
       inactive_layer_geometry:
         tube:
           r:
-            from: 9.0
+            from: 8.957282
             to: 10.0
           h: 10.0
           origin:
@@ -56,19 +58,19 @@ detectors:
         origin:
           z: 5.0
   contacts:
-    - name: "P+"
+    - name: "P⁺"
       id: 1
       material: HPGe
       potential: -500
       geometry:
         tube:
           r:
-            from: 0.5
+            from: 1.0
             to: 1.0
           h: 10.0
           origin:
             z: 5.0
-    - name: "N+"
+    - name: "N⁺"
       id: 2
       material: HPGe
       potential: 0
--- a/src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
+++ b/src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
@@ -1,0 +1,59 @@
+"""
+    struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+
+A PN junctin impurity model based on lithium thermal diffusion and constant bulk impurity density.
+The surface lithium density is saturated.
+
+ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
+ 
+## Fields
+* `Li_annealing_temperature::T`: lithium annealing temperature when the lithium is diffused into the crystal. The default value is 623 K.
+* `Li_annealing_time::T`: lithium annealing time. The default value is 18 minutes.
+* `calculate_depth2surface::Function`: the function for describing the depth to surface.
+* `bulk_imp::T`: the bulk impurity density constant.
+* `bulk/surface_imp_model::T`: there two variables will be constructed with above parameters.
+"""
+
+struct PtypePNJunctionImpurityDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+	Li_annealing_temperature::T
+	Li_annealing_time::T
+    calculate_depth2surface::Function
+    bulk_imp::T
+    surface_imp_model::ThermalDiffusionLithiumDensity{T}
+    bulk_imp_model::ConstantImpurityDensity{T}
+end
+function PtypePNJunctionImpurityDensity{T}(Li_annealing_temperature::T,
+    Li_annealing_time::T,
+    calculate_depth2surface::Function,
+    bulk_imp::T
+    ) where {T <: SSDFloat}
+    bulk_imp_model = ConstantImpurityDensity{T}(bulk_imp)
+    surface_imp_model = ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface)
+    PtypePNJunctionImpurityDensity(Li_annealing_temperature,Li_annealing_time,calculate_depth2surface, bulk_imp, surface_imp_model, bulk_imp_model)
+end
+
+function ImpurityDensity(T::DataType, t::Val{:PtypePNjunction}, dict::AbstractDict, input_units::NamedTuple)
+    Li_annealing_temperature = T(623)
+    Li_annealing_time= T(18*60)
+    calculate_depth2surface = pt->0.01-hypot(pt[1],pt[2]) # default for the public_TrueCoaxial.yaml
+    bulk_imp = T(-1e16)
+
+    if haskey(dict, "Li_annealing_temperature")
+        Li_annealing_temperature = _parse_value(T, dict["Li_annealing_temperature"], u"K")
+    end
+    if haskey(dict, "Li_annealing_time")
+        Li_annealing_time = _parse_value(T, dict["Li_annealing_time"], u"s")
+    end
+    if haskey(dict, "calculate_depth2surface")
+        calculate_depth2surface = dict["calculate_depth2surface"]
+    end
+    if haskey(dict, "bulk_imp")
+        bulk_imp = _parse_value(T, dict["bulk_imp"], input_units.length^(-3))
+    end
+
+    PtypePNJunctionImpurityDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface, bulk_imp)
+end
+
+function get_impurity_density(PtypePNjunction::PtypePNJunctionImpurityDensity{T}, pt::AbstractCoordinatePoint{T})::T where {T <: SSDFloat}
+    get_impurity_density(PtypePNjunction.bulk_imp_model, pt)+get_impurity_density(PtypePNjunction.surface_imp_model, pt)
+end
--- a/src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
+++ b/src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
@@ -1,0 +1,68 @@
+"""
+    struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+
+Lithium impurity density model
+
+ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
+ 
+## Fields
+* `Li_annealing_temperature::T`: lithium annealing temperature when the lithium is diffused into the crystal. The default value is 623 K.
+* `Li_annealing_time::T`: lithium annealing time. The default value is 18 minutes.
+* `calculate_depth2surface::Function`: the function for describing the depth to surface
+* `Li_surface::T`: the lithium concentration in the surface
+* `D_Li::T`: the diffusivity of lithium in germanium
+"""
+
+struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+	Li_annealing_temperature::T # in K
+	Li_annealing_time::T # in s
+    calculate_depth2surface::Function # pt -> depth (depth in m)
+	Li_surface::T # in m^-3
+    D_Li::T # in m^2*s^-1
+end
+function calculate_D_Li(Li_annealing_temperature::T)::T where {T <: SSDFloat}
+    if Li_annealing_temperature<=873
+        D0 = 2.5e-3*1e-4 # m^2*s^-1
+        H = 11800 # cal
+    else
+        D0 = 1.3e-3*1e-4
+        H = 10700
+    end
+    D_Li = D0 * exp(-H/(R_gas*Li_annealing_temperature))
+    D_Li
+end
+function calculate_Li_surface_saturated_density(Li_annealing_temperature::T)::T where {T <: SSDFloat}
+    10^(21.27 - 2610.0/(Li_annealing_temperature)) * 1e6 # m^-3
+end
+function ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature::T,
+    Li_annealing_time::T,
+    calculate_depth2surface::Function,
+    Li_surface::T=calculate_Li_surface_saturated_density(Li_annealing_temperature),
+    D_Li::T=calculate_D_Li(Li_annealing_temperature)
+    ) where {T <: SSDFloat}
+    ThermalDiffusionLithiumDensity(Li_annealing_temperature,Li_annealing_time,calculate_depth2surface, Li_surface, D_Li)
+end
+
+function ImpurityDensity(T::DataType, t::Val{:tdld}, dict::AbstractDict, input_units::NamedTuple)
+    Li_annealing_temperature = T(623)
+    Li_annealing_time= T(18*60)
+    calculate_depth2surface = pt->0.001
+
+    if haskey(dict, "Li_annealing_temperature")
+        Li_annealing_temperature = _parse_value(T, dict["Li_annealing_temperature"], u"K")
+    end
+    if haskey(dict, "Li_annealing_time")
+        Li_annealing_time = _parse_value(T, dict["Li_annealing_time"], u"s")
+    end
+    if haskey(dict, "calculate_depth2surface")
+        calculate_depth2surface = dict["calculate_depth2surface"]
+    end
+
+    ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface)
+end
+
+function get_impurity_density(tdld::ThermalDiffusionLithiumDensity{T}, pt::AbstractCoordinatePoint{T})::T where {T <: SSDFloat}
+    pt::CartesianPoint{T} = CartesianPoint(pt)
+    depth = tdld.calculate_depth2surface(pt)
+	tdld.Li_surface * Distributions.erfc(depth/2/sqrt(tdld.D_Li*tdld.Li_annealing_time))
+end
--- a/src/ImpurityDensities/ImpurityDensities.jl
+++ b/src/ImpurityDensities/ImpurityDensities.jl
@@ -47,3 +47,6 @@ include("LinearImpurityDensity.jl")
 include("CylindricalImpurityDensity.jl")
 
 include("BouleImpurityDensities/BouleImpurityDensities.jl")
+
+include("ThermalDiffusionLithiumDensity.jl")
+include("PtypePNJunctionImpurityDensity.jl")
--- a/src/MaterialProperties/MaterialProperties.jl
+++ b/src/MaterialProperties/MaterialProperties.jl
@@ -5,6 +5,7 @@ const elementary_charge = Float64(ustrip(u"C", Unitful.q))
 const ϵ0 = Float64(ustrip(u"F/m", Unitful.ϵ0))
 const kB = Float64(ustrip(u"J/K", Unitful.k))
 const me = Float64(ustrip(u"kg", Unitful.me))
+const R_gas = Float64(ustrip(u"cal/K/mol", Unitful.R))
 
 const material_properties = Dict{Symbol, NamedTuple}()
 
--- a/src/SolidStateDetectors.jl
+++ b/src/SolidStateDetectors.jl
@@ -77,7 +77,7 @@ export Simulation, simulate!
 export Event, drift_charges!
 export add_baseline_and_extend_tail
 export NBodyChargeCloud
-export LinearImpurityDensity, LinBouleImpurityDensity, LinExpBouleImpurityDensity
+export LinearImpurityDensity, LinBouleImpurityDensity, LinExpBouleImpurityDensity, ThermalDiffusionLithiumDensity, PtypePNJunctionImpurityDensity
 
 using Unitful: RealOrRealQuantity as RealQuantity
 const SSDFloat = Union{Float16, Float32, Float64}


examples/example_config_files/public_TrueCoaxial.yaml
```diff
@@ -26,23 +26,25 @@ medium: vacuum
...
 - semiconductor:
     material: HPGe
-    temperature: 77
+    temperature: 90
     impurity_density:
-      name: constant
-      value: 1e8cm^-3
+      name: PtypePNjunction
+      Li_annealing_temperature: 623
+      Li_annealing_time: 1080
```
fhagemann, line 33: We never define units of time and it is not clear, whether this should be seconds, minutes, hours, ...
Maybe force this value to have a unit? Otherwise, we have to add another field to the `units` section in the config file and define our default unit for time in the config file.

examples/example_config_files/public_TrueCoaxial.yaml
```diff
@@ -26,23 +26,25 @@ medium: vacuum
...
+      bulk_imp: -1e10cm^-3
```
fhagemann, line 34: As said, maybe this can become any other impurity density, something like
```suggestion
      bulk_imp:
        name: constant
        value: -1e10cm^-3
```
and then this is actually parsed as another (already existing) impurity density (?)

examples/example_config_files/public_TrueCoaxial.yaml
```diff
@@ -26,23 +26,25 @@ medium: vacuum
...
         τe: 1ms
-        τh_inactive: 80ns
-        τe_inactive: 80ns
+        τh_inactive: 1μs
+        τe_inactive: 1μs
       inactive_layer_geometry:
         tube:
           r:
-            from: 9.0
+            from: 8.957282
```
fhagemann, line 47: why?

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+
+A PN junctin impurity model based on lithium thermal diffusion and constant bulk impurity density.
```
fhagemann, line 4: Typo:
```suggestion
A PN junction impurity model based on lithium thermal diffusion and constant bulk impurity density.
```

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+struct PtypePNJunctionImpurityDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+	Li_annealing_temperature::T
+	Li_annealing_time::T
+    calculate_depth2surface::Function
+    bulk_imp::T
+    surface_imp_model::ThermalDiffusionLithiumDensity{T}
+    bulk_imp_model::ConstantImpurityDensity{T}
+end
```
fhagemann, line 24: As suggested before, I guess this could be generalized:
```suggestion
struct PtypePNJunctionImpurityDensity{T <: SSDFloat, ID <: AbstractImpurityDensity{T}} <: AbstractImpurityDensity{T}
	Li_annealing_temperature::T
	Li_annealing_time::T
    calculate_depth2surface::Function
    bulk_imp::T
    surface_imp_model::ThermalDiffusionLithiumDensity{T}
    bulk_imp_model::ID
end
```
as long as we allow for this (syntax-wise) in the config file.

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+* `Li_annealing_time::T`: lithium annealing time. The default value is 18 minutes.
+* `calculate_depth2surface::Function`: the function for describing the depth to surface.
+* `bulk_imp::T`: the bulk impurity density constant.
+* `bulk/surface_imp_model::T`: there two variables will be constructed with above parameters.
+"""
+
+struct PtypePNJunctionImpurityDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+	Li_annealing_temperature::T
+	Li_annealing_time::T
+    calculate_depth2surface::Function
```
fhagemann, line 20: As also suggested before: think about making this a `Geometry` and defining functions in SSD to calculate the distance to the surface of the geometries.
```suggestion
    contact_geometry::AbstractGeometry{T}
```

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+    ) where {T <: SSDFloat}
+    bulk_imp_model = ConstantImpurityDensity{T}(bulk_imp)
+    surface_imp_model = ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface)
+    PtypePNJunctionImpurityDensity(Li_annealing_temperature,Li_annealing_time,calculate_depth2surface, bulk_imp, surface_imp_model, bulk_imp_model)
+end
+
+function ImpurityDensity(T::DataType, t::Val{:PtypePNjunction}, dict::AbstractDict, input_units::NamedTuple)
+    Li_annealing_temperature = T(623)
+    Li_annealing_time= T(18*60)
+    calculate_depth2surface = pt->0.01-hypot(pt[1],pt[2]) # default for the public_TrueCoaxial.yaml
```
fhagemann, line 38: This default should be avoided AT ALL COSTS.
Then maybe define, that it should always use the contact with a specific ID (1?).

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+    bulk_imp::T
+    ) where {T <: SSDFloat}
+    bulk_imp_model = ConstantImpurityDensity{T}(bulk_imp)
+    surface_imp_model = ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface)
+    PtypePNJunctionImpurityDensity(Li_annealing_temperature,Li_annealing_time,calculate_depth2surface, bulk_imp, surface_imp_model, bulk_imp_model)
+end
+
+function ImpurityDensity(T::DataType, t::Val{:PtypePNjunction}, dict::AbstractDict, input_units::NamedTuple)
+    Li_annealing_temperature = T(623)
+    Li_annealing_time= T(18*60)
```
fhagemann, line 37: units?

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+    end
+    if haskey(dict, "Li_annealing_time")
+        Li_annealing_time = _parse_value(T, dict["Li_annealing_time"], u"s")
+    end
+    if haskey(dict, "calculate_depth2surface")
+        calculate_depth2surface = dict["calculate_depth2surface"]
+    end
+    if haskey(dict, "bulk_imp")
+        bulk_imp = _parse_value(T, dict["bulk_imp"], input_units.length^(-3))
+    end
```
fhagemann, line 52: There is a nice syntax to avoid this, that we should stick to
```suggestion
    bulk_imp = _parse_value(T, get(dict, "bulk_imp", default_value), input_units.length^(-3))
```
where `default_value` could then be something like the `-1e16` you define further above.
This will save lines of code.

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+
+function ImpurityDensity(T::DataType, t::Val{:PtypePNjunction}, dict::AbstractDict, input_units::NamedTuple)
+    Li_annealing_temperature = T(623)
+    Li_annealing_time= T(18*60)
+    calculate_depth2surface = pt->0.01-hypot(pt[1],pt[2]) # default for the public_TrueCoaxial.yaml
+    bulk_imp = T(-1e16)
+
+    if haskey(dict, "Li_annealing_temperature")
+        Li_annealing_temperature = _parse_value(T, dict["Li_annealing_temperature"], u"K")
+    end
```
fhagemann, line 43: This should be parsed in the units of the config file
```suggestion
    if haskey(dict, "Li_annealing_temperature")
        Li_annealing_temperature = _parse_value(T, dict["Li_annealing_temperature"], input_units.temperature)
    end
```

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,59 @@
...
+    if haskey(dict, "Li_annealing_time")
+        Li_annealing_time = _parse_value(T, dict["Li_annealing_time"], u"s")
```
fhagemann, line 45: You have to document somewhere, that the annealing time is in `u"s"`, ideally in the docstring AND the documentation. Or FORCE this number to have a unit.

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,68 @@
+"""
+    struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+
+Lithium impurity density model
+
+ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
+ 
+## Fields
+* `Li_annealing_temperature::T`: lithium annealing temperature when the lithium is diffused into the crystal. The default value is 623 K.
+* `Li_annealing_time::T`: lithium annealing time. The default value is 18 minutes.
```
fhagemann, line 10:
```suggestion
* `Li_annealing_time::T`: lithium annealing time in seconds. The default value is 18 minutes.
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,68 @@
+"""
+    struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+
+Lithium impurity density model
+
+ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
+ 
+## Fields
+* `Li_annealing_temperature::T`: lithium annealing temperature when the lithium is diffused into the crystal. The default value is 623 K.
```
fhagemann, line 9:
```suggestion
* `Li_annealing_temperature::T`: lithium annealing temperature in Kelvin, when the lithium is diffused into the crystal. The default value is 623 K.
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,68 @@
...
+* `Li_annealing_time::T`: lithium annealing time. The default value is 18 minutes.
+* `calculate_depth2surface::Function`: the function for describing the depth to surface
+* `Li_surface::T`: the lithium concentration in the surface
```
fhagemann, line 12: To me `Li_surface` would be something I would call a geometric object, not the actual doping concentration..

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,68 @@
...
+function ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature::T,
+    Li_annealing_time::T,
+    calculate_depth2surface::Function,
+    Li_surface::T=calculate_Li_surface_saturated_density(Li_annealing_temperature),
+    D_Li::T=calculate_D_Li(Li_annealing_temperature)
+    ) where {T <: SSDFloat}
+    ThermalDiffusionLithiumDensity(Li_annealing_temperature,Li_annealing_time,calculate_depth2surface, Li_surface, D_Li)
+end
+
+function ImpurityDensity(T::DataType, t::Val{:tdld}, dict::AbstractDict, input_units::NamedTuple)
```
fhagemann, line 46: `:tdld`?
Maybe use something more descriptive..

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,68 @@
...
+
+    if haskey(dict, "Li_annealing_temperature")
+        Li_annealing_temperature = _parse_value(T, dict["Li_annealing_temperature"], u"K")
+    end
+    if haskey(dict, "Li_annealing_time")
+        Li_annealing_time = _parse_value(T, dict["Li_annealing_time"], u"s")
+    end
+    if haskey(dict, "calculate_depth2surface")
+        calculate_depth2surface = dict["calculate_depth2surface"]
+    end
```
fhagemann, line 59: I thought, this cannot be defined in the config file (?)

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,68 @@
...
+
+    ThermalDiffusionLithiumDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface)
+end
+
+function get_impurity_density(tdld::ThermalDiffusionLithiumDensity{T}, pt::AbstractCoordinatePoint{T})::T where {T <: SSDFloat}
+    pt::CartesianPoint{T} = CartesianPoint(pt)
+    depth = tdld.calculate_depth2surface(pt)
+	tdld.Li_surface * Distributions.erfc(depth/2/sqrt(tdld.D_Li*tdld.Li_annealing_time))
```
fhagemann, line 67: The function `erfc` is actually from `SpecialFunctions`, which is imported in `Distributions`:
image

The cleanest way would be to add `SpecialFunctions` as an explicit dependency to the `Project.toml` and use that here, for better version control.

```suggestion
	tdld.Li_surface * SpecialFunctions.erfc(depth/2/sqrt(tdld.D_Li*tdld.Li_annealing_time))
```

fhagemann (changes requested):

docs/src/man/electric_potential.md
```diff
@@ -105,6 +105,25 @@ impurity_density:
...
+calculate_depth2surface = pt -> 0.01-hypot(pt[1], pt[2])
+Li_annealing_temperature::T=623
+Li_annealing_time::T=18*60
+Bulk_Impurity = p_type_density = -1e10 * 1e6
+
+pn_junction_impurity_density = PtypePNJunctionImpurityDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface, p_type_density)
+sim.detector = SolidStateDetector(sim.detector, pn_junction_impurity_density)
+```
+
+This model couldn't be defined in the configuration because a depth2surface function is needed, except when using the public_TrueCoaxial.yaml.
```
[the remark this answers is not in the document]
fhagemann: Then, we should still have the general method as default option, with the possibility for the user to add a more performant method.

Default could look something like
```julia
depth2surface = pt -> distance_to_surface(pt, sim.detector.contacts[1].geometry)
```

fhagemann:
@zpdg Let me know if you want to work on these comments or whether you want me to take over some of this

zpdg:
> @zpdg Let me know if you want to work on these comments or whether you want me to take over some of this
> 
> 

Soon I will work on these.

docs/src/man/electric_potential.md
```diff
@@ -105,6 +105,25 @@ impurity_density:
 Here, the impurity density at the origin is $10^{10}$cm$^{-3}$ and it increases radially with the gradient $10^{10}$cm$^{-4}$.
 If no units are given, `init` is parsed in units of `units.length`$^{-3}$ and `gradient` in units of `units.length`$^{-4}$.
 
+### P-type PN junction Impurity Density
+A PN junctin impurity model based on lithium thermal diffusion and constant bulk impurity density. The surface lithium density is saturated.
```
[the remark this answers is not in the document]
zpdg: Yes, we can support any model for the bulk. I wrote this way to make the struct construction more simple.

examples/example_config_files/public_TrueCoaxial.yaml
zpdg: That makes sense. I will take a look

docs/src/man/electric_potential.md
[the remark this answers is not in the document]
fhagemann: What do you mean with „step-wise call“?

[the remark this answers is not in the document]
zpdg: This `depth_to_surface` function will be called in each step of the drifting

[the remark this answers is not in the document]
zpdg: > 

Because we need the impurity density to calculate the carrier mobility at each step. (At least in the new inactive layer simulation code)

zpdg:
@fhagemann This commit adds `distance_to_surface` for two kinds of surface primitives, which means the inactive layer of the `TrueCoaxial` and `InvertedCoax` could be simulated with built-in depth calculation for lithium diffusion .  I will add support for other primitives later. The "hard-code" `r` of PN junction boundary in the `TrueCoaxial` config will be processed later because that value has little impact and it is just an example. 

I think we better merge as soon as possible if no big issue. Then I will pull request the final one (which is simpler).  Because Claudia needs this code recently.

fhagemann:
Thanks for pushing the latest changes, would you mind if I push some minor (style) changes before I merge?

docs/src/man/electric_potential.md
```diff
@@ -105,6 +105,25 @@ impurity_density:
 Here, the impurity density at the origin is $10^{10}$cm$^{-3}$ and it increases radially with the gradient $10^{10}$cm$^{-4}$.
 If no units are given, `init` is parsed in units of `units.length`$^{-3}$ and `gradient` in units of `units.length`$^{-4}$.
 
+### P-type PN junction Impurity Density
+A PN junction impurity model based on lithium thermal diffusion and constant bulk impurity density. The surface lithium density is saturated.
+ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
```
fhagemann: This still says `constant bulk impurity density`:

```suggestion
A PN junction impurity model based on lithium thermal diffusion and custom bulk impurity density. The surface lithium density is saturated.
ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)

```

examples/example_config_files/public_TrueCoaxial.yaml
```diff
@@ -26,23 +26,28 @@ medium: vacuum
...
     material: HPGe
-    temperature: 77
+    temperature: 90
     impurity_density:
-      name: constant
-      value: 1e8cm^-3
+      name: PtypePNjunction
+      Li_annealing_temperature: 623K
+      Li_annealing_time: 18minute
+      doped_contact_id: 2
```
fhagemann, line 34: Is there a default if this is not set or is this ALWAYS required?

examples/example_config_files/public_TrueCoaxial.yaml
```diff
@@ -26,23 +26,28 @@ medium: vacuum
...
         τe: 1ms
-        τh_inactive: 80ns
-        τe_inactive: 80ns
+        τh_inactive: 1μs
+        τe_inactive: 1μs
       inactive_layer_geometry:
         tube:
           r:
-            from: 9.0
+            from: 8.957282 # set to the depth of the pn junction boundary when lithium diffusion temperature is 623K, time is 18 minutes
```
fhagemann, line 50: This should be handled differently in the future... 
In the code, one could do something like
```julia
# if not fully depleted at point `pt` => if inactive at point `pt`
if sim.imp_scale(pt) != 1 
if is_undepleted_point_type(sim.point_types(pt))
```

src/ConstructiveSolidGeometry/DistanceToSurface.jl
```diff
@@ -0,0 +1,4 @@
+function distance_to_surface(point::AbstractCoordinatePoint{T}, c::AbstractGeometry{T}) where {T}
+    distance_list=[distance_to_surface(point,surf) for surf in surfaces(c)]
+    minimum(filter(!isnan,distance_list))
+end
```
fhagemann: I guess this is a method for `AbstractVolumePrimitive` or `AbstractConstructiveGeometry` (because only for those, `surfaces` is defined). Also, this fails if all entries in `distance_list` are `NaN` (can this ever happen?)

This can probably be written in a more optimized way, something like:
```suggestion
function distance_to_surface(point::AbstractCoordinatePoint{T}, c::Union{<:AbstractConstructiveGeometry{T}, <:AbstractVolumePrimitive{T}})::T where {T}
    min_distance::T = T(Inf)
    for surf in surfaces(c)
        d::T = distance_to_surface(point, surf)
        min_distance = ifelse(isnan(d), min_distance, min(d, min_distance))
    end
    return min_distance
end
```
This would return `Inf` in case that all distances to all surfaces are `NaN`. Is that a good default value?

src/ConstructiveSolidGeometry/GeometryRounding.jl
```diff
@@ -7,3 +7,27 @@ csg_default_tol(::Type{Float32}) = 1f-8
...
+    return CylindricalPoint{T}( geom_round(pt.r), geom_round(pt.φ), geom_round(pt.z)  )
+end
+
+function geom_round(pt::CartesianPoint{T})::CartesianPoint{T} where {T <: Real}
+    return CartesianPoint{T}( geom_round(pt.x), geom_round(pt.y), geom_round(pt.z)  )
+end
+
+function geom_round(vpt::Vector{CP}) where {CP <: AbstractCoordinatePoint}
+    [geom_round(pt) for pt in vpt]
+end
```
fhagemann:
```suggestion
function geom_round(vpt::AbstractVector{<:AbstractCoordinatePoint})
    geom_round.(vpt)
end
```

src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
```diff
@@ -321,3 +321,48 @@ end
...
+        return norm(seg[1] - point)
+    else
+        v_point_2 = CartesianVector{T}(point - seg[2])
+        if geom_round(dot(v12,v_point_2)) ≥ T(0)
+            return norm(seg[2] - point)
+        else
+            return sqrt(abs(dot(v_point_1,v_point_1) - proj_on_v12^2))
+        end
+    end
+end
```
fhagemann, line 342: Can you use `Edge` for this, which is also defined in this package?

src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
```diff
@@ -321,3 +321,48 @@ end
...
+
+@inline _left_linear_interval(z::Real) = -z
+@inline _right_linear_interval(z::Real) = z
+get_r_limits(c::ConeMantle{T, T, <:Any, <:Any}) where {T} = (T(c.r), T(c.r))
+get_r_limits(c::ConeMantle{T, <:Tuple, <:Any, <:Any}) where {T} = c.r
+get_z_limits(c::ConeMantle{T}) where {T} = (c.origin[3]-c.hZ, c.origin[3]+c.hZ)
```
fhagemann: This only works, if the `ConeMantle` is neither rotated, nor translated in `x` or `y`.

fhagemann: Also, in general, this shouldn't go here in `ConeMantle.jl` but somewhere where all the utils functions are. Can you not use any of the functions defined in `src/ConstructiveSolidGeometry/Intervals.jl`? Or at least add them there instead of here.

src/ConstructiveSolidGeometry/SurfacePrimitives/EllipticalSurface.jl
```diff
@@ -190,3 +190,45 @@ end
...
+@inline _in_cyl_r(p::CylindricalPoint, r::Union{Real, Tuple{T,T}}) where {T} =  _in_cyl_r(CartesianPoint(p), r)
+
+@inline _left_radial_interval(r::Real) = 0
+@inline _left_radial_interval(r::Tuple{T,T}) where {T} = r[1]
+
+@inline _right_radial_interval(r::Real) = r
+@inline _right_radial_interval(r::Tuple{T,T}) where {T} = r[2]
+
+get_r_limits(a::EllipticalSurface{T, <:Union{T, <:Any, Nothing}, <:Any}) where {T} =
+    (_left_radial_interval(a.r), _right_radial_interval(a.r))
```
fhagemann: This seems to do exactly what `_radial_endpoints` in `src/ConstructiveSolidGeometry/Intervals.jl` does.

src/IO/SigGenInterface.jl
```diff
@@ -35,7 +35,7 @@ function readsiggen(file_path::String; dicttype::Type = Dict{String,Any}, T::Typ
         "inner_taper_width"    => 0,
         "taper_angle"          => 0,
         "taper_length"         => 0,
-        "Li_thickness"         => 0,
+        "lithium_thickness"         => 0,
```
fhagemann: I don't think anything should be changed in this file at all. This might break stuff...

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,55 @@
...
+* `bulk_imp_model::AbstractImpurityDensity{T}`: the density of the p-type impurity.
+"""
+
+struct PtypePNJunctionImpurityDensity{T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}} <: AbstractImpurityDensity{T}
+	lithium_annealing_temperature::T
+	lithium_annealing_time::T
+    contact_with_lithium_doped::G
+    surface_imp_model::ThermalDiffusionLithiumDensity{T,G}
+    bulk_imp_model::AbstractImpurityDensity{T}
+end
```
fhagemann: I think this could be made more generic by removing everything lithium-related (the lithium part is anyhow saved in `surface_imp_model`)
```suggestion
struct PtypePNJunctionImpurityDensity{T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}} <: AbstractImpurityDensity{T}
    contact_doped::G
    surface_imp_model::ThermalDiffusionLithiumDensity{T,G}
    bulk_imp_model::AbstractImpurityDensity{T}
end
```

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -0,0 +1,55 @@
...
+function PtypePNJunctionImpurityDensity{T,G}(
+    lithium_annealing_temperature::T,
+    lithium_annealing_time::T,
+    contact_with_lithium_doped::G,
+    bulk_imp_model::AbstractImpurityDensity{T},
+    distance_to_contact::Function
+    ) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}}
+    surface_imp_model = ThermalDiffusionLithiumDensity{T,G}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, distance_to_contact=distance_to_contact)
+    PtypePNJunctionImpurityDensity{T,G}(lithium_annealing_temperature,lithium_annealing_time, contact_with_lithium_doped, surface_imp_model, bulk_imp_model)
+end
```
fhagemann: This can become one function if you pass `distance_to_contact` a default function.

```suggestion
function PtypePNJunctionImpurityDensity{T,G}(
    lithium_annealing_temperature::T,
    lithium_annealing_time::T,
    contact_with_lithium_doped::G,
    bulk_imp_model::AbstractImpurityDensity{T},
    distance_to_contact::Function = pt::AbstractCoordinatePoint{T} -> ConstructiveSolidGeometry.distance_to_surface(pt, contact_with_lithium_doped)
    ) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}}
    surface_imp_model = ThermalDiffusionLithiumDensity{T,G}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, distance_to_contact=distance_to_contact)
    PtypePNJunctionImpurityDensity{T,G}(lithium_annealing_temperature,lithium_annealing_time, contact_with_lithium_doped, surface_imp_model, bulk_imp_model)
end
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,66 @@
...
+"""
+
+struct ThermalDiffusionLithiumDensity{T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}} <: AbstractImpurityDensity{T}
+	lithium_annealing_temperature::T
+	lithium_annealing_time::T
+    contact_with_lithium_doped::G
+    distance_to_contact::Function
+	lithium_density_on_contact::T
+    lithium_diffusivity_in_germanium::T
+end
```
fhagemann: The indentation is a bit odd here.
```suggestion
struct ThermalDiffusionLithiumDensity{T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}} <: AbstractImpurityDensity{T}
    lithium_annealing_temperature::T
    lithium_annealing_time::T
    contact_with_lithium_doped::G
    distance_to_contact::Function
    lithium_density_on_contact::T
    lithium_diffusivity_in_germanium::T
end
```
Also, we might not need to save the `contact_geometry` here, if it is already saved in the `distance_to_contact` function. Not sure how this affects the runtime performance-wise, though.

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,66 @@
...
+    # H [cal]
+    if lithium_annealing_temperature<=873
+        D0 = 2.5e-3*1e-4
+        H = 11800
+    else
+        D0 = 1.3e-3*1e-4
+        H = 10700
+    end
+    D0 * exp(-H/(R_gas*lithium_annealing_temperature))
+end
```
fhagemann: Maybe explicitly force `D0` and `H` to be of type `T` here:
```suggestion
function calculate_lithium_diffusivity_in_germanium(lithium_annealing_temperature::T)::T where {T <: SSDFloat}
    # D0 [m^2*s^-1]
    # H [cal]
    D0::T, H::T = ifelse(lithium_annealing_temperature <= 873, T.((2.5e-3*1e-4, 11800)), T.((1.3e-3*1e-4, 10700)))
    D0 * exp(-H/(R_gas*lithium_annealing_temperature))
end
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,66 @@
...
+function calculate_lithium_saturated_density(lithium_annealing_temperature::T)::T where {T <: SSDFloat}
+    10^(21.27 - 2610.0/(lithium_annealing_temperature)) * 1e6
+end
```
fhagemann:
```suggestion
function calculate_lithium_saturated_density(lithium_annealing_temperature::T)::T where {T <: SSDFloat}
    exp10(21.27 - 2610.0/(lithium_annealing_temperature) + 6)
end
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -0,0 +1,66 @@
...
+    lithium_density_on_contact::T = calculate_lithium_saturated_density(lithium_annealing_temperature),
+    lithium_diffusivity_in_germanium::T = calculate_lithium_diffusivity_in_germanium(lithium_annealing_temperature)
+    ) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}}
+    ThermalDiffusionLithiumDensity{T,G}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, distance_to_contact, lithium_density_on_contact, lithium_diffusivity_in_germanium)
+end
+
+function ImpurityDensity(T::DataType, t::Val{:li_diffusion}, dict::AbstractDict, input_units::NamedTuple)
+    lithium_annealing_temperature = _parse_value(T, get(dict, "lithium_annealing_temperature", 623u"K"), internal_temperature_unit)
+    lithium_annealing_time = _parse_value(T, get(dict, "lithium_annealing_time", 18u"minute"), internal_time_unit)
+    contact_with_lithium_doped = haskey(dict, "contact_with_lithium_doped") ? dict["contact_with_lithium_doped"] : nothing # you don't have to pass the geometry of doped contact only when the distance_to_contact is passed
```
fhagemann, line 48: I would argue: you never need to store it.

src/SolidStateDetector/SolidStateDetector.jl
```diff
@@ -120,11 +120,25 @@ function SolidStateDetector{T}(config_file::AbstractDict, input_units::NamedTupl
...
+    end
+    if haskey(config_detector["semiconductor"], "impurity_density") &&
+       haskey(config_detector["semiconductor"]["impurity_density"], "name") &&
+       (config_detector["semiconductor"]["impurity_density"]["name"] == "li_diffusion" ||
+        config_detector["semiconductor"]["impurity_density"]["name"] == "PtypePNjunction")
+        doped_contact_id = Int(config_detector["semiconductor"]["impurity_density"]["doped_contact_id"])
+        config_detector["semiconductor"]["impurity_density"]["contact_with_lithium_doped"]=contacts[doped_contact_id].geometry
+    end
+
+    @assert haskey(config_detector, "semiconductor") "Each detector needs an entry `semiconductor`. Please define the semiconductor."
```
fhagemann: This line should be before L. 126

test/test_charge_drift_models.jl
```diff
@@ -82,14 +82,15 @@ end
...
         @test simA.detector.semiconductor.charge_trapping_model isa ConstantLifetimeChargeTrappingModel{T}
         @test simA.detector.semiconductor.charge_trapping_model.τh == T(1e-3)
         @test simA.detector.semiconductor.charge_trapping_model.τe == T(1e-3)
-        @test simA.detector.semiconductor.charge_trapping_model.τh_inactive == T(8e-8)
-        @test simA.detector.semiconductor.charge_trapping_model.τe_inactive  == T(8e-8)
+        @test simA.detector.semiconductor.charge_trapping_model.τh_inactive == T(1e-6)
+        @test simA.detector.semiconductor.charge_trapping_model.τe_inactive  == T(1e-6)
         @test simA.detector.semiconductor.charge_trapping_model.inactive_layer_geometry.origin == CartesianPoint{T}(0.0, 0.0, 0.005)
-        r0, r1 = T.((0.009, 0.01))
+        r0, r1 = T.((0.008957282, 0.01))
```
fhagemann, line 93: Still not a fan of this hard-coded number..

Project.toml
```diff
@@ -7,6 +7,7 @@ Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ArraysOfArrays = "65a8f2f4-9b39-5baf-92e2-a9cc46fdf018"
 Clustering = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
```
fhagemann, line 10: This is missing an entry in the `[compat]` section.
I had a quick look, can be something like
`SpecialFunctions = "1.6.1, 2"`

examples/example_config_files/public_TrueCoaxial.yaml
zpdg: There is a default, which is the value from the paper

fhagemann: No, I mean which contact is to use in case none is given

examples/example_config_files/public_TrueCoaxial.yaml
zpdg: Yes, now I choose the PN junction boundary as the sensitive/inactive boundary. However, we could also use the depletion boundary, which seems easier to implement, like this. And in this way, we don't need the `inactive_layer_geometry` anymore.

docs/src/man/electric_potential.md
```diff
@@ -106,24 +106,41 @@ Here, the impurity density at the origin is $10^{10}$cm$^{-3}$ and it increases
...
+distance_to_contact = pt -> 0.01 - hypot(pt[1], pt[2])
+Li_annealing_temperature::T = 623 # Kelvin
+Li_annealing_time::T = 18*60 # seconds
+p_type_density = -1e10 * 1e6
 
-pn_junction_impurity_density = PtypePNJunctionImpurityDensity{T}(Li_annealing_temperature, Li_annealing_time, calculate_depth2surface, p_type_density)
+# This code breaks and needs to be updated !!
+pn_junction_impurity_density = PtypePNJunctionImpurityDensity{T}(Li_annealing_temperature, Li_annealing_time, distance_to_contact, p_type_density)
 sim.detector = SolidStateDetector(sim.detector, pn_junction_impurity_density)
 ```
```
fhagemann: ToDo: provide a code example for the documentation which works.

src/ConstructiveSolidGeometry/DistanceToSurface.jl
zpdg: Yes, only for `AbstractVolumePrimitive` or `AbstractConstructiveGeometry`.

I noticed that the `NaN` corresponds to the zero area surface (basically 2d lines). 

I haven't dug into the detail. I basically adopt the code from the pull request 138, solve the error and test on the `TrueCoaxial` and `InvertedCoax`.

We could use `Inf`, but if the final return is `Inf`, we should at least throw an warning

examples/example_config_files/public_TrueCoaxial.yaml
fhagemann: Also, this should throw an error if no contact with the given ID exists.

src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
zpdg: I can look into this later

src/ConstructiveSolidGeometry/SurfacePrimitives/ConeMantle.jl
zpdg: Yes, I will look into this later

src/IO/SigGenInterface.jl
zpdg: Ah, sorry. I substituted all `Li_` with `lithium` in the command line. Didn't notice there is one

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
zpdg: Yes

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -41,7 +41,7 @@ function PtypePNJunctionImpurityDensity{T,G}(
 end
 
 function ImpurityDensity(T::DataType, t::Val{:PtypePNjunction}, dict::AbstractDict, input_units::NamedTuple)
-    lithium_annealing_temperature = _parse_value(T, get(dict, "lithium_annealing_temperature", 623u"K"), internal_temperature_unit)
+    lithium_annealing_temperature = _parse_value(T, get(dict, "lithium_annealing_temperature", 623u"K"), input_units.temperature)
     lithium_annealing_time = _parse_value(T, get(dict, "lithium_annealing_time", 18u"minute"), internal_time_unit)
```
zpdg, line 29: I use `internal_xxx_unit` following the code in `Boggs` trapping model. I think there is a reason for it:

The `internal_time_unit` is always `s` in SSD.jl. The time quantity in this model is also in unit of `s`. So if we use `input_units.time`, we need to force user to use `s` in the config when this model is adopted.

fhagemann: Yes, that’s why I updated the temperature but not the time one. Because the temperature is defined in the units section, I would update here to use whatever was defined there.

zpdg:
@fhagemann 
Could you merge this? There are two TODOs left (will be processed later):
1. Better support for the distance to surface calculation
	- consider translation and rotation of the geometry in the distance to surface calculation
	- use the `Edge` in the `distance_to_line_segment`
	- consider more surface primitives
2. Use the full depletion boundary depth as the partition line of inactive/sensitive region, instead of the hard-coded PN junction boundary depth

zpdg:
> I see you are still pushing commits. Is it done now or still ongoing work?

It's done

PR #498, opened 2025-06-11
Update default diffusion constants in germanium

giriPHM:


oschulz:
@giriPHM Can you explain what changes you're doing to the diffusion constant? Also, the PR seems to touch every line in MaterialProperties.jl?

fhagemann:
The old values are taken from [this PDF](https://github.com/user-attachments/files/20685027/Materials-Semiconductors-Si-and-Ge.pdf) (Materials-Semiconductors-Si-and-Ge.pdf).

Also, other sources quote this order of magnitude (see e.g. [this website](https://www.ioffe.ru/SVA/NSM/Semicond/Ge/electric.html)):
image


The new values probably correspond to the expected diffusion constant based on the [Einstein relation for charged particles](https://en.wikipedia.org/wiki/Einstein_relation_(kinetic_theory)), taking the low-field mobility values as input, at a temperature of ~78K.

$$ D_{e/h} = \dfrac{\mu_{e/h} k_B T}{q} $$

```julia
# electrons
uconvert(u"cm^2/s", 50000u"cm^2 / (V*s)" * Unitful.k * 78u"K" / Unitful.q)
=> 336.07599722366194 cm^2 s^-1

# holes
uconvert(u"cm^2/s", 40000u"cm^2 / (V*s)" * Unitful.k * 78u"K" / Unitful.q)
=> 268.86079777892957 cm^2 s^-1
```

fhagemann (changes requested):
If we change the values, then also the documentation needs an update:
image

https://juliaphysics.github.io/SolidStateDetectors.jl/stable/man/charge_drift/#Diffusion

fhagemann:
Note that there is also https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/476, where the diffusion constant is calculated via the Einstein relation from the mobility. The updated default values in this PR will be correct only for a temperature of ~78K. For the future, it then might make sense to ALWAYS determine them via the Einstein relation based on the temperature of the semiconductor rather than having one (obviously temperature-dependent) default value in material properties.

oschulz:
> For the future, it then might make sense to ALWAYS determine them via the Einstein relation based on the temperature of the semiconductor rather than having one (obviously temperature-dependent) default value in material properties

Sounds good to me.

fhagemann:
Okay, I also know where the old values might arise from.

Look at this table from G.F. Knoll "Radiation Detection and Measurement", 3rd edition (2000), p. 357:
image

If you take the electron and hole values at 300K from this table, you get
```julia
# electrons
uconvert(u"cm^2/s", 3900u"cm^2 / (V*s)" * Unitful.k * 300u"K" / Unitful.q)
=> 100.8227991670986 cm^2 s^-1

# holes
uconvert(u"cm^2/s", 1900u"cm^2 / (V*s)" * Unitful.k * 300u"K" / Unitful.q)
=> 49.11879959422752 cm^2 s^-1
```

With the values at 77K, we would get
```julia
# electrons
uconvert(u"cm^2/s", 3.6e4u"cm^2 / (V*s)" * Unitful.k * 77u"K" / Unitful.q)
=> 238.87247802666437 cm^2 s^-1

# holes
uconvert(u"cm^2/s", 4.2e4u"cm^2 / (V*s)" * Unitful.k * 77u"K" / Unitful.q)
=> 278.68455769777506 cm^2 s^-1
```

fhagemann (approved):
This looks good to me now, I'll let @oschulz decide about merging this.

PR #496, opened 2025-06-07
Add support for simulation for dead layer-Step5: charge drift model

zpdg:
Charge drift model for the inactive layer. The mobility curves:
image

Code for the plot:
```julia
using Revise
using Pkg
using StatsBase
using Plots
using Unitful
Pkg.develop(path="./SolidStateDetectors.jl")
using SolidStateDetectors
using SolidStateDetectors:Electron,Hole

T = Float64
mm=T(1/1000)
det_rin=1*mm
det_z=det_r=10*mm
z_draw=det_z/2
pn_r=8.957282*mm
sim = Simulation{T}(SSD_examples[:TrueCoaxial])

calculate_electric_potential!(sim, max_n_iterations = 10, grid=  Grid(sim), verbose = false, depletion_handling = true)
g = sim.electric_potential.grid
ax1, ax2, ax3 = g.axes
bulk_tick_dis=0.05*mm
dl_tick_dis=0.01*mm
user_additional_ticks_ax1=sort(vcat(ax1.interval.left:bulk_tick_dis:pn_r,pn_r:dl_tick_dis:ax1.interval.right))
user_ax1 = typeof(ax1)(ax1.interval, user_additional_ticks_ax1)
user_g = typeof(g)((user_ax1, ax2, ax3))
calculate_electric_potential!(sim, refinement_limits = 0.1, use_nthreads=8, grid =  user_g, depletion_handling = true)
calculate_electric_field!(sim)
calculate_weighting_potential!(sim, 1,use_nthreads=8,depletion_handling = true)
calculate_weighting_potential!(sim, 2,use_nthreads=8,depletion_handling = true);

cdm = sim.detector.semiconductor.charge_drift_model
depth_list = 0 : 0.01*mm : (det_r-pn_r)
hole_mobility_list=[]
electron_mobility_list=[]
for depth in depth_list
    r=det_r-depth
    pt = CylindricalPoint{T}([r,0,z_draw])
    pt = CartesianPoint{T}(pt)
    push!(hole_mobility_list, cdm.calculate_mobility(pt, Hole))
    push!(electron_mobility_list, cdm.calculate_mobility(pt, Electron))
end
plot(depth_list/mm, hole_mobility_list, label="Hole", lw=4)
plot!(depth_list/mm, electron_mobility_list, label="Electron", lw=4)
plot!(ylabel = "Mobility [cm\$^2\$/V/s]", xlabel = "Depth to surface [mm]",
[...]

PR #154, opened 2021-05-06
CSG: distance_to_surface for most surface primitives

hervasa2:
Added `distance_to_surface()` to the following surface primitives

- ConalPlane (Allocation free, worst case btime ~ 45ns)
- ConeMantle (Allocation free, worst case btime ~ 35ns)
- CylindricalAnnulus (Allocation free, worst case btime ~ 85ns)
- ToroidalAnnulus(Allocation free, worst case btime ~ 120ns)
- TorusMantle (Allocation free, worst case btime ~ 85ns)
- SphereMantle (Allocation free, worst case btime ~ 10ns)

Note that quoted btimes are worst case geometric scenarios, with the right conditions btime can be up to 100 times faster. These faster scenarios occur most often when close to a surface. 

`distance_to_surface()` is pending for the following surface primitives

- RegularPolygon
- RegularPrismMantle

This will wait until we have a decision on `φ_0` for these primitives.

Dependencies added for `distance_to_surface()`

- PlanarPoint (Problem is often reduced to 2D)
- LinePrimitives (`Arc` and `Line`. Problem is often reduced to 2D)
- Line constructors from surface primitives. Eg. `Arc(::TorusMantle)`, `LineSegment(::ConeMantle)`
- `distance_to_line()`

Various `φ` and `z` scans are shown below with the colorbar indicating distance to surface:

`cp = CSG.ConalPlane(1, 3, 3, 4, π/4, 0, 2)`
Screen Shot 2021-05-06 at 9 31 43 PM
distance_to_conal_plane

`cm = CSG.ConeMantle(3, 1, 0, 3π/2, 0 ,2)`
Screen Shot 2021-05-06 at 9 35 31 PM
distance_to_cone_mantle
distance_to_cone_mantle2

`ca = CSG.CylindricalAnnulus(1, 2, -π/4, 3π/2, 1)`
Screen Shot 2021-05-06 at 9 37 40 PM
distance_to_cylindrical_annulus

`ta = CSG.ToroidalAnnulus(3,1,2,π/3,-π/6,π,-1)`
Screen Shot 2021-05-06 at 9 38 37 PM
distance_to_toroidal_annulus

`tm = CSG.TorusMantle(3,2,0,π,0,3π/2,-1)`
Screen Shot 2021-05-06 at 9 39 32 PM
distance_to_torus_mantle
distance_to_torus_mantle2

`sm = CSG.SphereMantle{T}(2)`
Screen Shot 2021-05-06 at 9 40 19 PM
[...]

PR #502, opened 2025-06-18
Add support for simulation for dead layer: optimization, more geometry support, tutorials, etc

zpdg:
The almost final PR for the dead layer simulation:
- [x] Support the geometry of BEGe/PPCGe/ICPCGe
- [ ] Support more detector geometry, and consider the translation and rotation -- this will leave to another PR by someone else. 
- [x] Optimize the code, docs and docstrings
- [x] Write a tutorial subsection for the dead layer simulation

Add one more at 0620:
- [ ] Use the PointType or the effective charge to decide if a point is inside the inactive layer geometry for the trapping model instead of using the user-defined new inactive_layer_geometry -- this will leave to another PR.

### DIFF
--- a/docs/src/man/charge_drift.md
+++ b/docs/src/man/charge_drift.md
@@ -165,6 +165,12 @@ detectors:
         value: 5.6769e15cm^-3
 ```
 
+The `charge_drift_model` needs:
+- `model`: the name of the charge drift model, which in this case is `InactiveLayerChargeDriftModel`
+- `temperature`: the crystal temperature
+- `bulk_impurity_density`: the density of the p-type impurity in the crystal
+- `surface_impurity_density`: the density profile of lithium (n-type) in the surface layer
+- `neutral_impurity_density`: the density profile of the non-ionized impurity in the crystal
 
 ### Custom Charge Drift Model
 
--- a/docs/src/man/charge_drift.md
+++ b/docs/src/man/charge_drift.md
@@ -166,11 +166,11 @@ detectors:
 ```
 
 The `charge_drift_model` needs:
-- `model`: the name of the charge drift model, which in this case is `InactiveLayerChargeDriftModel`
-- `temperature`: the crystal temperature
+- `model`: the name of the charge drift model, which in this case is `InactiveLayerChargeDriftModel`.
+- `temperature`: the crystal temperature.
 - `bulk_impurity_density` (optional): the density of the p-type impurity in the crystal.
-- `surface_impurity_density` (optional): the density profile of lithium (n-type) in the surface layer
-- `neutral_impurity_density`: the density profile of the non-ionized impurity in the crystal
+- `surface_impurity_density` (optional): the density profile of lithium (n-type) in the surface layer.
+- `neutral_impurity_density`: the density profile of the non-ionized impurity in the crystal.
 
 > Note that the fields `bulk_impurity_density` and `surface_impurity_density` do not need to be explicitly defined if `PtypePNJunctionImpurityDensity` is chosen as the overall impurity density profile.
 
--- a/docs/src/man/charge_drift.md
+++ b/docs/src/man/charge_drift.md
@@ -135,6 +135,37 @@ If no temperature is given as a parameter, the calculations will be performed at
 It should be noted that the correct model has not yet been identified, and the parameters inside these configuration files -besides the default ADL ones- are just educated guesses.
 
 
+### Inactive Layer Charge Drift Model
+
+The [`InactiveLayerChargeDriftModel`](@ref) describes a system in which electrons and holes move along the electric field lines. The mobilities for electrons and holes are scalar ($+$ for holes, and $-$ for electrons), and thus, the velocity field has the same (or opposite) direction as the electric field.
+
+The mobilities are calculated considering three major scattering process: scattering off ionized impurities, neutral impurities and acoustic phonons. Thus, the mobilities depend on the impurity concentrations and temperature.
+
+The precision of the the calculation `T` (`Float32` or `Float64`) has to be given as a keyword `T`. Note that `T` has to be of the same type as the chosen in the simulation:
+
+The `InactiveLayerChargeDriftModel` can be specified in the configuration file as field `charge_drift_model` of the `semiconductor` of a detector, e.g.
+
+```yaml 
+detectors:
+  semiconductor:
+    # ...
+    charge_drift_model:
+      model: InactiveLayerChargeDriftModel
+      temperature: 90K
+      bulk_impurity_density:
+        name: constant
+        value: -1e10cm^-3
+      surface_impurity_density:
+        name: li_diffusion
+        lithium_annealing_temperature: 623K
+        lithium_annealing_time: 18minute
+        doped_contact_id: 2
+      neutral_impurity_density:
+        name: constant
+        value: 5.6769e15cm^-3
+```
+
+
 ### Custom Charge Drift Model
 
 The user can implement and use his own drift model.
--- a/docs/src/man/charge_drift.md
+++ b/docs/src/man/charge_drift.md
@@ -168,9 +168,11 @@ detectors:
 The `charge_drift_model` needs:
 - `model`: the name of the charge drift model, which in this case is `InactiveLayerChargeDriftModel`
 - `temperature`: the crystal temperature
-- `bulk_impurity_density`: the density of the p-type impurity in the crystal
-- `surface_impurity_density`: the density profile of lithium (n-type) in the surface layer
+- `bulk_impurity_density` (optional): the density of the p-type impurity in the crystal.
+- `surface_impurity_density` (optional): the density profile of lithium (n-type) in the surface layer
 - `neutral_impurity_density`: the density profile of the non-ionized impurity in the crystal
+
+> We could save the `bulk_impurity_density` and `surface_impurity_density` fields, if the `PtypePNJunctionImpurityDensity` is defined as the overall impurity density profile.
 
 ### Custom Charge Drift Model
 
--- a/docs/src/man/charge_drift.md
+++ b/docs/src/man/charge_drift.md
@@ -172,7 +172,7 @@ The `charge_drift_model` needs:
 - `surface_impurity_density` (optional): the density profile of lithium (n-type) in the surface layer
 - `neutral_impurity_density`: the density profile of the non-ionized impurity in the crystal
 
-> We could save the `bulk_impurity_density` and `surface_impurity_density` fields, if the `PtypePNJunctionImpurityDensity` is defined as the overall impurity density profile.
+> Note that the fields `bulk_impurity_density` and `surface_impurity_density` do not need to be explicitly defined if `PtypePNJunctionImpurityDensity` is chosen as the overall impurity density profile.
 
 ### Custom Charge Drift Model
 
--- a/docs/src/man/charge_drift.md
+++ b/docs/src/man/charge_drift.md
@@ -176,7 +176,7 @@ The `charge_drift_model` needs:
 
 ### Custom Charge Drift Model
 
-The user can implement and use his own drift model.
+The user can implement and use their own drift model.
 
 The first step is to define a `struct` for the model which is a subtype of `SolidStateDetectors.AbstractChargeDriftModel`:
 
--- a/docs/src/tutorials/dead_layer_simulation_lit.jl
+++ b/docs/src/tutorials/dead_layer_simulation_lit.jl
@@ -5,11 +5,8 @@
 # - constant lifetime trapping model
 # - mobility-tied diffusion model
 
-using Revise
-using Pkg
 using Plots
 using Unitful
-using Printf
 using SolidStateDetectors
 T = Float64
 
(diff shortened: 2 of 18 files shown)


zpdg:
@oschulz @fhagemann Hi, guys! The `distance_to_surface` for three surface primitives (elliptical surface, cone mantle and torus mantle) are supported now (at simple scenario). So, we could use this repo to simulate the inactive layer for PPCGe/BEGe/ICPCGe now. However, if we want full support, it would take much more time. Could someone else take this part? It's all about CSG support. I would like to focus on other parts.

zpdg:
> @oschulz @fhagemann Hi, guys! The `distance_to_surface` for three surface primitives (elliptical surface, cone mantle and torus mantle) are supported now (at simple scenario). So, we could use this repo to simulate the inactive layer for PPCGe/BEGe/ICPCGe now. However, if we want full support, it would take much more time. Could someone else take this part? It's all about CSG support. I would like to focus on other parts.

Actually, there are many history PRs about this. Very nice and cool work mainly from @hervasa2:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/127
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/138
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/139
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/154
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/156
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/159

But after CSG refactored, we didn't add this back. We should add this back and adapt to the current CSG implementation.

oschulz:
> However, if we want full support, it would take much more time. Could someone else take this part? It's all about CSG support

What do we need to do there?

zpdg:
> What do we need to do there?

Implement the `distance_to_surface` for all other surface primitives and consider possible rotation and translation. Right now, only three surface primitives are supported, and I didn't consider complex partial $\phi$ range, such $0-\pi$

oschulz:
> Implement the distance_to_surface for all other surface primitives 

Sure, we can do that, we need to do that anyway.

zpdg:
> And will you be working on those or do you also would like some support there?

Yes, I will work on this. Right now I don't need anything. I will reach out if I need.

docs/src/man/charge_drift.md
```diff
@@ -135,9 +135,48 @@ If no temperature is given as a parameter, the calculations will be performed at
...
+```
+
+The `charge_drift_model` needs:
+- `model`: the name of the charge drift model, which in this case is `InactiveLayerChargeDriftModel`
+- `temperature`: the crystal temperature
+- `bulk_impurity_density` (optional): the density of the p-type impurity in the crystal.
+- `surface_impurity_density` (optional): the density profile of lithium (n-type) in the surface layer
+- `neutral_impurity_density`: the density profile of the non-ionized impurity in the crystal
+
+> We could save the `bulk_impurity_density` and `surface_impurity_density` fields, if the `PtypePNJunctionImpurityDensity` is defined as the overall impurity density profile.
```
fhagemann: Is this comment supposed to be here?

docs/src/tutorials/dead_layer_simulation_lit.jl
```diff
@@ -0,0 +1,141 @@
+# # Inactive layer simulation Chain: True coaxial detector
+
+# ## Set the physical models for the inactive layer (dead layer) simulation with YAML
+# - drift model for the inactive layer
+# - constant lifetime trapping model
+# - mobility-tied diffusion model
+
+using Revise
```
fhagemann: Remove all development-related packages, that do not necessarily have to be part of the tutorial.

fhagemann: The docs build is failing because some packages in your tutorial are not in the `docs/Project.toml`
(Revise, Pkg, …)

docs/src/man/charge_drift.md
zpdg: Yes, for reminding users. Where it should be?

docs/src/tutorials/dead_layer_simulation_lit.jl
zpdg: sure

docs/src/man/charge_drift.md
fhagemann: Maybe I'm misunderstanding what you mean by "save"..
"Save" as in "write out to a file"? Or "save" as in "there is no need to spend"?

zpdg: Ah, I mean the latter.

fhagemann: Then maybe rephrase this to make it more clear.
Just to suggest something:
```suggestion
> Note that the fields `bulk_impurity_density` and `surface_impurity_density` do not need to be explicitly defined if `PtypePNJunctionImpurityDensity` is chosen as the overall impurity density profile.
```
but choose whatever you feel would describe it best!

docs/src/man/charge_drift.md
```diff
@@ -135,9 +135,48 @@ If no temperature is given as a parameter, the calculations will be performed at
 It should be noted that the correct model has not yet been identified, and the parameters inside these configuration files -besides the default ADL ones- are just educated guesses.
 
 
+### Inactive Layer Charge Drift Model
+
+The [`InactiveLayerChargeDriftModel`](@ref) describes a system in which electrons and holes move along the electric field lines. The mobilities for electrons and holes are scalar ($+$ for holes, and $-$ for electrons), and thus, the velocity field has the same (or opposite) direction as the electric field.
```
fhagemann, line 140: This currently breaks the docs build.
```suggestion
The `InactiveLayerChargeDriftModel` describes a system in which electrons and holes move along the electric field lines. The mobilities for electrons and holes are scalar ($+$ for holes, and $-$ for electrons), and thus, the velocity field has the same (or opposite) direction as the electric field.
```

zpdg: Do you know why? I found that there is an empty line between the docstring and the struct declaration of `InactiveLayerChargeDriftModel`. But there aren't empty lines in other structs. Is this the reason behind?

zpdg: Ah, seems so. I found this in the julia tutorial "https://docs.julialang.org/en/v1/manual/documentation/": "Note that no blank lines or comments may intervene between a docstring and the documented object."

zpdg:
> Is this good to go?

I have checked the docs. I think it's okay now.
Maybe we should leave this to another PR?: 
> - [ ] Use the PointType or the effective charge to decide if a point is inside the inactive layer geometry for the trapping model instead of using the user-defined new inactive_layer_geometry

In addition, the current lithium diffusion model calculates the impurity at each drift step each time rather than interpolating the pre-calculated impurity profile. We could also leave this update and the speed test to another PR.

fhagemann:
Yes, let‘s leave this PR as is (it is in a good checkpoint state), and add other stuff in additional PRa.

src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
```diff
@@ -7,27 +7,23 @@ Three factors are considered in the mobility calculation: ionized impurities, ne
...
-## Extra field for constructing the model
-- `temperature::T`: the crystal temperature (Kelvin), the default is 90 (Kelvin).
 """
 struct InactiveLayerChargeDriftModel{T <: SSDFloat} <: AbstractChargeDriftModel{T}
     calculate_mobility::Function
-    temperature::T
-    neutral_imp_model::AbstractImpurityDensity{T}
-    bulk_imp_model::AbstractImpurityDensity{T}
-    surface_imp_model::AbstractImpurityDensity{T}
 end
```
fhagemann: The thing is, now that this is now a very generic charge drift model. You don't really need to have an `InactiveLayer` anymore, but just some setting where you have a way of calculating the mobilities for `Electron` and `Hole`, which is basically what then `getVe` and `getVh` would do.
Wouldn't it make sense to keep the models and the temperature in the struct and to get rid of `calculate_mobility` by exporting its functionality to a dedicated `getVe` and `getVh`?

fhagemann: Then, we are back to what I commented on the previous PR:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/496#discussion_r2145059848

That it might make sense to remove `calculate_mobility` and to keep the other fields.

zpdg: But we need to call the `calculate_mobility` in the diffusion calculation.

zpdg:
@fhagemann 
Thanks for fixing the test!

src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
zpdg: My understanding is that the struct of the drift model should keep the variables for the following calculation. Then the `calculate_mobility` is the only thing useful.

But if we remove this, we need to call the `_calculate_mobility_with_impurities` in the `getVe`, `getVh`, and `_add_fieldvector_diffusion!` with three impurity models passed (bulk/surface/neutral), which is too complicated.

fhagemann: The thing is, right now the struct basically does "nothing" except for providing a function with two methods which essentially are `getVe` and `getVh`. I don't see an issue with calling `_calculate_mobility_with_impurities` in `getVe` and `getVh`.

I have an idea in mind, so if you're fine with it, I could push a commit (and if you do not like it we can still revert it afterwards).

zpdg: No problem. Let's see how it goes.

fhagemann:
I did not run the tests locally, so I hope they will pass.
@zpdg Let me know what you think about the code changes

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -306,17 +306,18 @@ function _drift_charge!(
     done::Vector{Bool} = broadcast(pt -> !_is_next_point_in_det(pt, det, point_types), startpos)
     normal::Vector{Bool} = deepcopy(done)
 
-    # if the drifting model has a calculate_mobility function (scalar mobility), then we use the mobility-tied diffusion model
-    # TODO: adapt the code to 3D-vector mobility
-    use_mobility_tied_diffusion = hasproperty(det.semiconductor.charge_drift_model, :calculate_mobility)
+    # if the drifting model has a method for calculate_mobility function (scalar mobility), then we use the mobility-tied diffusion model
+    cdm = det.semiconductor.charge_drift_model
+    ## TODO: make this a keyword argument, so the user can choose if they would like to use this ?
+    use_mobility_tied_diffusion = hasmethod(calculate_mobility, Tuple{typeof(cdm), CartesianPoint{T}, Type{CC}})
```
fhagemann, line 312: TODO: make this a keyword argument, so the user can choose if they would like to use this ?

test/test_charge_drift_models.jl
```diff
@@ -204,21 +204,32 @@ end
...
         # the charge drift model should define methods for coordinate points and charge carriers:
-        @test simA.detector.semiconductor.charge_drift_model.calculate_mobility(CartesianPoint{T}(0,0,0), SolidStateDetectors.Electron) isa T
-        @test simA.detector.semiconductor.charge_drift_model.calculate_mobility(CartesianPoint{T}(0,0,0), SolidStateDetectors.Hole) isa T
+        @test hasmethod(SolidStateDetectors.calculate_mobility, Tuple{typeof(simA.detector.semiconductor.charge_drift_model), CartesianPoint{T}, Type{SolidStateDetectors.Electron}})
+        @test hasmethod(SolidStateDetectors.calculate_mobility, Tuple{typeof(simA.detector.semiconductor.charge_drift_model), CartesianPoint{T}, Type{SolidStateDetectors.Hole}})
+        @test SolidStateDetectors.calculate_mobility(simA.detector.semiconductor.charge_drift_model, CartesianPoint{T}(0,0,0), SolidStateDetectors.Electron) isa T
+        @test SolidStateDetectors.calculate_mobility(simA.detector.semiconductor.charge_drift_model, CartesianPoint{T}(0,0,0), SolidStateDetectors.Hole) isa T
     end
 
+    # TODO: Add a test that the detector is not fully depleted?
```
fhagemann: TODO: Add a test that the detector is not fully depleted?

fhagemann (commented):
I also added two `TODO`s to the code as comments.

zpdg:
> I did not run the tests locally, so I hope they will pass. @zpdg Let me know what you think about the code changes.
> 
> Instead of passing a `calculate_mobility` function to `_add_fieldvector_diffusion!`, you will now have to pass a `ChargeDriftModel` for which the SSD method `calculate_mobility` is defined.
> 
> This will allow to use it also for other charge drift models in the future.

The change is nice. And I ran the tutorial code and the result is okay.

I think we could merge now. And consider to merge the transition repo into the main?

src/ChargeDrift/ChargeDrift.jl
zpdg: This TODO could be postponed to a diffusion-refactoring PR (add support for ADL model? anisotropic diffusion?).

test/test_charge_drift_models.jl
zpdg: Done!

PR #503, opened 2025-06-19
Disallow Geant4_julia_jll v0.2.3, it's broken

oschulz:
We can't limit Geant4_jll directly to avoid the `length` issue in  v11.3.2+0, since

```
Geant4_jll = "11.2.0 - 11.3.2+0"
```

is not allowed, but limiting Geant4_julia_jll seems to work as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SolidStateDetectors"
 uuid = "71e43887-2bd9-5f77-aebd-47f656f0a3f0"
-version = "0.10.17"
+version = "0.10.18"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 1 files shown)


fhagemann:
Can we not wait until Geant4 v0.2.2 is released?

oschulz:
> Can we not wait until Geant4 v0.2.2 is released? And then let the General registry figure out, which versions are compatible with each other?

People will have lot's of trouble at the pulse-shape simulation workshop next week, if SSD+Geant4 doesn't work out of the box ...

We can make a new release as soon as the new build is out and works.

oschulz:
There's really no downside to having an SSD version in between that restricts Geant4_jll.

oschulz:
Ping @fhagemann

fhagemann:
Can we wait until tomorrow evening and see if a new version of Geant4 gets released. The issue is, that within Geant4 `length` is used instead of `Base.length`. But as said, this is already fixed on `main`.

fhagemann:
Geant4 v0.2.2 was released.

Can we instead try restricting Geant4 to minimum 02.2, now that’s it’s released? This should fix the failing tests and the error reported here.

PR #505, opened 2025-06-20
Add geometric validation for electrode termination in charge drift

zql2021:
**Problem**
Charge carriers incorrectly terminated when point_types flagged positions as `CD_ELECTRODE `due to grid discretization, even though they hadn't reached actual electrode geometry.(see 
https://github.com/JuliaPhysics/SolidStateDetectors.jl/discussions/474)

**Solution**
Added `crossing_pos in det.contacts` validation in `_check_and_update_position!()`:
Only terminate when charge actually reaches electrode geometry
Reclassify failed cases as `CD_FLOATING_BOUNDARY` to continue surface drift

### DIFF
--- a/src/ChargeDrift/ChargeDrift.jl
+++ b/src/ChargeDrift/ChargeDrift.jl
@@ -36,7 +36,8 @@ end
 function _drift_charges(det::SolidStateDetector{T}, grid::Grid{T, 3}, point_types::PointTypes{T, 3},
                         starting_points::VectorOfArrays{CartesianPoint{T}}, energies::VectorOfArrays{T},
                         electric_field::Interpolations.Extrapolation{<:SVector{3}, 3},
-                        Δt::RQ; max_nsteps::Int = 2000, diffusion::Bool = false, self_repulsion::Bool = false, verbose::Bool = true)::Vector{EHDriftPath{T}} where {T <: SSDFloat, RQ <: RealQuantity}
+                        Δt::RQ; max_nsteps::Int = 2000, diffusion::Bool = false, self_repulsion::Bool = false, 
+                        geometry_check::Bool = false, verbose::Bool = true)::Vector{EHDriftPath{T}} where {T <: SSDFloat, RQ <: RealQuantity}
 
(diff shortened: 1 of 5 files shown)


fhagemann:
Thank you for this initiative!
Do you have any experience how much this would slow down large-scale simulations? Because the geometry checks can become quite resourceful if done too often, especially if the detector has a lot of (complex-geometry) contacts.

fhagemann:
@zql2021 You had a sketch in #474, can you show a before/after with this PR for the exact same event. And also what the execution time would be for the two?

fhagemann:
I see, so for events that might be affected by this issue, there seems to be a significant increase in runtime.
Maybe it makes sense to make this extra geometry check optional (we could talk about what the default should be) and allow the user to choose when this geometry check should be applied (?)

src/ChargeDrift/ChargeDrift.jl
```diff
@@ -196,10 +196,15 @@ function _check_and_update_position!(
...
-                drift_path[n,istep] = crossing_pos
-                current_pos[n] = crossing_pos
-            elseif cd_point_type == CD_FLOATING_BOUNDARY
+                if crossing_pos in det.contacts
+                    done[n] = true
+                    drift_path[n,istep] = crossing_pos
+                    current_pos[n] = crossing_pos
+                else
+                    cd_point_type = CD_FLOATING_BOUNDARY
+                end
```
fhagemann: Something like `geometry_check` in line 199 that would be passed down to this function and then allow/disable the geometry check that you are proposing in this PR.
```suggestion
                if !geometry_check || crossing_pos in det.contacts
                    done[n] = true
                    drift_path[n,istep] = crossing_pos
                    current_pos[n] = crossing_pos
                else
                    cd_point_type = CD_FLOATING_BOUNDARY
                end
```

fhagemann: @zql2021 Any thoughts?

fhagemann: --> 270fc09

PR #509, opened 2025-07-09
Add multithreading support in the electric field calculation

fhagemann:
Not sure why we never did this before, but the electric field calculation did not support multithreading. Adding it can result in a significant speedup.

Example:
```julia
using SolidStateDetectors, Unitful
sim = Simulation(SSD_examples[:CGD])
calculate_electric_potential!(sim, max_tick_distance = 0.05u"mm",
    max_n_iterations = 1, refinement_limits = Float64[])
```
```
Simulation: ExampleCuboid
Electric Potential Calculation
Bias voltage: 2000.0 V
Precision: Float32
Device: CPU
Max. CPU Threads: 64
Coordinate system: Cartesian
N Refinements: -> 0
Convergence limit: 1.0e-7  => 0.0002 V
Initial grid size: (374, 366, 370)

```
Before this PR:
```julia
@time calculate_electric_field!(sim)
```
```
  6.976945 seconds (16 allocations: 579.623 MiB, 3.73% gc time)
```

After this PR (using 64 threads)
```julia
@time calculate_electric_field!(sim)
```
```
  0.304666 seconds (277 allocations: 579.653 MiB, 22.28% gc time)
```

@oschulz Not sure if the increase in allocations is something to worry about or not, but it is definitely faster now.

### DIFF
--- a/src/ElectricField/ElectricField.jl
+++ b/src/ElectricField/ElectricField.jl
@@ -54,7 +54,7 @@ function ElectricField(epot::ElectricPotential{T, 3, S}, point_types::PointTypes
 end
 
 
-function get_electric_field_from_potential(epot::ElectricPotential{T, 3, Cylindrical}, point_types::PointTypes{T}, fieldvector_coordinates=:xyz; use_nthreads::Int = Base.Threads.threadid())::ElectricField{T, 3, Cylindrical} where {T <: SSDFloat}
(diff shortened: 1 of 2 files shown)


src/ElectricField/ElectricField.jl
```diff
@@ -157,9 +157,7 @@ function get_electric_field_from_potential(epot::ElectricPotential{T, 3, Cylindr
             end
         end
     end
-    if fieldvector_coordinates == :xyz
-        ef = convert_field_vectors_to_xyz(ef, axφ)
-    end
```
fhagemann, line 162: Also, I removed `fieldvector_coordinates` from `get_electric_field_from_potential`, because I cannot think of a scenario where this should NOT be wanted.. 

Maybe this function can even be mutating to avoid allocating memory.

PR #510, opened 2025-07-14
Explicitly add `Base.` to method for `NamedTuple` and `Symbol`

fhagemann:
In `julia-1.12`, I am getting some warnings as SSD defines methods for `NamedTuple` and `Symbol` from `Base`.

Instead of just adding `import Base: NamedTuple, Symbol`, I have explicitly added the prefix `Base.` to all of these methods, which is consistent with the way SSD defines methods for other `Base` methods (e.g. `convert`, ...)


fhagemann:
Tests failing on `pre` seem to be related to `CxxWrap`

fhagemann:
So, `Geant4` does not allow for `CxxWrap@0.17` yet in its released version, but it's already added on its `master` branch. So, these tests should pass once a new version of `Geant4` is released.

PR #511, opened 2025-07-17
Add `ADL2016ChargeDriftModel` with updated values and electron-drift model

fhagemann:
I added the charge drift model from a paper from [Bruyneel _et al._, EPJA 52, 70 (2016)](https://doi.org/10.1140/epja/i2016-16070-9), including the values quoted in Tab. 2, as well as the intervalley scattering for electrons described in [36] [Bruyneel _et al._, NIMA 569, 764 (2006)](https://doi.org/10.1016/j.nima.2006.08.130).

I named the new charge drift model `ADL2016ChargeDriftModel`, because it's also published by the same `ADL` group, just 10 years later than the `ADLChargeDriftModel` from 2006.

The values I see there are also consistent with what I saw in my PhD thesis.

imageimage

I still need to add tests and some stuff in the constructor is still hard-coded, so I'll leave this as a draft PR for now.

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         version: ['1.10', '1', 'pre']
+    continue-on-error: ${{ matrix.version == 'pre' }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


fhagemann:
TODO: Add documentation.
Maybe even consider making this the default drift model?

PR #512, opened 2025-07-21
Add support for simulation for dead layer - Step6: Support using point types to define the inactive layer

claudiaalvgar:
An extra bit to the point types has been added (inactive_layer_bit) to distinguish the undepleted region due to the low bias voltage or due to the inactive layer presence. Two functions have been added (get_inactivelayer_indices and mark_inactivelayer_bits) to detect the inactive layer geometry from the impurity density model (the model giving the inactive layer region is PtypePNJunctionImpurityDensity) and increase inactive_layer_bit if inside the inactive layer.

These point types are passed to ChargeTrapping.jl so internally, the code knows what trapping lifetime should use without needing to pass explicitly the inactive_layer_geometry to the yaml file. However, the functionality of passing the inactive_layer_geometry to the yaml file is still possible, so that it only uses the inactive_layer_geometry if it’s passed to the yaml file, if not, it uses the point types to detect the inactive layer in the charge trapping model. In addition, a yaml file for an inverted coaxial detector using the point types to distinguish the inactive layer has been included.

Since the point types are changed after all the calculations requiring point types are done (electric potential) and we just need this functionality for the charge trapping model, the best approach to distinguish the undepleted regions is to increase this extra bit (just if a surface impurity model is defined in the impurity density model). The point type for undepleted regions would be 0000 1111 = 15 (as previously) while the undepleted regions due to inactive layer presence would be 0001 1111 = 31 and 0001 0111 = 23 (being 0001 0000 the inactive_layer_bit). The depleted regions would remain as previously (0000 1101 = 13).

The plotting in src/PlotRecipes/Potentials.jl has also been changed so that the plotting is not restricted to just 3 bits.

Point types before updating point types for inactive layer : 
testPR_plot

Point types after using get_inactivelayer_indices!(sim_undep.point_types.data) : 
testPR_pointtypes_plot

Code:
``` using Pkg
Pkg.activate(".")

using Revise
using Pkg
using Plots
using Unitful
using SolidStateDetectors
using StatsBase
using Printf
using LegendHDF5IO
using Random

T = Float64
mm=T(1/1000)

Random.seed!(1234)  # Fixed seed

calculate_electric_potential!(sim
    ,max_n_iterations = 10
    ,grid=  Grid(sim)
    ,verbose = true
    ,depletion_handling = true)

ax1, ax2, ax3 = g.axes

tick_dis=0.2*mm
user_additional_ticks_ax1=sort(vcat(ax1.interval.left:tick_dis:ax1.interval.right))
user_additional_ticks_ax3=sort(vcat(ax3.interval.left:tick_dis:ax3.interval.right))[2:end]
user_ax3 = typeof(ax3)(ax3.interval, user_additional_ticks_ax3)
user_g = typeof(g)((user_ax1, ax2, user_ax3))
calculate_electric_potential!(sim, refinement_limits = 0.1, use_nthreads=8, grid =  user_g, depletion_handling = true)

# Partially depleted detector

sim_undep = deepcopy(sim)
sim_undep.detector = SolidStateDetector(sim_undep.detector, contact_id = 2, contact_potential = 500);

calculate_electric_potential!( sim_undep,
                               depletion_handling = true,
                               convergence_limit = 1e-6,
                               refinement_limits = [0.2, 0.1, 0.05, 0.01],
                               verbose = false)


plot( 
    plot(sim_undep.electric_potential),
    plot(sim_undep.point_types),
    plot(sim_undep.imp_scale),
    layout = (1, 3), size = (1500, 700)
)

### CODE REVIEW

File: src/PlotRecipes/Potentials.jl [hunk 1]
```diff
@@ -81,9 +81,12 @@ end
     grid::CylindricalGrid{T} = sp.grid
     data = sp.data
     if eltype(data) == PointType
-        data = data .& 0x07
+        #data = data .& 0x07
+        data = data
     end
+    @info "Data range" extrema(data)
     @series begin
+    	clims := (minimum(data), maximum(data))
```
fhagemann: I would also be careful with this one, because right now, depleted was always green and undepleted always yellow. Not it‘s data-dependent what is shown as „yellow“…

claudiaalvgar: To get the old colour scheme (green for depleted and yellow for undepleted), it’s enough with doing data = data .& 0x07 in ```src/PlotRecipes/Potentials.jl``` which just plots the 3 first bits of the point type and therefore, it doesn’t matter if the inactive_layer_bit is set or not. I just commented that row to see the difference for this Pull Request. Point types for contact (0) and for the outside (1) will be hard to distinguish anyway if there’s an inactive layer. This is how it’ll look like considering just  data = data .& 0x07  which is consistent with previous plots and could be the default plot depending on what you want to visualise:
testPR_plot_0x07pointtypesplot

fhagemann: Can we add a keyword argument to the plot types where either we do &0x07 and fix the clims to still create the old plot, or omit the &0x07 and adjust the clims accordingly? Because it would be great if the point type plots could stay somewhat consistent with previous SSD versions

claudiaalvgar: Sure, I'll include this together with the rest of suggestions once the code is checked

src/PlotRecipes/Potentials.jl
```diff
@@ -76,14 +76,17 @@ end
...
     grid::CylindricalGrid{T} = sp.grid
     data = sp.data
     if eltype(data) == PointType
-        data = data .& 0x07
+        if !all_point_types_bits
+            data = data .& 0x07
+        end
     end
     @series begin
+        clims := (minimum(data), maximum(data))
```
fhagemann: This should not be here..

src/PlotRecipes/Potentials.jl
```diff
@@ -223,14 +226,17 @@ end
...
     grid::CartesianGrid3D{T} = sp.grid
     data = sp.data
     if eltype(data) == PointType
-        data = data .& 0x07
+        if !all_point_types_bits
+            data = data .& 0x07
+        end
     end
     @series begin
+        clims := (minimum(data), maximum(data))
```
fhagemann: And neither should this...

claudiaalvgar: I don't understand why some tests are failing for the plotting. Could you have a look at this? also, the plotting change is not mandatory, I just changed it to see the difference in colours in this PR but if it requires many modifications we can just leave it as it was

examples/example_config_files/ivc_inactivelayer.yaml
```diff
@@ -0,0 +1,188 @@
...
+      #                   from: 5
+      #                   to: 6  #changed this
+      #               h: 55
+      #               origin:
+      #                 z: 52.5
+      #           - tube:
+      #               r: 5
+      #               h: 1 #changed this
+      #               origin:
+      #                 z: 25
```
fhagemann: Is this supposed to stay like this?

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,95 @@ end
 
 Base.convert(T::Type{NamedTuple}, x::PointTypes) = T(x)
 
+"""
+    get_inactivelayer_indeces(vec::AbstractVector{UInt8})::Vector{Vector{Int}}
+
+Scans a 1D vector and returns index groups that surround a `0x00` and
+contain both `0x07` and `0x0f`, bounded by `0x0d`.
+"""
```
fhagemann: Can you comment on what `0x07`, `0x0f` and `0x0d` would be?
And would this need to be redefined if we were to add more bits to the point types?

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,95 @@ end
 
 Base.convert(T::Type{NamedTuple}, x::PointTypes) = T(x)
 
+"""
+    get_inactivelayer_indeces(vec::AbstractVector{UInt8})::Vector{Vector{Int}}
```
fhagemann: `indeces` does not exist. --> `indices`

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,95 @@ end
 
 Base.convert(T::Type{NamedTuple}, x::PointTypes) = T(x)
 
+"""
+    get_inactivelayer_indeces(vec::AbstractVector{UInt8})::Vector{Vector{Int}}
+
+Scans a 1D vector and returns index groups that surround a `0x00` and
+contain both `0x07` and `0x0f`, bounded by `0x0d`.
+"""
+function get_inactivelayer_indeces(vec::AbstractVector{UInt8})
```
fhagemann:
```suggestion
function get_inactivelayer_indeces(vec::AbstractVector{PointType})

```

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,95 @@ end
...
+                end
+
+                if has_07 && has_0f
+                    push!(pt_indices, matching)
+                end
+            end
+        else
+            i += 1
+        end
+    end
```
fhagemann: Can you comment what this function does?

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,95 @@ end
...
 
+    # dim 3 (vary k, fix i & j)
+    for i in 1:sz1, j in 1:sz2
+        vec = @view point_types[i, j, :]
+        matches = get_inactivelayer_indeces(vec)
+        for match in matches, idx in match
+            vec[idx] |= inactive_layer_bit
+        end
+    end
+end
```
fhagemann: Can you also comment what this is doing?
Are there cases where this might not be doing the correct thing?

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,95 @@ end
...
+
+    return pt_indices
+end
+"""
+    get_inactivelayer_point_types!(point_types::Array{UInt8, 3})
+
+Applies the bitwise marking rule to slices across all three dimensions of
+a 3D array, modifying it in-place.
+"""
+function get_inactivelayer_point_types!(point_types::Array{UInt8, 3})
```
fhagemann:
```suggestion
function get_inactivelayer_point_types!(point_types::Array{PointType, 3})

```

examples/example_config_files/ivc_inactivelayer.yaml
```diff
@@ -0,0 +1,188 @@
...
+          from: 4
+          to: 14
+        h: 0.1
+        origin:
+          - 0
+          - 0
+          - -0.05
+    charge_density:
+      name: constant
+      value: 0C/m^3 # Reasonable density: +2e-11 # => 2*10⁻¹¹ C/m³
```
fhagemann: Remove this?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -148,21 +152,33 @@ struct ConstantLifetimeChargeTrappingModel{T <: SSDFloat, G <: Union{<:AbstractG
     inactive_layer_geometry::G
 end
 
+is_inactive_layer(point_types::Array{UInt8, 3}) = any(val -> (val & inactive_layer_bit) != 0, point_types)
```
fhagemann:
```suggestion
is_inactive_layer(point_types::Array{PointType, 3}) = any(val -> (val & inactive_layer_bit) != 0, point_types)
```

Also: shouldn't this be called something like `has_inactive_layer` rather than `is_inactive_layer`?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -148,21 +152,33 @@ struct ConstantLifetimeChargeTrappingModel{T <: SSDFloat, G <: Union{<:AbstractG
...
+    user_defined_inactive_layer_geom = false
     inactive_layer_exist = !isnothing(ctm.inactive_layer_geometry)
+
+    if inactive_layer_exist
+        user_defined_inactive_layer_geom = true
+    else
+        get_inactivelayer_point_types!(point_types.data)
+        inactive_layer_exist = is_inactive_layer(point_types.data)
+    end
+
```
fhagemann: Would it be possible to add the `point_types` to the charge trapping model in `inactive_layer_geometry` instead of passing it through all the functions?

src/PlotRecipes/ElectricField.jl
```diff
@@ -5,7 +5,8 @@
                     φ = missing,
                     z = missing,
                     contours_equal_potential=false,
-                    full_det = false ) where {T}
+                    full_det = false,
+                    all_point_types_bits = false) where {T}
```
fhagemann: I don't like that we need to add keywords to plot recipes where they don't make sense.
(`all_point_types_bits` in `ElectricField`??)

src/PlotRecipes/ElectricField.jl
```diff
@@ -16,14 +17,15 @@
     unitformat --> :slash
 
     ef_magn = norm.(ef)
-    ElectricPotential(ef_magn, grid), cross_section, idx, value, internal_efield_unit, contours_equal_potential, full_det
+    ElectricPotential(ef_magn, grid), cross_section, idx, value, internal_efield_unit, contours_equal_potential, full_det, all_point_types_bits
```
fhagemann:
```suggestion
    ElectricPotential(ef_magn, grid), cross_section, idx, value, internal_efield_unit, contours_equal_potential, full_det, false
```
But in general, I would be strongly in favor in reverting this and to deal with this in the `PointTypes` plot-specific function

src/PlotRecipes/Potentials.jl
```diff
@@ -208,26 +212,30 @@ function get_crosssection_idx_and_value(grid::CartesianGrid3D{T}, x::Union{Real,
     cross_section, idx, value, units
 end
 
-@recipe function f(sp::ScalarPotential{T,3,Cartesian}; x = missing, y = missing, z = missing, contours_equal_potential = false) where {T <: SSDFloat}
+@recipe function f(sp::ScalarPotential{T,3,Cartesian}; x = missing, y = missing, z = missing, contours_equal_potential = false, all_point_types_bits= false) where {T <: SSDFloat}
```
fhagemann: Or, if you make this a keyword argument...
```suggestion
@recipe function f(sp::ScalarPotential{T,3,Cartesian}; x = missing, y = missing, z = missing, contours_equal_potential = false; all_point_types_bits= false) where {T <: SSDFloat}
```

src/PlotRecipes/Potentials.jl
```diff
@@ -58,29 +58,33 @@ function get_crosssection_idx_and_value(grid::CylindricalGrid{T}, r::Union{Real,
...
+    end
     
     title --> name * " @ $(cross_section) = $(round(units, Float64(uconvert(units,value*(cross_section == :φ ? u"°" : internal_length_unit))), sigdigits=3))"
     colorbar_title --> name
     seriescolor --> gradient
     clims --> clims
     unitformat --> :slash
 
-    sp, cross_section, idx, value, punit, contours_equal_potential, full_det
+    sp, cross_section, idx, value, punit, contours_equal_potential, full_det, all_point_types_bits
```
fhagemann: ... then you can pass this down as attribute (I think), and don't necessarily need to add this for `ElectricField`
```suggestion
    all_point_types_bits --> all_point_types_bits
    
    sp, cross_section, idx, value, punit, contours_equal_potential, full_det
```

fhagemann (changes requested):
Could we maybe split this into two PRs (one for the actual changes to the drift code and one for plotting the point types)?

examples/example_config_files/ivc_inactivelayer.yaml
fhagemann, line 1: Do you envision to add any sort of test for this?

src/PlotRecipes/Potentials.jl
fhagemann: Ignore the tests failing on `pre`, this is because of `Geant4` and not related to this PR.

examples/example_config_files/ivc_inactivelayer.yaml
claudiaalvgar: This yaml file is just to show that it is very easy to define the inactive_layer_geometry in the case of the True Coaxial yaml file but why it is important to have the point types options internally defined in the code because to define it in the yaml file can get complicated. This is commented because with this PR it's not needed although BOTH options are supported (define inactive_layer_geometry or not)... I don't mind if it's deleted or if stays like this

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Yes, these are the point types surrounding the inactive layer so 0x00 = 0000 0000 (outside the detector), 0x07 = 0000 0111 (part of the inactive layer, not only surrounded by pn_junction point types), 0x0f = 0000 1111 (also part of the inactive layer) and 0x0d = 0000 1101 (the depleted region inside the detector). So it finds the inactive layer bounded by 0x00 and 0x0d. In principle it won't need to be redefined

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: This function finds the  0x07 and 0x0f (inactive layer) bounded by at least two 0x00 and 0x0d and return the indices where inactive layer is

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: This increases the inactive_layer_bit in the indices returned by the previous function (it marks the inactive layer with an extra bit)

examples/example_config_files/ivc_inactivelayer.yaml
claudiaalvgar: This is just a copy of what it's defined in ```public_ivc_config.yaml```

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: But the point of this is not having to define inactive_layer_geometry or anything in the yaml file since it's already given by impurities. Passing the point types to all these functions is also necessary for refactoring the charge trapping models (i.e. using for example Boggs model inside the inactive layer and a constant lifetime charge trapping model outside the inactive the layer). This would be included in my next PR since at the moment it's not possible to use different charge trapping models inside and outside the inactive layer

src/PlotRecipes/ElectricField.jl
claudiaalvgar: I think the best for the plotting is that I exclude the changes to ```ElectricField.jl``` and ```Potentials.jl``` in this PR and you can probably open a new PR with your plotting suggestions ?

examples/example_config_files/ivc_inactivelayer.yaml
claudiaalvgar: Not really it's basically ```public_ivc_config.yaml``` with the inactive layer model in ```public_TrueCoaxial.yaml```

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: Done

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Done

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Done

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Done

examples/example_config_files/ivc_inactivelayer.yaml
fhagemann: I see. Maybe it would make sense to make this a short tutorial for the documentation.
(And maybe show plots depending on whether you define your own inactive layer or not).

src/ScalarPotentials/PointTypes.jl
fhagemann: Then maybe call this `set_` or `mark_` instead of `get_`.
For bulk bits, we call it `mark_bulk_bits!`: https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/5951c99c30180aacdaa8d4be82a4ddd069000cb9/src/ScalarPotentials/PointTypes.jl#L70-L84

examples/example_config_files/ivc_inactivelayer.yaml
fhagemann: I see.
The `public_ivc_config.yaml` is our standard file, so we added this to showcase how it can to be done.
For the `inactive_layer` one, you could omit this to focus on the actual changes.

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: Sorry, I think I was unclear with what I meant.

`ctm.inactive_layer_geometry` can be either a `Geometry` or `nothing` (see the parametric type `G <: Union{<:AbstractGeometry, Nothing}`)
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/c8c3bbfbf5599d6feb27713d18f19411029afb5c/src/ChargeTrapping/ChargeTrapping.jl#L143-L149



We could think about making this `Geometry`, `PointTypes` or `nothing`, i.e. `G <: Union{<:AbstractGeometry, PointTypes, Nothing}` to include the point type option:
- `Geometry` -> use the geometry
- `PointTypes` -> use the point types
- `Nothing` -> no inactive layer expected

or maybe we do not need nothing anymore (?)

src/PlotRecipes/ElectricField.jl
fhagemann: Sounds good!

src/ScalarPotentials/PointTypes.jl
fhagemann: I would always be careful dealing with comparing all bits, because you might only wanna use the ones you're interested in.

If you don't care about the bulk bit, you can just do
`point_type & 0x07 == 0x07` to find both `0x07` and `0x0f` at the same time.
And also, if we add some more bits left of inactive layer bit, the whole value might change from `0x07` to for example `0x27` and this function will fail then.

examples/example_config_files/ivc_inactivelayer.yaml
claudiaalvgar: Yes, this could also be a tutorial but then, should the yaml stay there? or should it be moved somewhere else? this is also the yaml file I use to test the refactored charge trapping models so it will be changed in the near future

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Yes, you're right thanks for pointing this. I added a bitmask so that it just compares the 4 first bits in case the other 4 bits are modified in the future.

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Sure, I changed the function name

examples/example_config_files/ivc_inactivelayer.yaml
claudiaalvgar: Done

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: But inactive_layer_geometry is initialised in the ChargeTrappingModel constructor while at that stage we don't know what are the point types. For this, we need to run the electric potential function so I don't really know how to implement this

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -148,21 +152,33 @@ struct ConstantLifetimeChargeTrappingModel{T <: SSDFloat, G <: Union{<:AbstractG
...
 
+    user_defined_inactive_layer_geom = false
     inactive_layer_exist = !isnothing(ctm.inactive_layer_geometry)
+
+    if inactive_layer_exist
+        user_defined_inactive_layer_geom = true
+    else
+        mark_inactivelayer_point_types!(point_types.data)
+        inactive_layer_exist = has_inactive_layer(point_types.data)
+    end
```
fhagemann: In this code, you are marking the `inactivelayer_point_types` every time the function `_calculate_signal` is called, which is done for every event.
I would advise to either check whether this has already been done (to avoid doing it multiple times), or to move this to the end of the potential calculation where we also mark the bulk bits.

How long does it take for this to run? We could either run this always by default or add a keyword argument to the potential calculation then, to ask the user if this is wanted or not.
Alternatively, this could throw a warning here, that if once wants to get correct results, the user should run `mark_inactivelayer_point_types!`  themselves on `sim.point_types`.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -172,7 +188,18 @@ function _calculate_signal(
...
-        τi::T = (inactive_layer_exist && in(path[i], ctm.inactive_layer_geometry)) ? τ_inactive : τ
+
+        τi::T = τ
+        
+        if user_defined_inactive_layer_geom
+            τi = (inactive_layer_exist && in(path[i], ctm.inactive_layer_geometry)) ? τ_inactive : τ
+        else
+            path_in_pt_grid = find_closest_gridpoint(path[i], point_types.grid) #returns the grid point where the path is
+            idx1, idx2, idx3 = path_in_pt_grid
+            pt_value = point_types.data[idx1, idx2, idx3] #returns the point type value in the path
```
fhagemann: There is a method for `getindex` that does exactly this
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/ScalarPotentials/PointTypes.jl#L61

```suggestion
            pt_value = point_types[path[i]]
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -172,7 +188,18 @@ function _calculate_signal(
...
+        τi::T = τ
+        
+        if user_defined_inactive_layer_geom
+            τi = (inactive_layer_exist && in(path[i], ctm.inactive_layer_geometry)) ? τ_inactive : τ
+        else
+            path_in_pt_grid = find_closest_gridpoint(path[i], point_types.grid) #returns the grid point where the path is
+            idx1, idx2, idx3 = path_in_pt_grid
+            pt_value = point_types.data[idx1, idx2, idx3] #returns the point type value in the path 
+            τi = (inactive_layer_exist && is_in_inactive_layer(pt_value)) ? τ_inactive : τ
+        end
```
fhagemann: I feel like this `if`-`else` could be avoided using multiple dispatch.
--> Maybe @oschulz can give his opinion on this.

src/ScalarPotentials/PointTypes.jl
```diff
@@ -22,26 +23,28 @@ How to get information out of a `PointType` variable `point_type`:
...
+const update_bit         = 0x01 # parse(UInt8, "00000001", base=2) # 1 -> do update; 0 -> do not update
+const undepleted_bit     = 0x02 # parse(UInt8, "00000010", base=2) # 0 -> depleted point; 1 -> undepleted point
+const pn_junction_bit    = 0x04 # parse(UInt8, "00000100", base=2) # 0 -> point is not part of pn-junction; 1 -> point is part of the pn-junction
+const bulk_bit           = 0x08 # parse(UInt8, "00001000", base=2) # 0 -> point is surrounded by points that do not belong to the pn-junction; 1 -> point is only surrounded by points in the pn-junction
+const inactive_layer_bit = 0x10 # parse(UInt8, "00010000", base=2) # 0 -> point is not part of the inactive layer; 1 -> point is part of the inactive layer
 
 is_pn_junction_point_type(p::PointType) = p & pn_junction_bit > 0
 is_undepleted_point_type(p::PointType) = p & undepleted_bit > 0
 is_fixed_point_type(p::PointType) = p & update_bit == 0
+is_in_inactive_layer(p::PointType) = p & inactive_layer_bit != 0
```
fhagemann: Maybe follow the convention of lines 32 and 33 to use `>` instead of `!=`
```suggestion
is_in_inactive_layer(p::PointType) = p & inactive_layer_bit > 0

```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -128,18 +131,19 @@ end
 This constant-lifetime-based charge trapping model is similar to the Boggs model, which is constant-mean-free-path based.
 
 The model implemented has two sets of parameters for the sensitive (bulk) and inactive (dead layer/surface layer) volume respectively.
-The inactive layer effect will only be considered if the corresponding `inactive_layer_geometry` is defined.
+The inactive layer effect will only be considered if the corresponding `inactive_layer_exist` is true.
```
fhagemann: This is a variable defined in the code and nothing that the user can set, so I would remove this from the docstring or rephrase.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -148,21 +152,33 @@ struct ConstantLifetimeChargeTrappingModel{T <: SSDFloat, G <: Union{<:AbstractG
     inactive_layer_geometry::G
 end
 
+has_inactive_layer(point_types::Array{PointType, 3}) = any(val -> (val & inactive_layer_bit) != 0, point_types)
```
fhagemann: You can use the `is_in_inactive_layer` you defined in `src/ScalarPotentials/PointTypes.jl` here:
```suggestion
has_inactive_layer(point_types::Array{PointType, 3}) = any(is_in_inactive_layer, point_types)
```

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,97 @@ end
...
+    end
+
+    # dim 2 (vary j, fix i & k)
+    for i in 1:sz1, k in 1:sz3
+        vec = @view point_types[i, :, k]
+        matches = get_inactivelayer_indices(vec)
+        for match in matches, idx in match
+            vec[idx] |= inactive_layer_bit
+        end
+    end
```
fhagemann, line 318: Does this also work on cylindrical grids that are not phi-symmetric?

src/ScalarPotentials/PointTypes.jl
```diff
@@ -228,4 +231,97 @@ end
...
+    
+    return pt_indices
+end
+"""
+    mark_inactivelayer_point_types!(point_types::Array{PointType, 3})
+
+Applies the bitwise marking rule to slices across all three dimensions of
+a 3D array, modifying it in-place.
+"""
+function mark_inactivelayer_point_types!(point_types::Array{PointType, 3})
```
fhagemann: Are there any geometries or cases in which you expect this to not work?
Maybe something like the contact being super thin or the SSD simulation world being of the same size as the detector?

src/SignalGeneration/SignalGeneration.jl
```diff
@@ -22,10 +22,11 @@ function add_signal!(
...
         wpot::Interpolations.Extrapolation{T, 3}, 
         S::CoordinateSystemType,
-        ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}()
-    )::Nothing where {T <: SSDFloat}
+        ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}(),
+        point_types::Union{PointTypes{T, N, CS}, Nothing} = nothing
+    )::Nothing where {T <: SSDFloat, N, CS}
 
-    tmp_signal::Vector{T} = _calculate_signal(ctm, path, pathtimestamps, charge, wpot, S)
+    tmp_signal::Vector{T} = _calculate_signal(ctm, path, pathtimestamps, charge, wpot, S, point_types)
```
fhagemann: How about **ALWAYS** requiring to pass the point types (instead of `nothing`)?
(if `ctm.inactive_layer_geometry` is defined, we will just not look at point types)

You might then also be able to avoid passing `S` as it should be the same `CS` in point_types.

src/SignalGeneration/SignalGeneration.jl
```diff
@@ -38,16 +39,16 @@ end
...
 
 function add_signal!(signal::AbstractVector{T}, timestamps::AbstractVector{<:RealQuantity}, paths::Vector{<:EHDriftPath{T}}, charges::Vector{T}, 
-        wpot::Interpolations.Extrapolation{T, 3}, S::CoordinateSystemType, ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}())::Nothing where {T <: SSDFloat}
+        wpot::Interpolations.Extrapolation{T, 3}, S::CoordinateSystemType, ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}(), point_types::Union{PointTypes{T, N, CS}, Nothing} = nothing)::Nothing where {T <: SSDFloat, N, CS}
     for ipath in eachindex(paths)
-        add_signal!(signal, timestamps, paths[ipath], charges[ipath], wpot, S, ctm)
+        add_signal!(signal, timestamps, paths[ipath], charges[ipath], wpot, S, ctm, point_types)
     end
     nothing
 end
```
fhagemann: Same thing here: **ALWAYS** pass the point types, and just "ignore" them if not needed?
Good thing is then that passing `S` would no longer be needed.

src/SolidStateDetectors.jl
```diff
@@ -71,6 +71,7 @@ export calculate_electric_potential!, calculate_weighting_potential!, calculate_
 export ElectricFieldChargeDriftModel, ADLChargeDriftModel, IsotropicChargeDriftModel, InactiveLayerChargeDriftModel
 export LinearImpurityDensity, LinBouleImpurityDensity, LinExpBouleImpurityDensity, ThermalDiffusionLithiumDensity, PtypePNJunctionImpurityDensity
 export NoChargeTrappingModel, BoggsChargeTrappingModel, ConstantLifetimeChargeTrappingModel
+export mark_inactivelayer_point_types!
```
fhagemann: Do we want to export this?
I would say this makes only sense, if we expect the user to use it, e.g. when throwing a warning that this has to be done. If this is always done by SSD "under the hood" (like `mark_bulk_bits!`), I think we can be fine with not exporting this.

src/SignalGeneration/SignalGeneration.jl
```diff
@@ -22,10 +22,11 @@ function add_signal!(
         charge::T,          
         wpot::Interpolations.Extrapolation{T, 3}, 
         S::CoordinateSystemType,
-        ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}()
-    )::Nothing where {T <: SSDFloat}
+        ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}(),
+        point_types::Union{PointTypes{T, N, CS}, Nothing} = nothing
+    )::Nothing where {T <: SSDFloat, N, CS}
```
fhagemann: How about **ALWAYS** requiring to pass the point types (instead of `nothing`)?
(if `ctm.inactive_layer_geometry` is defined, we will just not look at point types)

And I also feel like `PointTypes` should have the same `CoordinateSystemType` as `S`?
We might even be able to get rid of `S`, as we can infer it from the `PointTypes` now.
```suggestion
        point_types::PointTypes{T, N, S},
        ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}()
    )::Nothing where {T <: SSDFloat, N, S}
```

If `add_signal!` is not exported, I think we could make this breaking change.

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: Can we then add this add the end of the electric potential calculation (maybe optional with a keyword argument) or ask the user to explicitly run `mark_inactivelayer_point_types!` once to avoid calling this every time we simulate an event.

Maybe @oschulz also has thoughts on this?

claudiaalvgar: But this function is only executed in the charge trapping model for the inactive layer, not every time you simulate an event

fhagemann: This function is executed in `_calculate_signal`.

When you simulate an `Event` you
- run `simulate!(evt, sim)`
  https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/Event/Event.jl#L279-L283
- which calls `get_signals!`
  https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/Event/Event.jl#L237-L249
- which calls `get_signal`
  https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/Simulation/Simulation.jl#L1210-L1218
- which calls `add_signal!`
  https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/SignalGeneration/SignalGeneration.jl#L17-L37
- which calls `_calculate_signal`
  https://github.com/claudiaalvgar/SolidStateDetectors.jl/blob/976937b686079966d61c3e4985abb76f6d90915c/src/ChargeTrapping/ChargeTrapping.jl#L157-L180
- which calls `mark_inactivelayer_point_types!`

So, this is called every time you simulate an event with this charge trapping model.

oschulz: > Can we then add this add the end of the electric potential calculation

Yes, I think that would be best. Electric potential calculation has to run before drift, so we can be sure it's done. And it really shouldn't be done several times. Can we just do it automatically, depending on whether the config includes a dead layer? Otherwise users may forget to set the kwarg.

src/ChargeTrapping/ChargeTrapping.jl
oschulz: I think this is probably the best way for now, `ifelse` would be quite an overhead. It should be type-stable like it is, `τi` is set to either `τ_inactive` or `τ` in both code paths. @claudiaalvgar is it guaranteed that those have the same type?

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I implemented it this way to minimise interference with the existing code, since I’m not sure whether increasing the point types would affect any of the calculations prior to the charge-trapping model. I also wanted to avoid making the code search for an inactive layer when it’s clear it’s not going to have one. Where would you suggest placing it instead?

oschulz: >  I’m not sure whether increasing the point types would affect any of the calculations prior to the charge-trapping model.

We should check that's that not the case (resp. fix it if it does). Adding dead-layer point types later shouldn't prevent field recalculation etc.

> Where would you suggest placing it instead?

I think it should happen when the other point types are "painted" on the grid. It's just another overlay, after all. Besides, field calculation need to know about it anyway, right?

fhagemann: To my knowledge, all functionalities should only access the bits they are interested in, using the convenience functions that we define in `src/ScalarPotentials/PointTypes.jl`:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/ScalarPotentials/PointTypes.jl#L25-L32

I would therefore be **very** surprised if this would break something (but doesn't harm to check via the tests).

Also: if there is no undepleted area, I guess there is also no inactive layer, so calculating the inactive-layer bits could be linked to either if `depletion_handling` is true. Or after marking all the undepleted bits and seeing if there are any.

fhagemann: > I think it should happen when the other point types are "painted" on the grid. It's just another overlay, after all. Besides, field calculation need to know about it anyway, right?

The field calculation only cares about its depletion level which is another bit (`0x02`). The additional step done here is just to check if any undepleted bit (`0x07` or `0x0f` = in the semiconductor, undepleted and at a fixed potential, `7` or `f` depending on whether it's close to a surface or not) is also part of the inactive layer or part of regular "bulk undepletion"

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: I was thinking more about something like
```suggestion
        τi = get_charge_carrier_lifetime(path[i], ctm.inactive_layer_geometry, point_types)
```
and depending on whether `ctm.inactive_layer_geometry` is `nothing` or `AbstractGeometry`, calling different functions here.

```julia
function is_in_inactive_layer(pt::CartesianPoint{T}, g::AbstractGeometry{T}, point_types::PointTypes{T})
    in(path[i], ctm.inactive_layer_geometry)) ? τ_inactive : τ
end

function is_in_inactive_layer(pt::CartesianPoint{T}, ::Nothing, point_types::PointTypes{T})
    is_in_inactivelayer(point_types[pt])
end
```

fhagemann: This might also then get rid of the need for `user_defined_inactive_layer` (and maybe even `inactive_layer_exist`)

src/SignalGeneration/SignalGeneration.jl
claudiaalvgar: The reason I initialised the point types to nothing is because ```add_signal``` is also used in ```MCEventsProcessing.jl``` in ```_generate_waveform``` that will also need to know about point types and also ```simulate_waveforms```, should I still make this change?

src/SolidStateDetectors.jl
claudiaalvgar: Yes, this function is not really necessary. I just added it to get a quick plotting of the point types without having to run the Charge Trapping model, but with the new change of marking the inactive layer bits after the potential calculation it shouldn't be required anymore

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Yes, this is a purely array-based algorithm so it doesn’t know about coordinate systems, it will work on grids that are not phi-symmetric as long as point point types are 3D arrays

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: This function will work as long as the inactive layer (0x07 and 0x0f) point types are surrounded by 0x00 (the contact outside the detector) and a depleted area inside the detector. The only case it won't distinguish the inactive layer is for detectors with undepleted regions touching the inactive layer. In this extreme case, defining the inactive layer geometry boundaries via inactive_layer_geometry would be the best way

src/SolidStateDetectors.jl
fhagemann: Then, I would not export it

src/SignalGeneration/SignalGeneration.jl
fhagemann: This is fine. We do not modify the input variables of the exported function itself, just how it is internally handled.

src/ScalarPotentials/PointTypes.jl
fhagemann: This might be an issue for n-type detectors, but I think this is fine. 
One could think about also looking at the sign and magnitude of the impurity density, but we should not worry about this for now.

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I moved ```mark_inactivelayer_point_types``` to ```Simulation.jl``` under the condition that surface_imp_model is defined

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: Done

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I added this function ```get_charge_carrier_lifetime``` on the new code

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: Done

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: Done

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: Done

src/SolidStateDetectors.jl
claudiaalvgar: Done

src/SignalGeneration/SignalGeneration.jl
claudiaalvgar: Okay, I modified this

src/MCEventsProcessing/MCEventsProcessing.jl
```diff
@@ -80,7 +80,7 @@ function simulate_waveforms( mcevents::TypedTables.Table, sim::Simulation{T};
     @info "Generating waveforms..."
     waveforms = map( 
         wpot ->  map( 
-            x -> _generate_waveform(x.dps, to_internal_units.(x.edeps), Δt, Δtime, wpot, S, unitless_energy_to_charge, ctm),
```
fhagemann, line 83: Maybe here, we can keep the similar order
(`sim.point_types` where `S` was before)

src/MCEventsProcessing/MCEventsProcessing.jl
```diff
@@ -147,12 +147,12 @@ end
 
 
 function _generate_waveform( drift_paths::Vector{<:EHDriftPath{T}}, charges::Vector{<:SSDFloat}, Δt::RealQuantity, dt::T,
-                             wpot::Interpolations.Extrapolation{T, 3}, S::CoordinateSystemType, unitless_energy_to_charge,
```
fhagemann, line 150: Same thing here.

src/Simulation/Simulation.jl
```diff
@@ -1015,6 +1015,10 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
     if isEP mark_bulk_bits!(sim.point_types.data) end
     if depletion_handling && isEP
         mark_undep_bits!(sim.point_types.data, sim.imp_scale.data)
+        
+        if isdefined(sim.detector.semiconductor.impurity_density_model, :surface_imp_model)
+            mark_inactivelayer_point_types!(sim.point_types.data)
+        end
```
fhagemann: This is the most elegant solution is my opinion!

Just one thing to note when looking at this now:
we might wanna be consistent in the naming and call this
`mark_inactivelayer_bits!` in accordance to `mark_bulk_bits!` and `mark_undep_bits!`

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -169,13 +188,9 @@ function _calculate_signal(
...
-    inactive_layer_exist = !isnothing(ctm.inactive_layer_geometry)
 
-    if inactive_layer_exist
-        user_defined_inactive_layer_geom = true
-    else
-        mark_inactivelayer_point_types!(point_types.data)
+    inactive_layer_exist = !isnothing(ctm.inactive_layer_geometry)
+    if !inactive_layer_exist
         inactive_layer_exist = has_inactive_layer(point_types.data)
     end
```
fhagemann: This `if` statement can be avoided by just having this line (also more elegant in my opinion):
```julia
inactive_layer_exist = !isnothing(ctm.inactive_layer_geometry) || has_inactive_layer(point_types.data)
```

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: I like this, thanks!

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -190,16 +205,8 @@ function _calculate_signal(
...
-        if user_defined_inactive_layer_geom
-            τi = (inactive_layer_exist && in(path[i], ctm.inactive_layer_geometry)) ? τ_inactive : τ
-        else
-            path_in_pt_grid = find_closest_gridpoint(path[i], point_types.grid) #returns the grid point where the path is
-            idx1, idx2, idx3 = path_in_pt_grid
-            pt_value = point_types.data[idx1, idx2, idx3] #returns the point type value in the path 
-            τi = (inactive_layer_exist && is_in_inactive_layer(pt_value)) ? τ_inactive : τ
-        end
-        
+        τi = inactive_layer_exist ? get_charge_carrier_lifetime(path[i], ctm.inactive_layer_geometry, point_types, τ, τ_inactive) : τi
```
fhagemann: This can become one line
```julia
τi::T = inactive_layer_exist ? get_charge_carrier_lifetime(path[i], ctm.inactive_layer_geometry, point_types, τ, τ_inactive) : τ
```

src/SignalGeneration/SignalGeneration.jl
```diff
@@ -21,11 +21,11 @@ function add_signal!(
...
         charge::T,          
         wpot::Interpolations.Extrapolation{T, 3}, 
-        S::CoordinateSystemType,
+        point_types::PointTypes{T, N, S},
         ctm::AbstractChargeTrappingModel{T} = NoChargeTrappingModel{T}()
-    )::Nothing where {T <: SSDFloat}
+    )::Nothing where {T <: SSDFloat, N, S}
 
-    tmp_signal::Vector{T} = _calculate_signal(ctm, path, pathtimestamps, charge, wpot, S)
+    tmp_signal::Vector{T} = _calculate_signal(ctm, path, pathtimestamps, charge, wpot, S, point_types)
```
fhagemann: It looks like we are always passing `point_types` to `_calculate_signal`, so we might as well just get rid of passing `S` here as well:
```suggestion
    tmp_signal::Vector{T} = _calculate_signal(ctm, path, pathtimestamps, charge, wpot, point_types)
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -6,8 +6,9 @@ function _calculate_signal(
         pathtimestamps::AbstractVector{T}, 
         charge::T,          
         wpot::Interpolations.Extrapolation{T, 3}, 
-        S::CoordinateSystemType
-    ) where {T <: SSDFloat}
+        S::CoordinateSystemType,
+        point_types::Union{PointTypes{T, N, CS}, Nothing} = nothing
+    ) where {T <: SSDFloat, N, CS}
```
fhagemann: And then remove it here:
```suggestion
        point_types::PointTypes{T, N, S}
    ) where {T <: SSDFloat, N, S}
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -27,8 +28,9 @@ function _calculate_signal(
         pathtimestamps::AbstractVector{T}, 
         charge::T,          
         wpot::Interpolations.Extrapolation{T, 3}, 
-        S::CoordinateSystemType
-    )::Vector{T} where {T <: SSDFloat}
+        S::CoordinateSystemType,
+        point_types::Union{PointTypes{T, N, CS}, Nothing} = nothing
+    )::Vector{T} where {T <: SSDFloat, N, CS}
```
fhagemann: Here:
```suggestion
        point_types::PointTypes{T, N, S}
    )::Vector{T} where {T <: SSDFloat, N, S}
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -68,8 +70,9 @@ function _calculate_signal(
...
```
fhagemann: Here:
```suggestion
        point_types::PointTypes{T, N, S}
    )::Vector{T} where {T <: SSDFloat, N, S}
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -148,42 +151,67 @@ struct ConstantLifetimeChargeTrappingModel{T <: SSDFloat, G <: Union{<:AbstractG
...
         path::AbstractVector{CartesianPoint{T}}, 
         pathtimestamps::AbstractVector{T}, 
         charge::T,          
         wpot::Interpolations.Extrapolation{T, 3}, 
-        S::CoordinateSystemType
-    )::Vector{T} where {T <: SSDFloat}
-
+        S::CoordinateSystemType,
+        point_types::Union{PointTypes{T, N, CS}, Nothing} = nothing
+    )::Vector{T} where {T <: SSDFloat, N, CS}
```
fhagemann: and here:
```suggestion
        point_types::PointTypes{T, N, S}
    )::Vector{T} where {T <: SSDFloat, N, CS}
```

fhagemann (approved):
One final review from my side and then this should be good to be merged:

claudiaalvgar:
I was also thinking to add to ```src/SolidStateDetector/Semiconductor.jl``` an info message like :

```
if haskey(constant_lifetime_ctm_dict, "inactive_layer_geometry")
	@info "Using YAML-defined inactive layer geometry"
         constant_lifetime_ctm_dict["inactive_layer_geometry"] = Geometry(T, dict["charge_trapping_model"]["inactive_layer_geometry"], input_units, transformations)
else
	@info "Using point types to define inactive layer geometry"	        
end
```

So that we are sure it is using the yaml geometry or the point types but this gives me a test error, should we add some printing to know what geometry is using?

oschulz:
Nice!!

PR #525, opened 2025-08-21
Add support for simulation for dead layer - Step7: Support plotting full range of point types

fhagemann:
Follow up of #512: add keyword argument `all_point_types` (happy to discuss/change the name if needed) to plot the full range of point types (`0` to `31`) rather than the standard reduced `point_types .& 0x07` (`0` to `7`).

I could also envision changing the name to `reduced_point_types` (`true` = 0-7, `false` = 0-31), no strong opinion here.

Plotting the reduced point types from `0` to `7` is still the default.

Code example:
```julia
using Plots
using Unitful
using SolidStateDetectors

T = Float64
sim_undep = Simulation{T}(SSD_examples[:IVCIlayer])

# Partially depleted detector

sim_undep.detector = SolidStateDetector(sim_undep.detector, contact_id = 2, contact_potential = 500);

calculate_electric_potential!( sim_undep,
                               depletion_handling = true,
                               convergence_limit = 1e-6,
                               refinement_limits = [0.2, 0.1, 0.05, 0.01],
                               verbose = true)


plot( 
    plot(sim_undep.electric_potential),
    plot(sim_undep.point_types),
    plot(sim_undep.imp_scale),
    layout = (1, 3), size = (1500, 700)
)
```
image

```julia

plot( 
    plot(sim_undep.electric_potential),
    plot(sim_undep.point_types, all_point_types = true),
    plot(sim_undep.imp_scale),
    layout = (1, 3), size = (1500, 700)
)
```
image

### DIFF
--- a/src/PlotRecipes/Potentials.jl
+++ b/src/PlotRecipes/Potentials.jl
@@ -1,3 +1,5 @@
+_get_potential_plot_information(sp::ScalarPotential, ::Bool) = _get_potential_plot_information(sp) # ignore point type bool for all other potentials
+
 function _get_potential_plot_information(p::ElectricPotential)::Tuple{Symbol, Tuple{Quantity, Quantity}, String, Unitful.Units}
     return (:viridis, extrema(p.data).*internal_voltage_unit, "Electric Potential", internal_voltage_unit)
 end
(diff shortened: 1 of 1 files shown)


claudiaalvgar:
This looks good to me, both naming options make sense, so use whichever you prefer. Thanks for the update!

Issue #513, opened 2025-07-24
Geant4 extension isotope source opening angle not changing


JunwenDiao:
Hi,
it seems like the opening angle of the isotope source is not changing with the set up.
Here are the distributions of the positions of the first energy deposition of the events for a monoenergetic source (5MeV gamma) and an isotope source (207Bi), both with opening angle set to zero:

gamma:
<img width="646" height="616" alt="Image" src="https://github.com/user-attachments/assets/38d4e934-9d70-4ed9-87f3-f02ffbd65a1a" />

207Bi:
<img width="643" height="620" alt="Image" src="https://github.com/user-attachments/assets/b819f20a-6e01-499b-abc9-23e03f988535" />
@Yuan-Ru-Lin

fhagemann:
Might be related to https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/508?

JunwenDiao:
Yes!

Yuan-Ru-Lin:
I will take a look after JuliaHEP

PR #526, opened 2025-08-25
Fix bug in source emission in `Geant4` extension

fhagemann:
So, I think I found a fix for #513:

When using `G4SingleParticleSource` instead of `G4ParticleGun`, the position, direction and energy of the source have to be defined a slightly different way (namely using distributions rather than fixed values).

Also, it looks like `G4SingleParticleSource` ignores `SetParticleMomentumDirection`. Thus, when setting a `MinTheta` and `MaxTheta`, this is always with respect to the negative z-axis (see page 26 in the [Geant4 Book for Application Developers](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/fo/BookForApplicationDevelopers.pdf)):

Screenshot from 2025-08-25 04-03-40

This requires to actually define the coordinate system using `angref1` and `angref2`.

I refactored the implementation of the `_init` and `_gen` methods for `MonoenergeticSource` and `IsotopeSource` to be based on `G4SingleParticleSource` and moved the definition of the source emission (line, conic, isotropic) from `_gen` to `_init`.

In order to check that the position and the momentum of the primary vertex is computed correctly, I tested the new code with a modified `endeventaction` to verify the position and momentum direction of every primary vertex:
```julia
function endeventaction(evt::G4Event, app::G4JLApplication)
    direction = evt |> (evt -> GetPrimaryVertex(evt, 0)) |> (vertex -> GetPrimary(vertex, 0)) |> GetMomentumDirection
    pos = evt |> (evt -> GetPrimaryVertex(evt, 0)) |> GetPosition
    d = [x(direction), y(direction), z(direction)]
    @info [x(pos), y(pos), z(pos)], normalize(d)
    return
end
```

I also performed some visual tests using the "ICPC in cryostat" example:

### MonoenergeticSource

```julia
sim = Simulation{T}(SSD_examples[:InvertedCoaxInCryostat])
source = MonoenergeticSource("gamma", 2.615u"MeV", CartesianPoint(0.05,0,0.05), CartesianVector(-1,0,0), 10u"°")
app = G4JLApplication(sim, source)
evts = run_geant4_simulation(app, 5_000)
```
The event distribution is still as expected:
image

### IsotopeSource
In this case a 228Th source, which also emits electrons - hence, different results are achieved depending on whether the source is placed inside or outside of the cryostat

```julia
sim = Simulation{T}(SSD_examples[:InvertedCoaxInCryostat])
source = IsotopeSource(90, 228, 0.0, 0.0, CartesianPoint(0.05,0,0.05), CartesianVector(-1,0,0), 10u"°") # inside cryostat
app = G4JLApplication(sim, source)
evts = run_geant4_simulation(app, 5_000)
```
All hits:
image

Just the first hit in the detector of each event:
image

```julia
sim = Simulation{T}(SSD_examples[:InvertedCoaxInCryostat])
source = IsotopeSource(90, 228, 0.0, 0.0, CartesianPoint(0.06,0,0.05), CartesianVector(-1,0,0), 10u"°") # outside cryostat
app = G4JLApplication(sim, source)
evts = run_geant4_simulation(app, 5_000)
```
All hits:
outside cryostat

Just the first hit in the detector of each event:
image

### MonoenergeticSource
Monoenergetic gammas with an energy of 2.615MeV:
image

### IsotopeSource
A point-like 228Th source placed outside of the cryostat:
image

### DIFF
--- a/ext/Geant4/g4jl_application.jl
+++ b/ext/Geant4/g4jl_application.jl
@@ -56,8 +56,11 @@ function endeventaction(evt::G4Event, app::G4JLApplication)
 end
 
 
+# We are using G4SingleParticleSource instead of G4ParticleGun for uniformity 
+# across MonoenergeticSource and IsotopeSource, see discussion
+# https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/526
 @with_kw mutable struct GeneratorData <: G4JLGeneratorData
-    gun::Union{Nothing, CxxPtr{G4ParticleGun}, CxxPtr{G4SingleParticleSource}} = nothing
+    gun::Union{Nothing, CxxPtr{G4SingleParticleSource}} = nothing
     particle::Union{Nothing, CxxPtr{G4ParticleDefinition}} = nothing
     position::G4ThreeVector = G4ThreeVector(0, 0, 0)
     direction::G4ThreeVector = G4ThreeVector(0, 0, 0)
@@ -75,7 +78,7 @@ function SSDGenerator(source::SolidStateDetectors.MonoenergeticSource;  kwargs..
         SetParticleDefinition(data.gun, particle)
 
         # Place the SingleParticleSource as point source
-            data.position = G4ThreeVector(
+        data.position = G4ThreeVector(
             source.position.x * Geant4.SystemOfUnits.meter,
             source.position.y * Geant4.SystemOfUnits.meter,
             source.position.z * Geant4.SystemOfUnits.meter
@@ -140,7 +143,8 @@ function SSDGenerator(source::SolidStateDetectors.IsotopeSource; kwargs...)
         SetMinTheta(angdist, 0.0)
         SetMaxTheta(angdist, ustrip(NoUnits, source.opening_angle))  # convert to radians
 
-        # SetMonoEnergy(GetEneDist(data.gun), 0.0 * Geant4.SystemOfUnits.MeV)
+        # Set kinetic energy of the decaying isotope to zero ("decay at rest")
+        SetMonoEnergy(GetEneDist(data.gun), 0.0 * Geant4.SystemOfUnits.MeV)
     end
 
     function _gen(evt::G4Event, data::GeneratorData)::Nothing
--- a/ext/Geant4/g4jl_application.jl
+++ b/ext/Geant4/g4jl_application.jl
@@ -13,17 +13,17 @@ struct SSDPhysics <: G4VUserPhysicsList
 end
 
 struct SDData{T} <: G4JLSDData
-  detectorHits::DetectorHits
-  SDData{T}() where {T} = new(DetectorHits((evtno = Int32[], detno = Int32[], thit = typeof(zero(T)*u"s")[], edep = typeof(zero(T)*u"keV")[], pos = SVector{3,typeof(zero(T)*u"mm")}[])))
+    detectorHits::DetectorHits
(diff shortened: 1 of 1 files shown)


ext/Geant4/g4jl_application.jl
```diff
@@ -66,31 +70,39 @@ function SSDGenerator(source::SolidStateDetectors.MonoenergeticSource;  kwargs..
...
 
     function _init(data::GeneratorData, ::Any)
-        gun = data.gun = move!(G4ParticleGun())
+        data.gun = move!(G4SingleParticleSource())
         particle = data.particle = FindParticle(source.particle_type)
-        data.position = G4ThreeVector(
+        SetParticleDefinition(data.gun, particle)
+
+        # Place the SingleParticleSource as point source
+            data.position = G4ThreeVector(
```
Yuan-Ru-Lin: Would be nice if this line is unindented by one level

ext/Geant4/g4jl_application.jl
```diff
@@ -103,16 +115,32 @@ function SSDGenerator(source::SolidStateDetectors.IsotopeSource; kwargs...)
...
+        data.direction = G4ThreeVector(d.x, d.y, d.z)
+
+        angdist = GetAngDist(data.gun)
+        SetAngDistType(angdist, "iso")
+        DefineAngRefAxes(angdist, "angref1", G4ThreeVector(a...))
+        DefineAngRefAxes(angdist, "angref2", G4ThreeVector(b...))
+        SetMinTheta(angdist, 0.0)
+        SetMaxTheta(angdist, ustrip(NoUnits, source.opening_angle))  # convert to radians
+
+        # SetMonoEnergy(GetEneDist(data.gun), 0.0 * Geant4.SystemOfUnits.MeV)
```
Yuan-Ru-Lin: Would the energy of isotopes be 0 if you don't run this line? That is, is `G4SingleParticleSource` by default not moving?

Yuan-Ru-Lin (commented):
General feedback

Yuan-Ru-Lin:
I like the idea of using `G4SingleParticleSource ` for both types of sources. Now their `_init`s differ only by one line, and this line

https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a28411a9f015f6550887ca9ce9f3002589f3a747/ext/Geant4/g4jl_application.jl#L60

can be reduced to just

```julia
    gun::Union{Nothing, CxxPtr{G4SingleParticleSource}} = nothing
```

I do think it'd be nice to leave the link to this PR in the code to elucidate this decision to the future developers, especially above this line

https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a28411a9f015f6550887ca9ce9f3002589f3a747/ext/Geant4/g4jl_application.jl#L73

with comments like "We are using `G4SingleParticleSource` instead of `G4ParticleGun` for uniformity across `MonoenergeticSource` and `IsotopeSource`".

I also left a minor question and a reminder for coding style above. Once those are resolved and comments like the one above are added in the code, this PR will be ready to be accepted. 

Thanks for this work @fhagemann!

ext/Geant4/g4jl_application.jl
fhagemann: I don't know and I might just have forgotten that I commented this line to see what the influence would be.
I uncommented this line

ext/Geant4/g4jl_application.jl
fhagemann: Done!

fhagemann:
> Once those are resolved and comments like the one above are added in the code, this PR will be ready to be accepted.

@Yuan-Ru-Lin Thanks for the comments, I should have incorporated them all!

PR #532, opened 2025-09-03
Add support for simulation for dead layer - Step8: Refactor combining charge trapping models for bulk and inactive layer

claudiaalvgar:
The functionality of running different charge trapping models (CTM) inside the bulk and inside the inactive layer has been added. The combinations supported are:
* Bulk: Boggs CTM, inactive layer: Constant lifetime CTM
* Bulk: Boggs CTM, inactive layer: No CTM
* Bulk: No CTM, inactive layer: Constant lifetime CTM
* Bulk: No CTM, inactive layer: No CTM
* Bulk: Constant lifetime CTM, inactive layer: Constant lifetime CTM
* Bulk: Constant lifetime CTM, inactive layer: No CTM


examples/example_config_files/ivc_inactivelayer.yaml
```diff
@@ -40,33 +40,39 @@ detectors:
...
+         nσe-1: 1020cm
+         nσh-1: 2040cm
+      #model: ConstantLifetime
+      #parameters:
+      #   τh: 1ms
+      #   τe: 1ms
+      model_inactive: ConstantLifetime
+      parameters_inactive:
+         τh_inactive: 1μs
+         τe_inactive: 1μs
```
fhagemann: If we include `inactive` in the `parameters` field name, I don't see why we would need it to `τe` and `τh` also, because it feels a bit redundant:
```suggestion
      model_inactive: ConstantLifetime
      parameters_inactive:
         τh: 1μs
         τe: 1μs
```

examples/example_config_files/ivc_inactivelayer.yaml
```diff
@@ -185,4 +191,4 @@ surroundings:
           - -0.05
     charge_density:
       name: constant
-      value: 0C/m^3
+      value: 0C/m^3 # Reasonable density: +2e-11 # => 2*10⁻¹¹ C/m³
```
fhagemann: Why these changes?

examples/example_config_files/public_TrueCoaxial.yaml
```diff
@@ -55,6 +55,8 @@ detectors:
       parameters:
         τh: 1ms
         τe: 1ms
+      model_inactive: ConstantLifetime
+      parameters_inactive:
         τh_inactive: 1μs
         τe_inactive: 1μs
```
fhagemann: Again here, I would just call it `τe` and `τh` without the `_inactive` (as this is already in the `parameters_inactive` field name.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -1,6 +1,7 @@
 abstract type AbstractChargeTrappingModel{T <: SSDFloat} end
 
 function _calculate_signal(
```
fhagemann: For backwards compatibility, I would vote for adding a method that makes sure the old method still works, something like:

```julia
_calculate_signal(ctm::AbstractChargeTrappingModel{T},  args...) where {T <: SSDFloat} = _calculate_signal(ctm, NoChargeTrappingModel{T}, args...)
```
(not 100% sure whether this works or whether this might end up in a `StackOverflowError` and one might need to explicitly define what `args...` is)

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -51,40 +123,93 @@ Charge trapping model presented in [Boggs _et al._ (2023)](https://doi.org/10.10
 * `nσe::T`: Trapping product for electrons (default: `(nσe)^-1 = 1020cm`).
 * `nσh::T`: Trapping product for holes (default: `(nσh)^-1 = 2040cm`).
 * `temperature::T`: Temperature of the crystal (default: `78K`).
+* `inactive_layer_geometry`: The geometry of the inactive layer. If it's not defined, it uses point_types to get the inactive layer geometry, else it uses the user-defined inactive_layer_geometry.
```
fhagemann: I'm a bit confused by this because `BoggsChargeTrappingModel` can also just be applied to the bulk (?)

src/SolidStateDetector/Semiconductor.jl
```diff
@@ -89,23 +92,33 @@ function Semiconductor{T}(dict::AbstractDict, input_units::NamedTuple, outer_tra
...
-            constant_lifetime_ctm_dict["inactive_layer_geometry"] = Geometry(T, dict["charge_trapping_model"]["inactive_layer_geometry"], input_units, transformations)
-        end
+
+    constant_lifetime_ctm_dict = haskey(dict, "charge_trapping_model") ? deepcopy(dict["charge_trapping_model"]) : Dict{String, Any}()
+
+    if haskey(constant_lifetime_ctm_dict, "inactive_layer_geometry")
+        constant_lifetime_ctm_dict["inactive_layer_geometry"] = Geometry(T, constant_lifetime_ctm_dict["inactive_layer_geometry"], input_units, transformations)
+    end
+
+    charge_trapping_model_inactive = if haskey(constant_lifetime_ctm_dict, "model_inactive") &&
```
fhagemann: Somewhere here: if it has the key `model_inactive` and `model`, create a combined one from the two.

docs/src/man/charge_drift.md
```diff
@@ -235,12 +235,13 @@ The second step would be to define a method for `_calculate_signal`, which retur
 using SolidStateDetectors: CoordinateSystemType, SSDFloat
 using Interpolations
 function SolidStateDetectors._calculate_signal( 
-        ctm::CustomChargeTrappingModel{T}, 
+        ctm::CustomChargeTrappingModel{T},
+        ctmil::AbstractChargeTrappingModel{T},
```
fhagemann: I guess this change in the documentation is then no longer valid, as they would both be passed as a `ctm::CombinedChargeTrappingModel` ?

docs/src/man/charge_drift.md
```diff
@@ -363,13 +361,74 @@ transformations = combine_transformations(inner_transformations, outer_transform
 inactive_layer_geometry = Geometry(T, inactive_layer_geometry_dict, sim.input_units, transformations)
 ## or simply use the already constructed geometry if the `inactive_layer_geometry ` geometry is defined in the configuration file
 # inactive_layer_geometry = sim.detector.semiconductor.charge_trapping_model.inactive_layer_geometry
-parameters = Dict("parameters" => Dict("τh" => τh, "τe" => τe, "τh_inactive" => τh_inactive, "τe_inactive" => τe_inactive), "inactive_layer_geometry" => inactive_layer_geometry)
+parameters = Dict("parameters" => Dict("τh" => τh, "τe" => τe), "inactive_layer_geometry" => inactive_layer_geometry)
```
fhagemann: Do we still have `inactive_layer_geometry` here?

docs/src/man/charge_drift.md
```diff
@@ -363,13 +361,74 @@ transformations = combine_transformations(inner_transformations, outer_transform
...
+```
+* Bulk: Boggs CTM, inactive layer: No CTM
+```yaml
+charge_trapping_model:
+      model: Boggs
+      parameters:
+        nσe-1: 1020cm
+        nσh-1: 2040cm
+      model_inactive:
+```
```
fhagemann: General thought: it makes no physical sense to have trapping in the bulk but not in the inactive layer... I feel like the only place where we should allow `NoChargeTrappingModel` would be in the bulk. `NoChargeTrappingModel` in the inactive layer only makes sense if we also simulate the bulk without charge trapping, which is when `charge_trapping_model` can just not be defined in the config file.

docs/src/man/charge_drift.md
```diff
@@ -363,13 +361,74 @@ transformations = combine_transformations(inner_transformations, outer_transform
...
+* Bulk: Constant lifetime CTM, inactive layer: No CTM
+```yaml
+charge_trapping_model:
+      model: ConstantLifetime
+      parameters:
+        τh: 1ms
+        τe: 1ms
+      model_inactive:
+
+```
```
fhagemann: This I would also not allow.

docs/src/man/charge_drift.md
```diff
@@ -363,13 +361,74 @@ transformations = combine_transformations(inner_transformations, outer_transform
...
+* Bulk: No CTM, inactive layer: Constant lifetime CTM
+```yaml
+charge_trapping_model:
+      model:
+      model_inactive: ConstantLifetime
+      parameters_inactive:
+        τh: 1μs
+        τe: 1μs
+
+```
```
fhagemann: Does this work also, when `model` is not defined at all?
That might be the preferred way I guess (?)

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -54,37 +63,53 @@ Charge trapping model presented in [Boggs _et al._ (2023)](https://doi.org/10.10
...
+    Δl::T,
+    q::Base.RefValue{T},
+    w::T,
+) where {T <: SSDFloat}
+
+    Δq::T = q[] * nσ * Δl
+    q[] -= Δq
+    running_sum[] += w * Δq
+    return running_sum[] + w * q[]
+end
```
fhagemann: Any reason why this isn't in `_calculate_signal`?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -54,37 +63,53 @@ Charge trapping model presented in [Boggs _et al._ (2023)](https://doi.org/10.10
...
-    q::T = charge
+    q = Ref(charge)
+    running_sum = Ref(zero(T))
+    
+    vth::T = sqrt(3 * kB * ctm.temperature / (ifelse(charge > 0, ctm.meffh, ctm.meffe) * me)) # in m/s
+    nσ::T = ifelse(charge > 0, ctm.nσh, ctm.nσe)
+    
     @inbounds for i in eachindex(tmp_signal)
+        
+        Δt::T = (i > 1) ? (pathtimestamps[i] - pathtimestamps[i-1]) : zero(T)
```
fhagemann: Is this used at all?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -120,58 +145,40 @@ function BoggsChargeTrappingModel{T}(config_dict::AbstractDict = Dict(); tempera
...
+    Δt::T,
+    q::Base.RefValue{T},
+    w::T,
+) where {T <: SSDFloat}
+
+    Δq::T = q[] * Δt / τ
+    q[] -= Δq
+    running_sum[] += w * Δq
+    return running_sum[] + w * q[]
 end
```
fhagemann: Make this part of `_calculate_signal!`?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +213,198 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+* `inactive_charge_trapping_model::AbstractChargeTrappingModel{T}`: CTM inside the inactive layer.
+* `inactive_layer_geometry::G`: Inactive layer geometry.
+
+Combinations supported: 
+Bulk: Boggs CTM, inactive layer: Constant lifetime CTM
+Bulk: Boggs CTM, inactive layer: No CTM
+Bulk: No CTM, inactive layer: Constant lifetime CTM
+Bulk: No CTM, inactive layer: No CTM
+Bulk: Constant lifetime CTM, inactive layer: Constant lifetime CTM
+Bulk: Constant lifetime CTM, inactive layer: No CTM
```
fhagemann: I'm not sure if we need to list them all in the docstring. We might forget to add documentation here when adding new charge trapping models.

Again: I would consider not listing "No CTM" in the inactive layer when having a CTM in the bulk.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +213,198 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+    in(pt, g)
+end
+
+function in_inactive_layer(
+    pt::CartesianPoint{T},
+    ::Nothing,
+    point_types::PointTypes{T}
+    )where {T}
+    is_in_inactive_layer(point_types[pt])
+end
```
fhagemann: Don't these functions already exist (for the charge drift models)?
Maybe it might make sense to move them to `src/ScalarPotentials/PointTypes.jl`?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +213,198 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+
+struct CombinedChargeTrappingModel{T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}} <: AbstractChargeTrappingModel{T}
+    bulk_charge_trapping_model::AbstractChargeTrappingModel{T}
+    inactive_charge_trapping_model::AbstractChargeTrappingModel{T}
+    inactive_layer_geometry::G
+end
```
fhagemann: Should we include the types of the charge trapping models in the parametric types?
```suggestion
struct CombinedChargeTrappingModel{T <: SSDFloat, BCTM <: AbstractChargeTrappingModel{T}, ICTM <: AbstractChargeTrappingModel{T}, G <: Union{<:AbstractGeometry, Nothing}} <: AbstractChargeTrappingModel{T}
    bulk_charge_trapping_model::BCTM
    inactive_charge_trapping_model::ICTM
    inactive_layer_geometry::G
end
```
This might improve type stability

src/SolidStateDetector/Semiconductor.jl
```diff
@@ -20,7 +20,7 @@ This is the volume in which electrons and holes will drift during the signal dev
 * `material::MT`: Material of the semiconductor.
 * `impurity_density_model::IDM`: Impurity density model for the points inside the semiconductor.
 * `charge_drift_model::CDM`: Model that describes the drift of electrons and holes inside the semiconductor.
-* `charge_trapping_model::CTM`: Model that describes the trapping of electrons and holes inside the semiconductor.
+* `charge_trapping_model::CTM`: Model that describes the trapping of electrons and holes inside the semiconductor bulk.
```
fhagemann: Doesn't have to be bulk anymore.

src/SolidStateDetector/Semiconductor.jl
```diff
@@ -114,7 +118,7 @@ function println(io::IO, d::Semiconductor{T}) where {T <: SSDFloat}
     println("\t-Impurity Density Model: $(typeof(d.impurity_density_model).name.name)")
     println("\t-Charge Drift Model:\t $(typeof(d.charge_drift_model).name.name)")
     if !(d.charge_trapping_model isa NoChargeTrappingModel)
-        println("\t-Charge Trapping Model:\t $(typeof(d.charge_trapping_model).name.name)")
+        println("\t-Charge Trapping Model in bulk:\t $(typeof(d.charge_trapping_model).name.name)")
```
fhagemann: Also revert adding the `bulk` part here.

src/SolidStateDetector/Semiconductor.jl
```diff
@@ -89,18 +89,22 @@ function Semiconductor{T}(dict::AbstractDict, input_units::NamedTuple, outer_tra
...
-        BoggsChargeTrappingModel{T}(dict["charge_trapping_model"], temperature = temperature)
-    elseif haskey(dict, "charge_trapping_model") &&
-                               haskey(dict["charge_trapping_model"], "model") &&
-                               dict["charge_trapping_model"]["model"] == "ConstantLifetime"
-        constant_lifetime_ctm_dict = deepcopy(dict["charge_trapping_model"])
-        if haskey(constant_lifetime_ctm_dict, "inactive_layer_geometry")
-            constant_lifetime_ctm_dict["inactive_layer_geometry"] = Geometry(T, dict["charge_trapping_model"]["inactive_layer_geometry"], input_units, transformations)
-        end
+
+    constant_lifetime_ctm_dict = haskey(dict, "charge_trapping_model") ? deepcopy(dict["charge_trapping_model"]) : Dict{String, Any}()
```
fhagemann: It's a bit misleading that this variable is called `constant_lifetime_ctm_dict`, when it could be any charge trapping mode. I liked the previous `charge_trapping_model` better.

docs/src/man/charge_drift.md
claudiaalvgar: No, I forgot to change this part, I fixed the documentation now

docs/src/man/charge_drift.md
claudiaalvgar: Although it's not physical, I think it could be useful to disable the inactive layer charge trapping model, not to make physical models but to do tests like study the effect of the charge trapping model with the inactive layer vs. no inactive layer charge trapping effect. If you don't use the combined model, the other models would apply the charge trapping to the whole volume (including inactive layer) so it could be useful that this combined model allows to disable its effect

docs/src/man/charge_drift.md
claudiaalvgar: No, since the condition in ```Semiconductor.jl``` is:
```
charge_trapping_model = if haskey(ctm_dict, "model") && !haskey(ctm_dict, "model_inactive") && ctm_dict["model"] == "Boggs"
        BoggsChargeTrappingModel{T}(ctm_dict, temperature = temperature)

    elseif haskey(ctm_dict, "model") && !haskey(ctm_dict, "model_inactive") && ctm_dict["model"] == "ConstantLifetime"
        ConstantLifetimeChargeTrappingModel{T}(ctm_dict)

    elseif haskey(ctm_dict, "model") && haskey(ctm_dict, "model_inactive")
        CombinedChargeTrappingModel{T}(ctm_dict, temperature = temperature)

    else
        NoChargeTrappingModel{T}()
    end
```

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: ```_calculate_signal``` is also defined. The reason this part has to be separate is because it needs to be called also from the Combined Charge Trapping Model  ```_calculate_signal``` . By factoring it out into _Boggs_signal!, we avoid duplicating that logic

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: No, sorry, I deleted this now

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I changed this

src/SolidStateDetector/Semiconductor.jl
claudiaalvgar: done

src/SolidStateDetector/Semiconductor.jl
claudiaalvgar: done

src/SolidStateDetector/Semiconductor.jl
claudiaalvgar: ```charge_trapping_model``` is already defined, I made it to be ```ctm_dict```

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: This needs to be separate since it's used in ```ConstantLifetimeChargeTrappingModel``` and in ```CombinedChargeTrappingModel```

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I deleted this

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: Yes, I moved this to ```src/ScalarPotentials/PointTypes.jl```

docs/src/man/charge_drift.md
fhagemann: Am I misunderstanding something, or shouldn't the default be "no inactive layer" = "behaving as the bulk"?

claudiaalvgar: Yes, if there's no inactive layer it doesn't do anything but if there's inactive layer, you can have the inactive layer to have the same charge trapping as the bulk or not to have charge trapping at all. It's like a way of disable the inactive layer charge trapping model when it's defined... I think it doesn't do any harm to have it and it could be useful

docs/src/man/charge_drift.md
fhagemann: If you do this
```julia
charge_trapping_model = if haskey(ctm_dict, "model_inactive")
    CombinedChargeTrappingModel{T}(ctm_dict, temperature = temperature)
elseif haskey(ctm_dict, "model") && !haskey(ctm_dict, "model_inactive") && ctm_dict["model"] == "Boggs"
    BoggsChargeTrappingModel{T}(ctm_dict, temperature = temperature)
elseif haskey(ctm_dict, "model") && !haskey(ctm_dict, "model_inactive") && ctm_dict["model"] == "ConstantLifetime"
    ConstantLifetimeChargeTrappingModel{T}(ctm_dict)
else
    NoChargeTrappingModel{T}()
end
```
and then add to the constructor of `CombinedChargeTrappingModel` that if there is no `model`, it's just gonna be `NoChargeTrappingModel` for the bulk.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +212,178 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+        nσ = ifelse(charge > 0, ctm.bulk_charge_trapping_model.nσh, ctm.bulk_charge_trapping_model.nσe)
+    end
+
+    if inactive_layer_ctmodel == "ConstantLifetime"
+        τ_inactive = ifelse(charge > 0, ctm.inactive_charge_trapping_model.τh, ctm.inactive_charge_trapping_model.τe)
+        Δt_minimum = minimum(diff(pathtimestamps))
+        if τ_inactive < Δt_minimum
+            throw(ArgumentError("The carrier lifetime should be at least bigger than Δt"))
+        end
+    end
```
fhagemann: Can this code duplication somehow be avoided? 
Is there any simple way we can just call the original `_calculate_signal!` of the respective (bulk) charge drift models?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +212,178 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+
+    
+    @inbounds for i in eachindex(tmp_signal)
+
+        in_inactive_region = in_inactive_layer(path[i], ctm.inactive_layer_geometry, point_types)
```
fhagemann, line 312: Do we check if `bulk_charge_trapping_model` and `inactive_layer_charge_trapping_model` are identical (and if so to avoid creating a `CombinedChargeTrappingModel`, which would avoid the time-consuming geometry checks)?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +212,178 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+        
+        Δt::T = (i > 1) ? (pathtimestamps[i] - pathtimestamps[i-1]) : zero(T)
+        w::T = i > 1 ? get_interpolation(wpot, path[i], S) : zero(T)
+
+        Δldrift::T = (i > 1) ? norm(path[i] .- path[i-1]) : zero(T)
+        Δl::T = Δldrift > 0 ? hypot(Δldrift, vth * (pathtimestamps[i] - pathtimestamps[i-1])) : zero(T)
```
fhagemann, line 318: If we are calculating everything anyhow, can't we just make a method that also takes the charge drift model (not its name as string) as an argument and dispatch on this instead of defining the very specific `_Boggs_signal!` etc.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -212,23 +212,178 @@ ConstantLifetimeChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = C
...
+        throw(ConfigFileError("`CombinedChargeTrappingModel` does not have `parameters_inactive`"))
+    end
+    
+    bulk_charge_trapping_model = if haskey(config_dict, "model") && config_dict["model"] == "Boggs"
+        BoggsChargeTrappingModel{T}(config_dict, temperature = temperature)
+    elseif haskey(config_dict, "model") && config_dict["model"] == "ConstantLifetime"
+        ConstantLifetimeChargeTrappingModel{T}(config_dict) 
+    else
+        NoChargeTrappingModel{T}()
+    end
```
fhagemann: It seems like, if no `model` is passed, that then `NoChargeTrappingModel` is the default. I would opt against having the option to define `model: ` as being `nothing`.

src/ScalarPotentials/PointTypes.jl
```diff
@@ -326,3 +326,32 @@ function mark_inactivelayer_bits!(point_types::Array{PointType, 3})
...
+    in(pt, g)
+end
+
+function in_inactive_layer(
+    pt::CartesianPoint{T},
+    ::Nothing,
+    point_types::PointTypes{T}
+    )where {T}
+    is_in_inactive_layer(point_types[pt])
+end
```
fhagemann: I feel like these functions could all be one-liners.
```suggestion
@inline in_inactive_layer(pt::CartesianPoint{T}, g::AbstractGeometry{T}, ::PointTypes{T}) where {T} = in(pt, g)
@inline in_inactive_layer(pt::CartesianPoint{T}, ::Nothing, point_types::PointTypes{T}) where {T} = is_in_inactive_layer(point_types[pt])
```

src/SolidStateDetector/Semiconductor.jl
fhagemann: This was not removed (?)

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: But even if they're equal, the CombinedChargeTrappingModel has to be called since the lifetime in the inactive layer is different to the one in the bulk, so the geometry check is needed

src/ScalarPotentials/PointTypes.jl
claudiaalvgar: done

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: Yes, I removed this duplicated code by adding ```check_lifetime``` function

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: With the new push, model: should now not be allowed for the bulk

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I don't think I understand this comment, could you explain it a bit more? or I'm also happy to discuss it in a meeting

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -355,15 +347,22 @@ function CombinedChargeTrappingModel{T}(config_dict::AbstractDict = Dict(); temp
...
-    bulk_charge_trapping_model = if haskey(config_dict, "model") && config_dict["model"] == "Boggs"
-        BoggsChargeTrappingModel{T}(config_dict, temperature = temperature)
-    elseif haskey(config_dict, "model") && config_dict["model"] == "ConstantLifetime"
-        ConstantLifetimeChargeTrappingModel{T}(config_dict) 
-    else
-        NoChargeTrappingModel{T}()
+
+    bulk_charge_trapping_model = begin
+        model = get(config_dict, "model", nothing)
+        if model === nothing
```
fhagemann: Not sure if this makes a difference, but I always use `isnothing(x)` instead of `x === nothing`

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: I like this better, thank you!

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: I see. I was more wondering if they are really identical (not only of identical type), meaning they also have the same lifetimes for example. Because then, we might be able to avoid the costly geometry checks.

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: Now that I'm reading my comment myself, I am also confused :sweat_smile: 

So instead of having new functions with new names like `_Boggs_signal!` or `_constantlifetime_signal!`, just have one method (e.g. `_signal!`) that takes all arguments and dispatches on the charge drift model.

Writing some code (no guarantee if it works):
```julia
# Some function _signal! that takes it all
function _signal!(
    ::BoggsChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T,
    running_sum::Base.RefValue{T},
    nσ::T,
    Δl::T,
    τ::T,
    Δt::T,
) where {T}
    # whatever we need for Boggs
end

function _signal!(
    ::ConstantLifetimeChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T,
    running_sum::Base.RefValue{T},
    nσ::T,
    Δl::T,
    τ::T,
    Δt::T,
) where {T}
    # whatever we need for ConstantLifetime
end

function _signal!(
    ::ConstantLifetimeChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T,
    running_sum::Base.RefValue{T},
    nσ::T,
    Δl::T,
    τ::T,
    Δt::T,
) where {T}
    # whatever we need for NoChargeTrapping
end
```
or maybe making everyhing after `w::T` keyword arguments, so that you only define the ones you need and discard the rest using `kwargs...`

Then you might be able to avoid the `if`-`else` statements in the functions that calculate the waveforms.

docs/src/man/charge_drift.md
fhagemann: Alright, then let's leave this. But I would maybe not advertise this in the docs (especially because it's not physical..)

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I added these extra lines ```charge_trapping_model = if haskey(ctm_dict, "model_inactive")
        if haskey(ctm_dict, "parameters_inactive") && haskey(ctm_dict, "parameters") &&
            ctm_dict["parameters"]==ctm_dict["parameters_inactive"] && ctm_dict["model"] == "ConstantLifetime"
            ConstantLifetimeChargeTrappingModel{T}(ctm_dict)``` in ```Semiconductor.jl``` to check for this

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: done

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I changed this

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -314,22 +324,22 @@ function _calculate_signal(
...
                 # Boggs trapping model in bulk
-                tmp_signal[i] = _Boggs_signal!( running_sum, nσ, Δl, q, w )
+                tmp_signal[i] = _signal!( ctm.bulk_charge_trapping_model, q, w; running_sum=running_sum, nσ=nσ, Δl=Δl )
             else
                 # No charge trapping in bulk
-                tmp_signal[i] = _nochargetrapping_signal!(w, charge)
+                tmp_signal[i] = _signal!( ctm.bulk_charge_trapping_model, q, w )
             end
         end
     end
```
fhagemann: Do you need these `if`-`else` statements at all now?
My idea behind `_signal!` was to be able to replace this whole thing with 
```julia
if in_inactive_region
    tmp_signal[i] = _signal!(ctm.inactive_charge_trapping_model, q, w; running_sum, τ = τ_inactive, Δt)
else
    tmp_signal[i] = _signal!(ctm.bulk_charge_trapping_model, q, w; running_sum, τ, Δt, nσ, Δl)
end
```
(No need to check for `bulk_ctmodel` or `inactive_layer_ctmodel` anymore, because that is passed to `_signal!` now.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -375,6 +387,7 @@ function CombinedChargeTrappingModel{T}(config_dict::AbstractDict = Dict(); temp
     inactive_charge_trapping_model = if haskey(tmp, "model_inactive") && tmp["model_inactive"] == "ConstantLifetime"     
         ConstantLifetimeChargeTrappingModel{T}(tmp)
     else
+        if !isnothing(config_dict["model_inactive"]) @warn "Unknown inactive charge trapping model, running `NoChargeTrappingModel` inside inactive" end
```
fhagemann, line 372: I would argue that we might wanna use the same charge trapping model as the bulk here as fallback default (?) instead of `NoChargeTrappingModel`.

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -171,12 +183,15 @@ function check_lifetime(model::AbstractChargeTrappingModel{T}, charge::T, pathti
...
+function _signal!(
+    ::ConstantLifetimeChargeTrappingModel{T},
     q::Base.RefValue{T},
-    w::T,
+    w::T;
+    running_sum::Base.RefValue{T}=Ref(zero(T)),
+    nσ::T=zero(T),
+    Δl::T=zero(T),
+    τ::T=zero(T),
+    Δt::T=zero(T)
```
fhagemann: Here, every keyword argument that is not used can just become `kwargs...` (also in case we might wanna add more charge trapping models in the future).
And maybe do not use `τ = 0` as default value, because that would kill the charge immediately and return `NaN` ^^

```julia
function _signal!(
    ::ConstantLifetimeChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T;
    running_sum::Base.RefValue{T}=Ref(zero(T)),
    τ::T=T(Inf),
    Δt::T=zero(T),
    kwargs...
) where {T <: SSDFloat}
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -71,12 +80,15 @@ struct BoggsChargeTrappingModel{T <: SSDFloat} <: AbstractChargeTrappingModel{T}
...
+function _signal!(
+    ::BoggsChargeTrappingModel{T},
     q::Base.RefValue{T},
-    w::T,
+    w::T;
+    running_sum::Base.RefValue{T}=Ref(zero(T)),
+    nσ::T=zero(T),
+    Δl::T=zero(T),
+    τ::T=zero(T),
+    Δt::T=zero(T)
```
fhagemann: Same thing here: all unused keywords arguments can just become `kwargs...`:
```julia
function _signal!(
    ::BoggsChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T;
    running_sum::Base.RefValue{T}=Ref(zero(T)),
    nσ::T=zero(T),
    Δl::T=zero(T),
    kwargs...
) where {T <: SSDFloat}
```

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -21,16 +21,22 @@ This model is the default when no charge trapping model is defined in the config
...
+function _signal!(
+    ::NoChargeTrappingModel{T},
+    q::Base.RefValue{T},
+    w::T;
+    running_sum::Base.RefValue{T}=Ref(zero(T)),
+    nσ::T=zero(T),
+    Δl::T=zero(T),
+    τ::T=zero(T),
+    Δt::T=zero(T)
 ) where {T <: SSDFloat}
```
fhagemann: And here:
```julia
function _signal!(
    ::NoChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T;
    kwargs...
) where {T <: SSDFloat}
```

fhagemann (commented):
One final comment and then this should be good to go.

src/ChargeTrapping/ChargeTrapping.jl
claudiaalvgar: I coded it like this so that if model_inactive is set to a not known model but it is defined for example: model_inactive: Random, then it'll throw a warn and run NoChargeTrappingModel and if it's defined in the yaml as (model_inactive:) then it'll also run NoChargeTrappingModel (this would be the case that we discussed previously of allowing disabling the inactive layer). If one wants to run the bulk charge trapping model also inside the inactive layer, then the combined model shouldn't be called i.e. model_inactive shouldn't be written at all in the yaml.
So:
```model_inactive: Random``` written in yaml : throw a warning, calls CombinedChargeTrappingModel and runs NoChargeTrappingModel{T}()
```model_inactive: ``` written in yaml allows to disable the inactive layer calls : CombinedChargeTrappingModel and runs NoChargeTrappingModel{T}()
```model_inactive:``` not written in yaml : runs the bulk charge trapping model inside the bulk and inside the inactive layer

File: src/ChargeTrapping/ChargeTrapping.jl (hunk 27 above)
Comment (claudiaalvgar) [reply src/ChargeTrapping/ChargeTrapping.jl]: I added this

File: src/ChargeTrapping/ChargeTrapping.jl (hunk 29 above)
Comment (claudiaalvgar) [reply src/ChargeTrapping/ChargeTrapping.jl]: I changed these functions to get kwargs...

File: src/ChargeTrapping/ChargeTrapping.jl (hunk 23 above)
Comment (claudiaalvgar) [reply src/ChargeTrapping/ChargeTrapping.jl:318]: I included running_sum in the kwargs because of this comment: "or maybe making everyhing after w::T keyword arguments" and because it's not used in:
```
function _signal!(
    ::NoChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T;
    kwargs...
) where {T <: SSDFloat}
```
but I can add it as normal argument if that's better ?

Comment (fhagemann) [reply src/ChargeTrapping/ChargeTrapping.jl:318]: I would say let’s do that to avoid having to give it a default value.

And here it doesn’t need to even have a name, it can just be 
```julia
function _signal!(
    ::NoChargeTrappingModel{T},
    q::Base.RefValue{T},
    w::T,
    ::Base.RefValue{T};
    kwargs...
) where {T <: SSDFloat}
```

PR #489, opened 2025-05-16
Change AbstractCoordinatePoint to not be a FieldVector

oschulz:
Corresponds to mathematically cleaner "affine vs. linear" language.

PR #540, opened 2025-10-06
Add support for simulation for dead layer

fhagemann:
I merged the current `main` into `transition` and solved the merge conflicts, so this should be the final PR to merge all changes into `main`.

We should see if the current documentation on `main` is still up-to-date or needs some revision before merging these changes. Then, the question is whether this should become 0.10.24 or 0.11.0 (we also made 0.10.0 after adding the Geant4 extension, so maybe the support for dead-layer simulations should also be resembled by a bump in major version).

CC @claudiaalvgar @zpdg


fhagemann:
I'm interested to see if the 32-bit tests fail..

zpdg:
> I merged the current `main` into `transition` and solved the merge conflicts, so this should be the final PR to merge all changes into `main`.
> 
> We should see if the current documentation on `main` is still up-to-date or needs some revision before merging these changes. Then, the question is whether this should become 0.10.24 or 0.11.0 (we also made 0.10.0 after adding the Geant4 extension, so maybe the support for dead-layer simulations should also be resembled by a bump in major version).
> 
> CC @claudiaalvgar @zpdg

Thanks! I think 0.11.0 is better.

oschulz:
> Thanks! I think 0.11.0 is better.

Yes - we should also merge #489 in first, since that's also technically breaking (though most users won't notice).

fhagemann:
> > Thanks! I think 0.11.0 is better.
> 
> Yes - we should also merge #489 in first, since that's also technically breaking (though most users won't notice).

"first" as in "before making a new release" or "before merging this PR".

And do you envision (`->` = merging):
`transition -> main`, `main -> point-types`, `point-types -> main`
or
`point-types -> transition`, `transition -> main`
or
`point-types -> main`, `main -> transition`, `transition -> main`
?

PR #542, opened 2025-10-13
Update type constraint for ParExpBouleImpurityDensity

hervasa2:
The type constraint was incorrect in the Unitfull constructor of ParExpBouleImpurityDensity. This PR fixes this. Note that now its identical to all other BouleImpurityDensity Unitful constructors.

### DIFF
--- a/src/ImpurityDensities/BouleImpurityDensities/ParExpBouleImpurityDensity.jl
+++ b/src/ImpurityDensities/BouleImpurityDensities/ParExpBouleImpurityDensity.jl
@@ -24,7 +24,7 @@ struct ParExpBouleImpurityDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
     det_z0::T
 end
 
-function ParExpBouleImpurityDensity{T}(pars::Vector{Number}, det_z0::Number) where {T}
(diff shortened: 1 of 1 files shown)


fhagemann:
> Will open a new PR with an enhancement for `BouleImpurityDensity` which will include this fix

@hervasa2 will the enhancement be breaking? IF so, we might want to add this first and release it as 0.10.26,
and add the new code in the breaking release 0.11.0

hervasa2:
> Is there any way that we do not have require defining the corrections (*) and (+) for EVERY impurity density that will ever be defined in the future? That was a bit tricky for the Lithium density defined in https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/493

I would argue that we don't need to define it for all. But then we would need to remove these lines and add them to each individual case where we do define them
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/020a26656ae621aa4f05e33b0f998301dea7c4a7/src/ImpurityDensities/ImpurityDensities.jl#L30-L31

Please now see #545 where the type constraint in the Unitfull constructor of ParExpBouleImpurityDensity was also fixed.

PR #545, opened 2025-10-21
Add `SplineBouleImpurityDensity`

hervasa2:
An additional `BouleImpurityDensity` model was added. `SplineBouleImpurityDensity` allows the user to input impurity measurements along the boule which are then interpolated by a cubic spline. This functionality is similar to what exists in `fieldgen` where the user inputs an impurity profile via points in a file. 

Additional tests were added, where a detector is given a `SplineBouleImpurityDensity` with points derived from a `LinExpBouleImpurityDensity`. The spline should then (given the spacing of the input points) approximate the `LinExpBouleImpurityDensity` well. This is checked with a depletion voltage comparison.

### DIFF
--- a/src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
+++ b/src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
@@ -35,7 +35,6 @@ function SplineBouleImpurityDensity{T}(z::AbstractVector{<:RealQuantity}, ρ::Ab
         n = isapprox(ceil(steps) - steps, 1) ? Int(ceil(steps)) : Int(ceil(steps)) + 1
         z_r = range(z_min, stop = z_max, length = n)
         itp = Interpolations.LinearInterpolation(z, ρ, extrapolation_bc=Flat())
-        @info steps, n, z_r
         z_r, itp.(z_r)
     end
 


src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -0,0 +1,36 @@
...
+    ρ::SVector{N, T}
+    spline::Interpolations.Extrapolation{T}
+    det_z0::T
+end
+
+function SplineBouleImpurityDensity{T}(z::Vector{<:Number}, ρ::Vector{<:Number}, det_z0::Number) where {T}
+    N = length(z)
+    @assert N == length(ρ) > 1 "Vectors must be the same length, and at least 2 points are required for cubic spline interpolation."
+    z = to_internal_units.(z)
+    ρ = SVector{N, T}(to_internal_units.(ρ))
```
fhagemann: Maybe a question for @oschulz: do we expect any performance penalties for reusing variables that change their type within a function (`z` is initially a `Vector` passed to the function and then an `SVector` later in the function; same for `ρ`)

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -0,0 +1,36 @@
...
+    spline = cubic_spline_interpolation(range(z[1], stop = z[end], length = N), ρ)
```
fhagemann: Maybe explicitly add that this method is from `Interpolations`, so that people do not look for it in this package?
```suggestion
    spline = Interpolations.cubic_spline_interpolation(range(z[1], stop = z[end], length = N), ρ)
```

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -0,0 +1,36 @@
...
+    SplineBouleImpurityDensity{N, T}(SVector{N, T}(z), ρ, spline, T(to_internal_units(det_z0)))
+end
+
+function get_impurity_density(idm::SplineBouleImpurityDensity, pt::AbstractCoordinatePoint{T})::T where {T}
+    cpt = CartesianPoint(pt)
+    z = cpt[3]
```
fhagemann: The conversion to `CartesianPoint` might not be needed as the third coordinate should always be `z`.

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -0,0 +1,36 @@
...
+* `ρ::SVector{N, T}`: impurity density values at the given z coordinates.
+* `spline::Interpolations.Extrapolation{T}`: cubic spline interpolation of the impurity density values.
+* `det_z0::T`: z coordinate of the detector origin in boule coordinates. The z-direction of the detector is opposite to the z-direction of the boule coordinates.
+"""
+struct SplineBouleImpurityDensity{N, T <: SSDFloat} <: AbstractImpurityDensity{T}
+    z::SVector{N, T}
+    ρ::SVector{N, T}
+    spline::Interpolations.Extrapolation{T}
+    det_z0::T
+end
```
fhagemann, line 17: Do we actually need to store `z` and `ρ` in this struct?
I feel like just `spline` and `det_z0` should be enough.

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -0,0 +1,36 @@
...
+
+function get_impurity_density(idm::SplineBouleImpurityDensity, pt::AbstractCoordinatePoint{T})::T where {T}
+    cpt = CartesianPoint(pt)
+    z = cpt[3]
+    idm.spline(idm.det_z0 - z)
+end
+
+(*)(scale::Real, idm::SplineBouleImpurityDensity{N, T}) where {N, T} = SplineBouleImpurityDensity{T}(Vector(idm.z), Vector(scale*idm.ρ), idm.det_z0)
+
+(+)(offset::Union{<:Real, <:Quantity{<:Real, Unitful.𝐋^(-3)}}, idm::SplineBouleImpurityDensity{N, T}) where {N, T} = SplineBouleImpurityDensity{T}(Vector(idm.z), Vector(idm.ρ .+ to_internal_units(offset)), idm.det_z0)
```
fhagemann: Maybe to avoid having to convert `idm.z` to `Vector`, you can define the constructor for `SplineBouleImpurityDensity` to take an `AbstractVector` instead of just `Vector`

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -0,0 +1,36 @@
...
+* `det_z0::T`: z coordinate of the detector origin in boule coordinates. The z-direction of the detector is opposite to the z-direction of the boule coordinates.
+"""
+struct SplineBouleImpurityDensity{N, T <: SSDFloat} <: AbstractImpurityDensity{T}
+    z::SVector{N, T}
+    ρ::SVector{N, T}
+    spline::Interpolations.Extrapolation{T}
+    det_z0::T
+end
+
+function SplineBouleImpurityDensity{T}(z::Vector{<:Number}, ρ::Vector{<:Number}, det_z0::Number) where {T}
```
fhagemann:
```suggestion
function SplineBouleImpurityDensity{T}(z::AbstractVector{<:Number}, ρ::AbstractVector{<:Number}, det_z0::Number) where {T}
```
(and maybe `RealQuantity` instead of `Number` (?))

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
fhagemann: Looking at this: is this even correct?
Shouldn't `ρ` be evaluated at the actual `z` positions, even when they are unevenly spaced. This seems to create an evenly space range

fhagemann:
```
help?> Interpolations.cubic_spline_interpolation
  etp = cubic_spline_interpolation(knots, A; bc=Line(OnGrid()), extrapolation_bc=Throw())

  A shorthand for extrapolate(scale(interpolate(A, BSpline(Cubic(bc))),
  knots), extrapolation_bc).

  Consider using interpolate, scale, or extrapolate explicitly as needed
  rather than using this convenience constructor. Performance will improve
  without scaling or extrapolation.
```

hervasa2: It currently assumes an evenly spaced input. But in hindsight, this can lead to unexpected behaviour for the user. I will add an evenly spaced resampler to construct the spline.

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
hervasa2: I would keep it in, maybe the user wants to access them after creation.

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
hervasa2: We use Number everywhere in BouleImpurityDensity constructors, and want to allow for unitless numbers

fhagemann: `RealQuantity` is an SSD alias for `Unitful.RealOrRealQuantity` 
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/020a26656ae621aa4f05e33b0f998301dea7c4a7/src/SolidStateDetectors.jl#L87
which is a `Union` of `Real` or `Quantity` (https://github.com/JuliaPhysics/Unitful.jl/blob/9cc01eb486eb1fbf16129e04381ce4817ded4f25/src/types.jl#L238, i.e. real numbers with and without units)

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
fhagemann: shouldn't they be saved in the `spline` object though ?

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -16,21 +16,35 @@ struct SplineBouleImpurityDensity{N, T <: SSDFloat} <: AbstractImpurityDensity{T
     det_z0::T
 end
 
-function SplineBouleImpurityDensity{T}(z::Vector{<:Number}, ρ::Vector{<:Number}, det_z0::Number) where {T}
```
fhagemann: Do you want to update going from `Vector{<:Number}` to `AbstractVector{<:RealQuantity}` also for all other `BouleImpurityDensity` constructors?

hervasa2: I tested this. It doesn't work for other because there is a mix of units in the vectors

fhagemann: I see. But we could still allow for `AbstractVector`, in case somebody might pass an `SVector`, a `LH5Array` (or anything else) for whatever reason

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
fhagemann: This comment is obsolete now, as `spline` might be resampled and we might want to keep the original `z` and `ρ`.

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -16,21 +16,35 @@ struct SplineBouleImpurityDensity{N, T <: SSDFloat} <: AbstractImpurityDensity{T
...
+    z_range, resampled_ρ = if all(x -> isapprox(x, dxs[1]; atol=1e-6), dxs)
+        range(first(z), stop = last(z), length = length(z)), ρ
+    else
+        @info "Resampling impurity density to uniform steps in z"
+        min_step = minimum(dxs)
+        z_min, z_max = first(z), last(z)
+        n = Int(cld(z_max - z_min, min_step))
+        z_r = range(z_min, stop = z_max, length = n)
+        itp = Interpolations.LinearInterpolation(z, ρ, extrapolation_bc=Flat())
+        z_r, itp.(z_r)
```
fhagemann, line 38: Just cross-checking: do you want to require your `spline` to pass through the points that you pass. Because with this code it does not:
```julia 
z = [1,5,6,10]
ρ = [1,2,3,4]
scatter(z, ρ)
plot!(range(extrema(z)..., length = 100),z ->  spline(z))
```
image

hervasa2: Hmm, good eye. They of course pass through the resampled points. Lets see if a come up with a quick fix

hervasa2: This case was fixed with the correct calculation of the length of the range. But yes, in general this is unavoidable if the input points don't match up perfectly with the resampled ones, ie. the range is not divisible by the min step

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
```diff
@@ -25,15 +25,17 @@ function SplineBouleImpurityDensity{T}(z::AbstractVector{<:RealQuantity}, ρ::Ab
...
     else
         @info "Resampling impurity density to uniform steps in z"
         min_step = minimum(dxs)
         z_min, z_max = first(z), last(z)
-        n = Int(cld(z_max - z_min, min_step))
+        steps = (z_max - z_min)/min_step
+        n = isapprox(ceil(steps) - steps, 1) ? Int(ceil(steps)) : Int(ceil(steps)) + 1
         z_r = range(z_min, stop = z_max, length = n)
         itp = Interpolations.LinearInterpolation(z, ρ, extrapolation_bc=Flat())
+        @info steps, n, z_r
```
fhagemann: Do you intend to keep the `@info` statement here? If yes, consider adding more description to the output. If not, please remove it or convert it to a `@debug` statement.

src/ImpurityDensities/BouleImpurityDensities/SplineBouleImpurityDensity.jl
fhagemann: Just adding the plot here with the updated code for reference:
image

fhagemann: It doesn't work for all cases though:
```julia
z = [1,5,6.5,10]
ρ = [1,2,3,4]
scatter(z, ρ)
plot!(range(extrema(z)..., length = 100),z ->  spline(z))
```
image

I can't judge if there are cases where this might be catastrophic (?)

fhagemann:
```julia
z = [1,5,6.5,10]
ρ = [1,2,3,2]
scatter(z, ρ)
plot!(range(extrema(z)..., length = 100),z ->  spline(z))
```
image

hervasa2: Yes, as I said this should only happen now when the range is not divisible by the minimum step. Like in your second example. If users have sparse measurements they should not trust a spline anyhow.

fhagemann:
Okay, I will merge this now onto main (that is on the way of becoming 0.11), and also backport it to the 0.10 branch and release this as 0.10.26 in addition.

fhagemann:
@hervasa2 Do you want to perform some more tests/checks before releasing this as 0.10.26?
https://github.com/JuliaPhysics/SolidStateDetectors.jl/tree/release/0.10

fhagemann:
@hervasa2 if we care about the spline not going through the data points, there seem to be some solutions we could look into (for example https://stackoverflow.com/a/68172187)

PR #549, opened 2025-10-27
Add keyword argument to return waveforms in units of energy

### LINKED ISSUE #284: The user should be able to choose the "unit" of waveforms.
While I agree that it is correct that we simulate the actual induced charge on the preamplifiers and it does e.g. influence the number of charges in the charge cloud, I think in many use-case scenarios it is an unnecessary complication to manually multiply the output waveform with the corresponding `E_ionisation`, to obtain a "calibrated" pulse. This could be integrated into SSD.

One solution would be introducing a keyword to the `get_signals!` function, e.g. `calibrated = true`. 
Then SSD would return e.g. 100 keV pulse for a 100 keV energy deposition, while still handling things correctly under the hood.

An alternate name for the keyword could be `unit = :energy` or `:charge`, respectively. Thoughts?

fhagemann:
This might resolve #284, even though I'm open to discuss the naming for this.

EDIT:
I changed this PR. Instead of having `calibrate::Bool`, we have `signal_unit::Unitful.Units` ~`waveform_unit::Unitful.Units`~ now, that allows to set the unit of the signal. Default is `u"e_au"` (= units of elementary charge), which has been the current default.

INITIAL COMMENT:
~I added a keyword argument `calibrate` to all functions generating `RDWaveforms` to allow the user to obtain the simulated waveforms in units of energy (if set to `true`) or in units of charge (if set to `false`).
Default is `false`, so that after merging this PR, the code should behave the same way as before.~

~I might change some of the tests to use this new keyword argument to check that it performs as intended.
But I would like to hear comments on the keyword name and its implementation first!~

### DIFF
--- a/test/test_isochrone.jl
+++ b/test/test_isochrone.jl
@@ -29,6 +29,7 @@ ev = Event(spawn_positions[in_idx]);
 time_step = T(2)u"ns"
 max_nsteps = 10000
 drift_charges!(ev, sim, Δt = time_step, max_nsteps = max_nsteps, verbose = false)
+SolidStateDetectors.get_signal!(ev, sim, 1, signal_unit = u"keV", Δt = time_step)
 
 DT_h = Array{typeof(time_step),2}(fill(typeof(time_step)(NaN),length(x_axis),length(z_axis)))
 DT_e = Array{typeof(time_step),2}(fill(typeof(time_step)(NaN),length(x_axis),length(z_axis)))


fhagemann:
https://juliaphysics.github.io/SolidStateDetectors.jl/previews/PR549/tutorials/complete_simulation_chain_IVC/
Plotting seems to work now:

image

oschulz:
Hm, I get returning waveforms in units of charge, but in keV seems a bit weird. But if users need it - maybe instead of a bool or symbol one could just pass the desired unit as argument?

oschulz:
Maybe `signal_unit` instead of `waveform_unit`?

PR #551, opened 2025-10-27
Fix plotting `CartesianPoint` with correct units

### LINKED ISSUE #550: Plotting `CartesianPoints` does not seem to be correct
After having merged #489, the plotting of `CartesianPoints` seems to be wrong.
Look at this plot in the documentation: 
https://juliaphysics.github.io/SolidStateDetectors.jl/main/tutorials/complete_simulation_chain_IVC/#Simulation-of-charge-drifts

Image

fhagemann:
Quick attempt to fix #550:
`CartesianPoints` can now have units or be unitless.
However, when plotting, it would be nice to always have them have units.
Right now, if a CartesianPoint does not have units, there is no way that the plot recipe knows what units the coordinates should have. According to our docstrings, we expect x, y and z to be in units of meters though, so this should also be reflected in the plots or be updated.
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/610bcce7ccdd765b0c3e7e2a5b87e8a5305eb8d7/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl#L46-L60

### DIFF
--- a/src/ConstructiveSolidGeometry/plotting/PointsAndVectors/Points.jl
+++ b/src/ConstructiveSolidGeometry/plotting/PointsAndVectors/Points.jl
@@ -7,7 +7,9 @@
     end
     @series begin
         seriestype --> :scatter
-        [pt.x], [pt.y], [pt.z]
+        [to_internal_units(pt.x) * internal_length_unit], 
+        [to_internal_units(pt.y) * internal_length_unit], 
+        [to_internal_units(pt.z) * internal_length_unit]
     end
 end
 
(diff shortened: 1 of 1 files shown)


src/ConstructiveSolidGeometry/plotting/PointsAndVectors/Points.jl
```diff
@@ -7,7 +7,9 @@
     end
     @series begin
         seriestype --> :scatter
-        [pt.x], [pt.y], [pt.z]
+        [to_internal_units(pt.x) * internal_length_unit],
```
fhagemann, line 10: Quick explanation: `to_internal_units` will convert the value to `internal_length_unit` (in our case meters) and then strip the unit (or do nothing if the value does not have any units), and then add the unit again afterwards.
Therefore, we can handle both numbers with and without units.

oschulz: Yes, that seems very safe! I like it.

Issue #552, opened 2025-10-28
Operations with CartesianPoints not supported and causing errors


hervasa2:
Multiplication of scalar and  a `CartesianPoint` and sum of `CartesianPoint`'s is used in `move_charges_inside_semiconductor!`. This is causing errors during charge drift simulation. Given that this functionality was removed in #489 it needs to be explicitly added back in. This can be fixed by a simple
```julia
Base.:(*)(pt::AbstractCartesianPoint{<:Real}, u::Real) = CartesianPoint(get_x(pt) * u, get_y(pt) * u, get_z(pt) * u)
Base.:(*)(u::Real, pt::AbstractCartesianPoint{<:Real}) = pt * u

Base.:(+)(pt1::AbstractCartesianPoint{<:Real}, pt2::AbstractCartesianPoint{<:Real}) = CartesianPoint(get_x(pt1) + get_x(pt2), get_y(pt1) + get_y(pt2), get_z(pt1) + get_z(pt2))
Base.:(-)(pt1::AbstractCartesianPoint{<:Real}, pt2::AbstractCartesianPoint{<:Real}) = CartesianPoint(get_x(pt1) - get_x(pt2), get_y(pt1) - get_y(pt2), get_z(pt1) - get_z(pt2))
```
However, I see this comment

https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/610bcce7ccdd765b0c3e7e2a5b87e8a5305eb8d7/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl#L23

So I did not open a PR at this stage. Instead we need a systematic review of all previously supported methods involving Points as #543 states. I am creating this issue separate to release 0.11 (#543) because of its urgency.

fhagemann:
Will someone look into this in more detail? I will certainly not find the time for this in the near future.

oschulz:
@claudiaalvgar can you have a look sometime is there's other places were linear algebra is attempted with points?

fhagemann:
Or maybe also @benedikt-nagler can have a look (and also write/change the tests to increase the code coverage) ?
I think the issue right now is that there is some substantial code not being covered by the tests and things like this might slip through.

fhagemann:
<img width="1103" height="442" alt="Image" src="https://github.com/user-attachments/assets/bf7da8c9-58f2-4c22-b132-7d5a9d9f4bac" />

[...]

PR #553, opened 2025-10-28
Use vectors instead of points to `move_charges_inside_semiconductor`

hervasa2:
Unsupported operations (since https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/489) with `CartesianPoint`'s were used in `move_charges_inside_semiconductor` causing errors when drifting charges (see #552)

I switched them to vectors by using `point` -> `point - cartesian_zero`. 
This PR fixes these issues and correctly passes `verbose` when using `move_charges_inside_semiconductor`
```julia
simulate_waveforms( mcevents_table::AbstractVector{<:NamedTuple}, sim::Simulation{T}
```


fhagemann:
Can we add a test for this, to make sure that things like this are also tested in the future and covered by the tests?
(Just put a very large charge cloud very close to a surface in any of the tests, maybe the diffusion tests?)

fhagemann:
Downgrade tests failed because the current method of computing a weighted mean of `CartesianPoints` requires a change introduced in `StatsBase@0.33.14` --> https://github.com/JuliaStats/StatsBase.jl/commit/9916e14dac05c507f86d717f9553255776df9593

PR #555, opened 2025-10-29
Fix parsing of linear and cylindrical charge densities

fhagemann:
When parsing **charge** (not impurity!) densities from a config file, the current implementation throws an error.
This is because `internal_charge_density_gradient_unit` was not defined (so SSD does not know how to convert it to internal units), and also `internal_charge_unit` was misspelled in `CylindricalChargeDensity`.

MWE:
```julia
using Unitful, SolidStateDetectors
d = Dict("charge_density" => Dict(
    "name" => "linear",
        "x" => Dict(
            "init" => 1e-10,
            "gradient" => 1.0e-11
       )
    )
)
SolidStateDetectors.ChargeDensity(Float32, d["charge_density"], SolidStateDetectors.default_unit_tuple())
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -50,8 +50,8 @@ end
     include("test_refine_existing_potential.jl")
 end
 
-@timed_testset "Test boule impurity densities" begin
-    include("test_boule_impurity_densities.jl")
+@timed_testset "Test impurity densities" begin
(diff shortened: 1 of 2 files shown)


fhagemann:
I backported this also to `releases/0.10` to be released in the future as 0.10.x

PR #558, opened 2025-11-04
Fix plotting methods with new `CartesianPoint` implementation

fhagemann:
Addressing #552: since #489, the plots for Box did not work properly (`vertices` were swapped and there is one line for sample surface, adding a `Point` to a `Vector` instead of a `Vector` to a `Point`).

This should work now.

```
import SolidStateDetectors.ConstructiveSolidGeometry as CSG
path_to_example_primitives_config_files = joinpath(dirname(dirname(pathof(SolidStateDetectors))), "examples", "example_primitive_files")
cfn = joinpath(path_to_example_primitives_config_files, "Box.yaml")
box = CSG.Geometry(T, cfn)
plot(box)
```
image

HOWEVER: all samplesurface plots appear empty in their default implementation, so something with `auto_strokewidth` or `auto_markersize` does not seems to work well. This still needs to be revisited.i

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -28,6 +28,10 @@ jobs:
       - name: Add LegendJuliaRegistry
         run: julia -e 'using Pkg; Pkg.Registry.add("General"); Pkg.Registry.add(Pkg.RegistrySpec(url = "https://github.com/legend-exp/LegendJuliaRegistry"))'
         shell: bash
+      - name: Downgrade NearestNeighbors # Temporary fix: this is needed to work with older versions of StaticArrays
+        run: julia --project=. -e 'using Pkg; Pkg.add(name = "NearestNeighbors", version = "0.4.13"); Pkg.compat("NearestNeighbors", "0.4.13")'
+        shell: bash
+        if: matrix.version == '1.10'
       - uses: julia-actions/julia-downgrade-compat@v2
         with:
           julia_version: ${{ matrix.version }}
--- a/test/test_geant4.jl
+++ b/test/test_geant4.jl
@@ -69,7 +69,7 @@ end
     @test length(evts) == 100
 
     # Cluster events by radius
-    clustered_evts = @test_nowarn SolidStateDetectors.cluster_detector_hits(evts, 10u"µm")
+    clustered_evts = SolidStateDetectors.cluster_detector_hits(evts, 10u"µm")
     @test length(clustered_evts) == length(evts)
     @test length(flatview(clustered_evts.pos)) <= length(flatview(evts.pos))
     @test eltype(first(clustered_evts.pos)) <: CartesianPoint
@@ -83,7 +83,7 @@ end
 
     # Try the same method using StaticVectors as eltype of pos
     evts_static = Table(evts; pos = VectorOfVectors(broadcast.(p -> SVector{3}(p.x, p.y, p.z), evts.pos)))
-    clustered_evts_static = @test_nowarn SolidStateDetectors.cluster_detector_hits(evts_static, 10u"µm")
+    clustered_evts_static = SolidStateDetectors.cluster_detector_hits(evts_static, 10u"µm")
     @test length(clustered_evts_static) == length(evts_static)
     @test length(flatview(clustered_evts_static.pos)) <= length(flatview(evts_static.pos))
     @test eltype(first(clustered_evts_static.pos)) <: StaticVector{3}


src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl
```diff
@@ -55,7 +55,8 @@ function sample(p::Polygon{N,T}, spacing::T)::Vector{CartesianPoint{T}} where {N
 end
 
 function extremum(p::Polygon{N,T})::T where {N,T}
-    c = sum(p.points)/N
+    c = cartesian_zero + mean(reinterpret(CartesianVector{T}, p.points))  
+    # c = mean(p.points)
```
fhagemann: I can't use `c = mean(p.points)` here, because `p.points` is a `StaticVector{CartesianPoint{T}}` and `mean` here then defaults to the [`mean` method defined in `StaticArraysStatisticsExt` ](https://github.com/JuliaArrays/StaticArrays.jl/blob/a9333fa830d27b85d6c8aa284d32df1b6de0634f/ext/StaticArraysStatisticsExt.jl#L15).

@oschulz any way we can "force" SSD/CSG to use the method defined for `CartesianPoint{T}` that avoids using `+`?

To reproduce the error:
```julia
using SolidStateDetectors, StaticArrays, Statistics; T = Float32
points = @SVector ([CartesianPoint{T}(0,0,0), CartesianPoint{T}(1,1,1)])
mean(points) # throws an error
```

src/ConstructiveSolidGeometry/plotting/CSG/CSG.jl
```diff
@@ -8,6 +8,15 @@ end
...
     if st == :samplesurface
+        label --> l
+        seriesalpha --> 0.2
+        seriescolor --> 1
+        markerstrokewidth --> 0
+        markersize --> 3
+        if occursin("GRBackend", string(typeof(plotattributes[:plot_object].backend)))
+            aspect_ratio --> 1.0
+        end 
+        seriestype := :scatter
```
fhagemann, line 19: By adding default values for `markersize` and `markerstrokewidth` for the `samplesurface` plotting method, we seem to get decent plots:

image
image
image
image
image
image

src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl
oschulz: We should specialize mean for vectors of points. Taking a (weighted or unweighted) barycenter of a point cloud is a well-defined operation in affine geometry. If specializing mean causes ambiguities, we can also define a `barycenter` function ourselves and use that.

fhagemann: There is a `mean` method in CSG, but I think it is causing ambiguities:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a4357bf3ca43ab79d5d7b37518f6eefd2cd5ffab/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl#L127-L138

fhagemann: Or do you mean using `AbstractVector` here instead of `AbstractArray`?

oschulz: It should actually work for any collection, doesn't have to be a vector. mean/barycenter doesn't have any connection to indexing scheme, after all. Even sets are fine. Maybe it's easiest if we really use a separate operation:

```
function barycenter(points)
    vectors = points .- Ref(cartesian_zero)
    mean_vector = mean(vectors)
    R = eltype(points)
    return convert(R, mean_vector + cartesian_zero)::R
end
```

oschulz: Or, better:


```julia
barycenter(X) = _barycenter_impl(eltype(xs), X)

function _barycenter_impl(::Type{T}, points) where {T<:AbstractCoordinatePoint}
    vectors = points .- Ref(cartesian_zero)
    mean_vector = mean(vectors)
    return convert(T, mean_vector + cartesian_zero)::T
end
```

This way, barycenter could be extended to volumes, etc., in the future.

oschulz: Let's reduce memory allocation:

```julia
barycenter(X) = _barycenter_impl(eltype(xs), X)

function _barycenter_impl(::Type{T}, points) where {T<:AbstractCoordinatePoint}
    mean_vector = mean(Base.Fix2(-, cartesian_zero), points)
    return convert(T, mean_vector + cartesian_zero)::T
end
```

fhagemann: In case, we still want to allow for `mean` (even if it's calling some other method `barycenter` afterwards), what's the best way to resolve the ambiguity?

Define a separate method for
```julia
Statistics.mean(v::StaticVector{<:Any,<:P}; kwargs...) where {P <: AbstractCoordinatePoint}
```
?

oschulz: The problem is that many packages specialize `mean` by container type, so we really can't specialize it by element-type for a large range of container types. So it's better to use a more specific method name, I think, esp. since it's the one used on maths for this kind of thing.

fhagemann: I added the new method `barycenter` (had to tweak it a bit), and added tests for both `mean` and `barycenter`,
including that the tests are broken for `mean(::StaticVector)` --> 7512e9b

@oschulz Can you have a look and see if this PR is good to be merged now?

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -125,10 +125,8 @@ end
 @inline Base.adjoint(pt::CartesianPoint) = CartesianPoint(adjoint(pt.x), adjoint(pt.y), adjoint(pt.z))
 
 # Barycentric combination
-function  Statistics.mean(A::AbstractArray{<:CartesianPoint{T}}; dims = :) where T
-    # Equivalent to A .- Ref(cartesian_zero), but allocation-free:
-    B = reinterpret(CartesianVector{T}, A) # ToDo: Avoid this memory allocation.
-    return cartesian_zero + mean(B; dims = dims)
+function Statistics.mean(A::AbstractArray{<:CartesianPoint}; dims = :)
```
oschulz: I would remove using mean for this completely. It'll just cause confusion if it works for some containers and doesn't work for others.

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -298,19 +310,12 @@ Base.iszero(pt::CylindricalPoint) = iszero(pt.r) && iszero(pt.z)
 @inline Base.adjoint(pt::CylindricalPoint) = CylindricalPoint(adjoint(pt.r), adjoint(pt.φ), adjoint(pt.z))
 
 # Barycentric combination
-function  Statistics.mean(A::AbstractArray{<:CylindricalPoint}; dims = :)
-    CylindricalPoint(mean(CartesianPoint, A; dims = dims))
+function Statistics.mean(A::AbstractArray{<:CylindricalPoint}; dims = :)
```
oschulz: I would remove this.

oschulz (changes requested):

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
fhagemann: I removed support for `Statistics.mean` in 050aecc

oschulz (approved):

Issue #523, opened 2025-08-13
Add option to cluster detector hits by time


fhagemann:
The current clustering algorithm in https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/main/src/ChargeClustering/ChargeClustering.jl only clusters by position, but does not consider timing information.

Similar to a `cluster_radius`, it might also make sense to add a `cluster_time`.

Also, when clustering a Geant4 output table, the times `thit` are not clustered but remain untouched.
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e68da8bda44a8b7228745e0477827a34493ba2f7/src/ChargeClustering/ChargeClustering.jl#L59-L74
It might make sense to add (optional) clustering of `thit` also here.

fhagemann:
Thanks for reaching out, 
maybe I can try to explain this issue in a bit more detail using an example:
```julia
using SolidStateDetectors, Geant4, Unitful
sim = Simulation{Float32}(SSD_examples[:InvertedCoax])
source = IsotopeSource(90, 228, 0.0, 0.0, CartesianPoint(0.05,0,0.05))
app = G4JLApplication(sim, source)
t = run_geant4_simulation(app, 1_000)
```
Here, we get a lot of entries in `pos` (position), `edep` (deposited energy) and `thit` (time of hit).

```julia
t.edep[1]
```
```
7-element view(::Vector{Quantity{Float32, 𝐋^2 𝐌 𝐓^-2, Unitful.FreeUnits{(keV,), 𝐋^2 𝐌 𝐓^-2, nothing}}}, 1:7) with eltype Quantity{Float32, 𝐋^2 𝐌 𝐓^-2, Unitful.FreeUnits{(keV,), 𝐋^2 𝐌 𝐓^-2, nothing}}:
  132.01038f0 keV
 5685.4146f0 keV
  103.50545f0 keV
 6288.1274f0 keV
  116.61237f0 keV
 6778.213f0 keV
    5.398038f0 keV
```
```julia
t.thit[1]
```
```
7-element view(::Vector{Quantity{Float32, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}}, 1:7) with eltype Quantity{Float32, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}:
 1.7250684f6 s
 1.7991591f6 s
 1.7991591f6 s
 1.7992252f6 s
 1.7992252f6 s
[...]

PR #559, opened 2025-11-06
Add option to split MC-table events into separate events by time

sm2909:
for issue #523 

Clustering was supposed to be based on position and time simultaneously.
To achieve this, I implemented a Spatio-Temporal DBSCAN (ST-DBSCAN) algorithm, which includes a point in a cluster only when both spatial and temporal constraints are satisfied.

Traditional DBSCAN uses a Euclidean metric, which mixes position and time dimensions. This could incorrectly group points that are spatially close but temporally far apart (for example, position inside but time outside the threshold, and still euclidean distance inside threshold). The new approach avoids that issue by handling both constraints explicitly.

That is why I have used this new algorithm. Any prior implementation of this did not exist in julia libraries, so I have coded it from scratch

A new dependency Neighbors.jl was added in project.toml for this algorithm

`
tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm", 100u"s")
`

When this is run, clustering will happen with time constraint as well. Cluster properties (pos, edep) are calculated as before with the addition of thit as well.

```
tc.thit[1]
3-element view(::Vector{Quantity{Float32, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}}, 1:3) with eltype Quantity{Float32, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}:
 1.7250684f6 s
 1.7992052f6 s
 1.7991948f6 s
```

(This was run in the same case with which you demonstrated me in previous comment)

The problem that it has is that it is taking a lot of computer resources to compute the clusters. I tried using the best approaches (KDTree, inrange from Neighbors.jl), but the issue is arising due to the complexity of the problem as we have to simultaneously cluster based on position and time.

This is the computer resources typically being used when clustering using cluster_detector_hits function:

`133.839229 seconds (35.63 M allocations: 1.751 GiB, 0.57% gc time, 99.71% compilation time)`

(There might be a better approach or performance gains to be found. I tried my best and came up with this.)

Open for any modifications, which might be required for it to be integrated with whole project.

### DIFF
--- a/test/test_geant4.jl
+++ b/test/test_geant4.jl
@@ -128,13 +128,22 @@ end
     
     # Generate waveforms
     timed_simulate!(sim, refinement_limits = [0.2,0.1,0.05,0.03,0.02])
+    evt = Event(evts[1], T)
+    @test evt isa Event{T}
+    timed_simulate!(evt, sim, Δt = 1u"ns", max_nsteps = 2000)
 
-    # Generate waveforms with unclustered events
+    # Test generating electron and hole contribution separately
+    (; electron_contribution, hole_contribution) = get_electron_and_hole_contribution(evt, sim, 1)
+    @test electron_contribution isa RDWaveform
+    @test hole_contribution isa RDWaveform
+
+    # Use the raw table
     wf = timed_simulate_waveforms(evts, sim, Δt = 1u"ns", max_nsteps = 2000)
     @test wf isa Table
     @test SolidStateDetectors.get_detector_hits_table(wf[findall(wf.chnid .== 1)]) == evts
     @test :waveform in columnnames(wf)
     @test length(wf) == length(evts) * sum(.!ismissing.(sim.weighting_potentials))
+    @test wf.waveform[1] ≈ evt.waveforms[1]
 
     # Generate waveforms from the clustered events
     wf_clustered = timed_simulate_waveforms(clustered_evts, sim, Δt = 1u"ns", max_nsteps = 2000)
--- a/src/ConstructiveSolidGeometry/GeometryRounding.jl
+++ b/src/ConstructiveSolidGeometry/GeometryRounding.jl
@@ -1,23 +1,21 @@
 csg_default_tol(::Type{T}) where {T} = T(1e-12)
 csg_default_tol(::Type{Float32}) = 1f-8
 
-geom_sigdigits(::Type{Int64})::Int = 12
 geom_sigdigits(::Type{Float32})::Int = 6
 geom_sigdigits(::Type{Float64})::Int = 12
 
-geom_atol_zero(::Type{Int64})::Float64 = 1e-12
 geom_atol_zero(::Type{Float32})::Float32 = 1e-6
 geom_atol_zero(::Type{Float64})::Float64 = 1e-12
 
-function geom_round(v::T)::T where {T}
+function geom_round(v::T)::T where {T <: Union{Float32, Float64}}
     return isapprox(v, 0, atol = geom_atol_zero(T)) ? T(0) : round(v, sigdigits = geom_sigdigits(T))
 end
 
-function geom_round(pt::CylindricalPoint{T})::CylindricalPoint{T} where {T <: Real}
+function geom_round(pt::CylindricalPoint{T})::CylindricalPoint{T} where {T <: Union{Float32, Float64}}
     return CylindricalPoint{T}( geom_round(pt.r), geom_round(pt.φ), geom_round(pt.z)  )
 end
 
-function geom_round(pt::CartesianPoint{T})::CartesianPoint{T} where {T <: Real}
+function geom_round(pt::CartesianPoint{T})::CartesianPoint{T} where {T <: Union{Float32, Float64}}
     return CartesianPoint{T}( geom_round(pt.x), geom_round(pt.y), geom_round(pt.z)  )
 end
 
--- a/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
+++ b/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
@@ -13,9 +13,12 @@ function Base.isapprox(a::AbstractCartesianPoint, b::AbstractCartesianPoint; kwa
            isapprox(get_z(a), get_z(b); kwargs...)
 end
 
+
 # Unitful uses the `*` and `/` operators to combine values with units. But mathematically that's really not an algebraic
 # product (not defined for affine points), but a cartesian product, so supporting this should be fine:
 Base.:(*)(pt::AbstractCartesianPoint{<:Real}, u::Unitful.Units{<:Any,Unitful.𝐋}) = CartesianPoint(get_x(pt) * u, get_y(pt) * u, get_z(pt) * u)
+# ToDo: Uncomment once units in points are supported
+#=
 Base.:(/)(pt::AbstractCartesianPoint{<:Quantity{<:Real, Unitful.𝐋}}, u::Unitful.Units{<:Any,Unitful.𝐋}) = CartesianPoint(get_x(pt) / u, get_y(pt) / u, get_z(pt) / u)
 
 
@@ -31,10 +34,7 @@ function Unitful.uconvert(u::Unitful.Units{<:Any,Unitful.NoDims}, pt::AbstractCa
 end
 
 Unitful.ustrip(pt::AbstractCartesianPoint) = CartesianPoint(ustrip(get_x(pt)), ustrip(get_y(pt)), ustrip(get_z(pt)))
-
-
-
-
+=#
 
 
 """
@@ -111,8 +111,8 @@ Base.getindex(p::CartesianPoint, k::Symbol) = getfield(p, k)
 AbstractCoordinatePoint{T, Cartesian}(x::Real, y::Real, z::Real) where T = CartesianPoint{T}(x, y, z)
 
 
-Base.zero(::CartesianPoint{T}) where {T} = CartesianPoint{T}(zero(T),zero(T),zero(T))
-# @inline Base.Tuple(pt::CartesianPoint) = (pt.x, pt.y, pt.z)
+@inline Base.zero(::CartesianPoint{T}) where {T} = CartesianPoint{T}(zero(T),zero(T),zero(T))
+@inline Base.iszero(pt::CartesianPoint) = iszero(pt.x) && iszero(pt.y) && iszero(pt.z)
 @inline Base.copy(pt::CartesianPoint) = CartesianPoint(pt.x, pt.y, pt.z)
 
 @inline Base.:(+)(pt::CartesianPoint, v::CartesianVector) = CartesianPoint(pt.x + v.x, pt.y + v.y, pt.z + v.z)
@@ -210,17 +210,20 @@ end
     return CartesianVector(zero(R), zero(R), zero(R))
 end
 
-@inline Base.zero(::CartesianZero{T}) where {T} = CartesianZero{T}()
+@inline Base.zero(::Union{<:CartesianZero{T}, <:Type{CartesianZero{T}}}) where {T} = CartesianZero{T}()
 @inline Base.iszero(::CartesianZero) = true
 
-# ToDo: Revert this once we support units internally.
-#Base.:(*)(::CartesianZero{T}, u::Unitful.Units{<:Any,Unitful.𝐋}) where {T<:Real} = CartesianZero{Quantity{T, Unitful.𝐋, typeof(u)}}()
-Base.:*(z::CartesianZero, u::Unitful.Quantity) = z
 
+# ToDo: Revert this once we support units internally.
+Base.:(*)(::CartesianZero{T}, u::Unitful.Units{<:Any,Unitful.𝐋}) where {T<:Real} = CartesianZero{Quantity{T, Unitful.𝐋, typeof(u)}}()
+Base.:(*)(z::CartesianZero, u::Unitful.Quantity) = z
+
+# ToDo: Uncomment once units in points are supported
+#=
 function Unitful.uconvert(u::Unitful.Units{T,Unitful.NoDims}, pt::CartesianZero{<:Quantity{U, Unitful.NoDims}}) where {T,U}
     CartesianZero{promote_type(T,U)}()
 end
-
+=#
 
 
 """
@@ -323,8 +326,8 @@ function Base.isapprox(a::CylindricalPoint, b::CylindricalPoint; kwargs...)
     return isapprox(a.r, b.r; kwargs...) && isapprox(a.φ, b.φ; kwargs...) && isapprox(a.z, b.z; kwargs...)
 end
 
-Base.zero(::CylindricalPoint{T}) where {T} = CylindricalPoint{T}(zero(T), zero(T), zero(T))
-Base.iszero(pt::CylindricalPoint) = iszero(pt.r) && iszero(pt.z)
+@inline Base.zero(::CylindricalPoint{T}) where {T} = CylindricalPoint{T}(zero(T), zero(T), zero(T))
+@inline Base.iszero(pt::CylindricalPoint) = iszero(pt.r) && iszero(pt.z)
 
 @inline Base.copy(pt::CylindricalPoint) = CylindricalPoint(pt.r, pt.φ, pt.z)
 
@@ -351,5 +354,5 @@ function to_internal_units(pt::CylindricalPoint)
 end
 
 function to_internal_units(pt::AbstractCoordinatePoint)
-    error("Unsupported point type $(typeof(pt)). Expected CartesianPoint or CylindricalPoint.")
+    throw(ArgumentError("Unsupported point type $(typeof(pt)). Expected CartesianPoint or CylindricalPoint."))
 end
--- a/src/ConstructiveSolidGeometry/PointsAndVectors/PointsAndVectors.jl
+++ b/src/ConstructiveSolidGeometry/PointsAndVectors/PointsAndVectors.jl
@@ -1,6 +1,6 @@
 abstract type AbstractCoordinatePoint{T, S} end
 
-zero(PT::Type{<:AbstractCoordinatePoint{T}}) where {T} = PT(zero(T),zero(T),zero(T))
+@inline Base.zero(PT::Type{<:AbstractCoordinatePoint{T}}) where {T} = PT(zero(T),zero(T),zero(T))
 
 @inline StaticArrays.Size(::PT) where {PT<:AbstractCoordinatePoint} = Size(PT)
 @inline StaticArrays.Size(::Type{PT}) where {PT<:AbstractCoordinatePoint} = Size{(fieldcount(PT),)}()
--- a/src/ConstructiveSolidGeometry/PointsAndVectors/Vectors.jl
+++ b/src/ConstructiveSolidGeometry/PointsAndVectors/Vectors.jl
@@ -51,7 +51,7 @@ CartesianVector(; x = 0, y = 0, z = 0) = CartesianVector(x,y,z)
 
 CartesianVector{T}(; x = 0, y = 0, z = 0) where {T} = CartesianVector{T}(T(x),T(y),T(z))
 
-@inline Base.zero(::CartesianVector{T}) where {T} = CartesianVector{T}(zero(T),zero(T),zero(T))
+@inline Base.zero(::Union{<:CartesianVector{T}, <:Type{CartesianVector{T}}}) where {T} = CartesianVector{T}(zero(T),zero(T),zero(T))
 @inline Base.iszero(v::CartesianVector) = iszero(v.x) && iszero(v.y) && iszero(v.z)
 
 # Need to specialize multiplication and division with units for CartesianVector, otherwise result would just be an SVector:
--- a/src/Event/Event.jl
+++ b/src/Event/Event.jl
@@ -124,22 +124,20 @@ end
 function Event(evt::NamedTuple{DHE_column_names, <:Tuple{DHE_column_types...}}, T = eltype(to_internal_units(evt.edep[:])))::Event{T}
     return Event(
         [CartesianPoint{T}.(to_internal_units.(evt.pos[:]))],
-        [T.(to_internal_units.(evt.edep[:]))]
+        [T.(to_internal_units.(evt.edep[:]))];
     )
 end
 
 in(evt::Event, det::SolidStateDetector) = all( pt -> pt in det, flatview(evt.locations))
-in(evt::Event, sim::Simulation) = all( pt -> pt in sim.detector, flatview(evt.locations))
+in(evt::Event, sim::Simulation) = in(evt, sim.detector)
 
-function move_charges_inside_semiconductor!(evt::Event{T}, det::SolidStateDetector{T}; fraction::T = T(0.2), verbose::Bool = true)::Event{T} where {T <: SSDFloat}
+function move_charges_inside_semiconductor!(evt::Event{T}, det::SolidStateDetector{T}; fraction::T = T(0.2), verbose::Bool = true) where {T <: SSDFloat}
     move_charges_inside_semiconductor!(evt.locations, evt.energies, det; fraction, verbose)
-    evt
 end
 
 function move_charges_inside_semiconductor!(
         locations::AbstractVector{<:AbstractVector{CartesianPoint{T}}}, energies::AbstractVector{<:AbstractVector{T}},
         det::SolidStateDetector{T}; fraction::T = T(0.2), verbose::Bool = true) where {T <: SSDFloat}
-    global g_state = (;locations, energies, det)
     for n in eachindex(locations)
         idx_in = broadcast( pt -> pt in det.semiconductor, locations[n]);
         if !all(idx_in)
@@ -164,12 +162,12 @@ function move_charges_inside_semiconductor!(
             end
             charge_center_new = barycenter(locations[n], Weights(edep_weights))
             if verbose
-                @warn "$(sum(.!idx_in)) charges of the charge cloud at $(round.(charge_center - cartesian_zero, digits = (T == Float64 ? 12 : 6)))"*
-                " are outside. Moving them inside...\nThe new charge center is at $(round.(charge_center_new - cartesian_zero, digits = (T == Float64 ? 12 : 6))).\n"
+                @warn "$(sum(.!idx_in)) charges of the charge cloud at $(geom_round(charge_center))"*
+                " are outside. Moving them inside...\nThe new charge center is at $(geom_round(charge_center_new)).\n"
             end
         end
     end
-    nothing
+    return nothing
 end
 
 """
@@ -200,13 +198,13 @@ drift_charges!(evt, sim, Δt = 1u"ns", verbose = false)
     Using values with units for `Δt` requires the package [Unitful.jl](https://github.com/JuliaPhysics/Unitful.jl).
 """
 function drift_charges!(evt::Event{T}, sim::Simulation{T}; max_nsteps::Int = 1000, Δt::RealQuantity = 5u"ns", diffusion::Bool = false, self_repulsion::Bool = false, end_drift_when_no_field::Bool = true, geometry_check::Bool = false, verbose::Bool = true)::Nothing where {T <: SSDFloat}
-    !in(evt, sim.detector) && move_charges_inside_semiconductor!(evt, sim.detector; verbose)
+    !in(evt, sim) && move_charges_inside_semiconductor!(evt, sim.detector; verbose)
     evt.drift_paths = drift_charges(sim, evt.locations, evt.energies; Δt, max_nsteps, diffusion, self_repulsion, end_drift_when_no_field, geometry_check, verbose)
     nothing
 end
 function get_signal!(evt::Event{T}, sim::Simulation{T}, contact_id::Int; Δt::RealQuantity = 5u"ns", signal_unit::Unitful.Units = u"e_au")::Nothing where {T <: SSDFloat}
-    @assert !ismissing(evt.drift_paths) "No drift path for this event. Use `drift_charges(evt::Event, sim::Simulation)` first."
-    @assert !ismissing(sim.weighting_potentials[contact_id]) "No weighting potential for contact $(contact_id). Use `calculate_weighting_potential!(sim::Simulation, contact_id::Int)` first."
+    ismissing(evt.drift_paths) && throw(AssertionError("The charge drift is not yet simulated. Please run `drift_charges!(evt::Event, sim::Simulation)` first."))
+    ismissing(sim.weighting_potentials[contact_id]) && throw(AssertionError("No weighting potential for contact $(contact_id). Please run `calculate_weighting_potential!(sim::Simulation, contact_id::Int)` first."))
     if ismissing(evt.waveforms)
         evt.waveforms = Union{Missing, RadiationDetectorSignals.RDWaveform}[missing for i in eachindex(sim.detector.contacts)]
     end
@@ -239,15 +237,11 @@ SolidStateDetectors.get_signals!(evt, sim, Δt = 1u"ns") # if evt.drift_paths we
 !!! note 
     This method only works if `evt.drift_paths` has already been calculated and is not `missing`.
 """
-function get_signals!(evt::Event{T}, sim::Simulation{T}; Δt::RealQuantity = 5u"ns", signal_unit::Unitful.Units = u"e_au")::Nothing where {T <: SSDFloat}
-    @assert !ismissing(evt.drift_paths) "No drift path for this event. Use `drift_charges!(evt::Event, sim::Simulation)` first."
-    if ismissing(evt.waveforms)
-        evt.waveforms = Union{Missing, RadiationDetectorSignals.RDWaveform}[missing for i in eachindex(sim.detector.contacts)]
-    end
+function get_signals!(evt::Event{T}, sim::Simulation{T}; kwargs...)::Nothing where {T <: SSDFloat}
+    any(ismissing, sim.weighting_potentials) && @warn "No weighting potential(s) for some contact(s).."
     for contact in sim.detector.contacts
-        if any(ismissing, sim.weighting_potentials) "No weighting potential(s) for some contact(s).." end
         if !ismissing(sim.weighting_potentials[contact.id])
-            evt.waveforms[contact.id] = get_signal(sim, evt.drift_paths, flatview(evt.energies), contact.id; Δt, signal_unit)
+            get_signal!(evt, sim, contact.id; kwargs...)
         end
     end
     nothing
@@ -318,11 +312,8 @@ function add_baseline_and_extend_tail(wv::RadiationDetectorSignals.RDWaveform{T,
     else
         new_signal[n_baseline_samples+1:end] = wv.signal[1:total_waveform_length - n_baseline_samples]
     end
-    new_times = if TV <: AbstractRange
-        range( zero(first(wv.time)), step = step(wv.time), length = total_waveform_length )
-    else
-        error("Not yet defined for timestamps of type `$(TV)`")
-    end
+    TV <: AbstractRange || throw(ArgumentError("Not yet defined for timestamps of type `$(TV)`"))
+    new_times = range(first(wv.time), step = step(wv.time), length = total_waveform_length )
     return RDWaveform( new_times, new_signal )
 end
 
@@ -362,7 +353,7 @@ See also [`plot_electron_and_hole_contribution`](@ref).
 function get_electron_and_hole_contribution(evt::Event{T}, sim::Simulation{T, S}, contact_id::Int; signal_unit::Unitful.Units = u"e_au",
             )::NamedTuple{(:electron_contribution, :hole_contribution), <:Tuple{RDWaveform, RDWaveform}} where {T <: SSDFloat, S}
     
-    @assert !ismissing(evt.drift_paths) "The charge drift is not yet simulated. Please use `drift_charges!(evt, sim)`!"
+    ismissing(evt.drift_paths) && throw(AssertionError("The charge drift is not yet simulated. Please run `drift_charges!(evt::Event, sim::Simulation)` first."))
     
     dt::T = T(ustrip(u"ns", diff(evt.drift_paths[1].timestamps_e)[1]*u"s"))
     wp::Interpolations.Extrapolation{T, 3} = interpolated_scalarfield(sim.weighting_potentials[contact_id])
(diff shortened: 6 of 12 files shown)


fhagemann:
Thanks for the initiative, I will look into this in more detail once I find a quite moment.
Maybe a quick question to get an order of magnitude: the `133.839229 seconds` you quote is for how many events/rows of the initial table?

sm2909:
1000 rows

These are the exact commands that I ran:

```
using SolidStateDetectors, Geant4, Unitful
sim = Simulation{Float32}(SSD_examples[:InvertedCoax])
source = IsotopeSource(90, 228, 0.0, 0.0, CartesianPoint(0.05,0,0.05))
app = G4JLApplication(sim, source)
t = run_geant4_simulation(app, 1_000)
@time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm", 100u"s")
```

fhagemann:
I see, thanks!
So, what I'm seeing is that these changes do two things:
- First, updating the current method of `cluster_detector_hits` to the new implementation of `CartesianPoint` (#489), which on the current `main` branch does not seem to be working with the Geant4 output.
- Update the clustering algorithm to cluster both time and/or positions

Would it be very cumbersome for you to split these two things into two commits, to see what exactly the changes to the clustering algorithm are?

And whether the long run-time of 2 minutes results from the updates to the new `CartesianPoints` (maybe benchmark this before implementing the timing clustering and see how long it takes to run for 1000 events) or the new clustering?

test/test_geant4.jl
```diff
@@ -66,6 +67,11 @@ end
     @test evts isa Table
     @test length(evts) == 100
 
+    # Cluster events by radius and time
+    @test_nowarn clusted_evts = SolidStateDetectors.cluster_detector_hits(evts, 10u"µm", 100u"s")
```
fhagemann: I'm sorry, my previous message had a typo which is causing these tests to fail :sweat_smile: 
```suggestion
    @test_nowarn clustered_evts = SolidStateDetectors.cluster_detector_hits(evts, 10u"µm", 100u"s")
```

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -44,28 +51,36 @@ function cluster_detector_hits(
                 esum = ustrip(esum_u)
                 if esum ≈ 0
                     push!(r_pos, mean(c_hits.pos))
+                    push!(r_thit, mean(c_hits.thit))
                 else
                     weights = ustrip.(c_hits.edep) .* inv(esum)
                     push!(r_pos, sum(c_hits.pos .* weights))
+                    push!(r_thit, sum(c_hits.thit .* weights))
```
fhagemann: You could ensure to convert `r_pos` back to its original type (`StaticVector` or `CartesianPoint`) when returning this.

Also, seems like we're not adding the `posunit` back on (it would be good if the positions in the clustered `Table` would be of identical type as the original one).

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -8,31 +8,38 @@ function cluster_detector_hits(
...
+    thitunit = unit(TTT)
+    ustripped_ctime = ustrip(thitunit, cluster_time)
     
     for d_hits_nt in grouped
         d_hits = Table(d_hits_nt)
         d_detno = first(d_hits.detno)
         @assert all(isequal(d_detno), d_hits.detno)
         if length(d_hits) > 3
-            clusters = Clustering.dbscan(ustrip.(flatview(d_hits.pos)), ustripped_cradius, leafsize = 20, min_neighbors = 1, min_cluster_size = 1)
+            clusters = st_dbscan(hcat(ustrip.(d_hits.pos)...)', ustrip.(d_hits.thit), eps_space = ustripped_cradius, eps_time = ustripped_ctime, min_cluster_size = 1)
```
fhagemann: Maybe instead of converting everything first to an `SVector` and then to this `Matrix` for `st_dbscan`, we might consider defining a function that defines this matrix directly from `grouped` or `d_hits.pos`

Could be something like (I did not test this, just wrote this out in this comment as a starting point):
```julia
function create_dbscan_matrix(pos::AbstractVector{CartesianPoint{T}})
    ustrip.(hcat(getfield.(pos, :x), getfield.(pos, :y), getfield.(pos, :z)))'
end

function create_dbscan_matrix(pos::AbstractVector{StaticVector{T}})
    ustrip.(hcat(d_hits.pos)...))'
end

File: src/ChargeClustering/ChargeClustering.jl [hunk 4]
```diff
@@ -130,3 +145,78 @@ function group_points_by_distance(pts::AbstractVector{CartesianPoint{T}}, energi
     clustersidx, elem_ptr = _group_points_by_distance(pts, group_distance)
     VectorOfVectors(pts[clustersidx], elem_ptr), VectorOfVectors(energies[clustersidx], elem_ptr), VectorOfVectors(radius[clustersidx], elem_ptr)
 end
+
+function st_dbscan(pos::AbstractMatrix, time::AbstractVector;
```
fhagemann: Does this method still work with `Clustering@0.14` or do we need to discontinue that support in the `Project.toml`?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -130,3 +145,78 @@ function group_points_by_distance(pts::AbstractVector{CartesianPoint{T}}, energi
...
+    cluster_id = 0
+    
+    # --- Convert position input to Float64 matrix if needed ---
+    if eltype(pos) <: CartesianPoint
+        pos = hcat([SVector(p.x, p.y, p.z) for p in pos]...)'
+    elseif eltype(pos) <: StaticVector
+        pos = hcat(pos...)'   # convert Vector of SVector{3,T} to Nx3 Matrix
+    elseif !(pos isa AbstractMatrix{<:AbstractFloat})
+        error("Unsupported position format: must be a matrix of floats or vector of points")
+    end
```
fhagemann: Do we still need this or is this handled is `cluster_detector_hits` already.
Because you require your input to this method to already be an `AbstractMatrix`.

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -130,3 +145,78 @@ function group_points_by_distance(pts::AbstractVector{CartesianPoint{T}}, energi
...
+
+    # KDTree for spatial queries
+    tree = NearestNeighbors.KDTree(pos'; leafsize = leafsize)
```
fhagemann: Your transposing `pos` in `cluster_detector_hits` when passing this to `st_dbscan`, but transpose this here again. Can this be avoided?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -130,3 +145,78 @@ function group_points_by_distance(pts::AbstractVector{CartesianPoint{T}}, energi
...
+                visited[j] = true
+                spatial_j = NearestNeighbors.inrange(tree, pos[j, :], eps_space)
+                neighbors_j = [l for l in spatial_j if abs(time[j] - time[l]) ≤ eps_time]
+                cluster_mask[neighbors_j] .= true
+                if length(neighbors_j) ≥ min_neighbors
+                    core_mask[j] = true
+                    append!(to_explore, neighbors_j)
+                end
+            end
+        end
```
fhagemann: I'm wondering here: would it maybe be more efficient to 
1. split all events by time (separate a row into multiple rows if the hits are more than `eps_time` apart),
2. then perform the spatial clustering like we used to before (or the more efficient implementation)?

Project.toml
```diff
@@ -57,14 +58,15 @@ FillArrays = "0.9, 0.10, 0.11, 0.12, 0.13, 1"
...
-Interpolations = "0.14, 0.15, 0.16"
+Interpolations = "0.14.2, 0.15, 0.16"
 IntervalSets = "0.6, 0.7"
 InverseFunctions = "0.1"
 JSON = "0.21.2, 1.1"
 KernelAbstractions = "0.8, 0.9"
 LaTeXStrings = "1.1"
 LightXML = "0.9"
 LinearAlgebra = "<0.0.1, 1"
+NearestNeighbors = "0.4.22"
```
fhagemann: Could this also allow for older versions of `NearestNeighbors` or is that the oldest version we should support?

Project.toml
```diff
@@ -57,14 +58,15 @@ FillArrays = "0.9, 0.10, 0.11, 0.12, 0.13, 1"
 Format = "1.2"
 GPUArrays = "8, 9, 10, 11"
 Geant4 = "0.1.13, 0.2"
-Interpolations = "0.14, 0.15, 0.16"
+Interpolations = "0.14.2, 0.15, 0.16"
```
fhagemann: Some `[compat]` changes in the `Project.toml` are not up-to-date with the current `main` branch.

fhagemann (commented):
Here are some comments.

I'm wondering here: would it maybe be more efficient to 
1. split all events by time (separate a row into multiple rows if the hits are more than `eps_time` apart). I expect most of the `Geant4` hits to have identical/very similar `thit` entries if they resulted from the same nuclear decay.
2. then, perform the spatial clustering like we used to before (or the more efficient implementation)?

Based on your experience in implementing this, do you think that this would be faster?
Or do you have some feeling for what is limiting the runtime of this clustering?

sm2909:
> Here are some comments.
> 
> I'm wondering here: would it maybe be more efficient to
> 
> 1. split all events by time (separate a row into multiple rows if the hits are more than `eps_time` apart). I expect most of the `Geant4` hits to have identical/very similar `thit` entries if they resulted from the same nuclear decay.
> 2. then, perform the spatial clustering like we used to before (or the more efficient implementation)?
> 
> Based on your experience in implementing this, do you think that this would be faster? Or do you have some feeling for what is limiting the runtime of this clustering?

That would lead to incorrect clustering. Consider this toy example:

eps_time = 10s
eps_pos = 10m
(consider position 1d here for easier visualization)

hit 1: 1s, 1m
hit 2: 10s, 100m
hit 3: 15s, 5m

First clustering by time, all hits are in same cluster (eps_time < 10).
Then cluster based on position, [1, 3] are a cluster and [2] is another cluster.
But [1, 3] should not belong to same cluster as eps_time > 10 between them.

(Earlier hit 2 formed the common link between hit 1 and 3 for eps_time. But as finally 2 is not in the same cluster, hit 1 and 3 should not be kept in the same cluster.)

fhagemann:
> That would lead to incorrect clustering. Consider this toy example:
> 
> eps_time = 10s eps_pos = 10m (consider position 1d here for easier visualization)
> 
> hit 1: 1s, 1m hit 2: 10s, 100m hit 3: 15s, 5m
> 
> First clustering by time, all hits are in same cluster (eps_time < 10). Then cluster based on position, [1, 3] are a cluster and [2] is another cluster. But [1, 3] should not belong to same cluster as eps_time > 10 between them.
> 
> (Earlier hit 2 formed the common link between hit 1 and 3 for eps_time. But as finally 2 is not in the same cluster, hit 1 and 3 should not be kept in the same cluster.)

I see what you mean.
The idea behind time clustering would be to separate events that are more than a given `eps_time` apart. This is because, in our pulse shape simulation, we start drifting all charge clouds at the same initial time.
Other than with distance, I would also like to make sure that grouped events are never more than `eps_time` apart.

So, even if the hits were 
```
hit 1: 1s, 1m 
hit 2: 10s, 1m
hit 3: 15s, 1m
```
and `eps_time` was `10s`, I would like to NEVER group hit 1 and hit3.

Let`s assume this toy model:
```
for N in 1:1000
hit (N): (N) ns, 1m

```
If I choose `eps_time = 1u"ns"` because my simulation step is `1u"ns"`, I would group all hits within the `1u"µs"` together, even though this might not be what I want. I, at most, would like the first two hits to be combined and start a new group with the third hit etc. 

Just to think out loud (doesn't mean that this is the best way to go): the time grouping could either start with the first hit and combine everything that's withing `eps_time` after the first hit.
OR we can sort all events by time and start grouping with the two hits that are closest together, group them, and then keep on adding more events if they are withing `eps_time` or start with new events afterwards.

The time clustering should be there to make sure that we only drift hits together, if they actually coincide "enough" in time, whereas the space clustering makes sure that we do not simulate thousands of small charge clouds (which might be the output of Geant4), especially if they are all very close together.

src/ChargeClustering/ChargeClustering.jl
sm2909: Could we have a direct StaticVector to CartesianPoint converter in 
`SolidStateDetectors.jl/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl`
so that it is convenient for anyone who wants it.
Or do you plan on not having it completely as I see a comment saying it will be removed where conversions are happening
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a4357bf3ca43ab79d5d7b37518f6eefd2cd5ffab/src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl#L67-L80



posunit is already added in the result as r_pos was declared with
`r_pos = similar(pos, 0)`
before stripping the units of pos

fhagemann: @oschulz ?

sm2909:
> > That would lead to incorrect clustering. Consider this toy example:
> > eps_time = 10s eps_pos = 10m (consider position 1d here for easier visualization)
> > hit 1: 1s, 1m hit 2: 10s, 100m hit 3: 15s, 5m
> > First clustering by time, all hits are in same cluster (eps_time < 10). Then cluster based on position, [1, 3] are a cluster and [2] is another cluster. But [1, 3] should not belong to same cluster as eps_time > 10 between them.
> > (Earlier hit 2 formed the common link between hit 1 and 3 for eps_time. But as finally 2 is not in the same cluster, hit 1 and 3 should not be kept in the same cluster.)
> 
> I see what you mean. The idea behind time clustering would be to separate events that are more than a given `eps_time` apart. This is because, in our pulse shape simulation, we start drifting all charge clouds at the same initial time. Other than with distance, I would also like to make sure that grouped events are never more than `eps_time` apart.
> 
> So, even if the hits were
> 
> ```
> hit 1: 1s, 1m 
> hit 2: 10s, 1m
> hit 3: 15s, 1m
> ```
> 
> and `eps_time` was `10s`, I would like to NEVER group hit 1 and hit3.
> 
> Let`s assume this toy model:
> 
> ```
> for N in 1:1000
> hit (N): (N) ns, 1m
> ```
> 
> If I choose `eps_time = 1u"ns"` because my simulation step is `1u"ns"`, I would group all hits within the `1u"µs"` together, even though this might not be what I want. I, at most, would like the first two hits to be combined and start a new group with the third hit etc.
> 
> Just to think out loud (doesn't mean that this is the best way to go): the time grouping could either start with the first hit and combine everything that's withing `eps_time` after the first hit. OR we can sort all events by time and start grouping with the two hits that are closest together, group them, and then keep on adding more events if they are withing `eps_time` or start with new events afterwards.
> 
> The time clustering should be there to make sure that we only drift hits together, if they actually coincide "enough" in time, whereas the space clustering makes sure that we do not simulate thousands of small charge clouds (which might be the output of Geant4), especially if they are all very close together.


Thanks for clarifying the use-case of this algorithm. 

As we just want to bring together the hits "close enough" in space and time, we might not need dbscan algorithm at all. Rather it could be achieved with a lot simpler algorithms.

For the 1:1000 hits example that you gave, would give same cluster if we do it by dbscan, however we implement it. As long as the hits are dense in time (or space), it will keep on adding. It would lead to huge clusters which is not what we need.

Using the following algorithm would be better in solving our purpose (logic):

```
for each hit:
    if hit already in a cluster: continue
    search all other hits which are not already part of a cluster & lie with eps of hit
    if satisfies above condition: cluster with hit
end
```
 
The implementation for this would be straightforward as well which means runtime will be low.

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -7,61 +7,106 @@
...
-    pos::AbstractVector{<:StaticVector{3,PT}},
-    cluster_radius::RealQuantity
-) where {TT<:RealQuantity, PT <: RealQuantity}
+    pos::AbstractVector{CartesianPoint{PT}},
+    thit::AbstractVector{TTT},
+    cluster_radius::RealQuantity,
+    cluster_time::RealQuantity
+) where {TT<:RealQuantity, PT <: RealQuantity, TTT <: RealQuantity}
+
+    pos = isa(first(pos), CartesianPoint) ? [SVector(p.x, p.y, p.z) for p in pos] : pos     # converting CartesianPoint to SVector
```
fhagemann: Is this still needed then, if we only allow for `pos::AbstractVector{CartesianPoint{PT}}` ?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -7,61 +7,106 @@
...
-                if esum ≈ 0
-                    push!(r_pos, mean(c_hits.pos))
-                else
-                    weights = ustrip.(c_hits.edep) .* inv(esum)
-                    push!(r_pos, sum(c_hits.pos .* weights))
+
+        # sort hits by time
+        t_sort_idx = sortperm(d_hits.thit)
+        d_hits = d_hits[t_sort_idx]
+        t_vals = ustrip.(d_hits.thit)
```
fhagemann: How are you ensuring that the units of `d_hits.thit` match the unit of `cluster_time`?
If `d_hits.thit` has units of `u"ns"` and `cluster_time` is `1u"ms"`, will this still produce correct output?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -7,61 +7,106 @@
...
+                push!(time_groups, current_group)
+                current_group = [i]
+                start_time = t_vals[i]
+            end
+        end
+        push!(time_groups, current_group)
+
+
+        # spatial clustering within each temporal cluster
+        for tg in time_groups
```
fhagemann: Is this something you could export to a function and call in line 52 and 57 instead of pushing to `time_groups`? This might save some allocations.

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -7,61 +7,106 @@
...
+            t_hits = view(d_hits, tg)
+            if length(t_hits) > 3
+                clusters = Clustering.dbscan(ustrip.(flatview(t_hits.pos)), ustripped_cradius;
```
fhagemann: If you know that we always expect `CartesianPoints` now, could this line be adjusted for it?
It would be very good to avoid the conversion to `SVector` and back to `CartesianPoint` at the end of this function if possible.

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -7,61 +7,106 @@
...
+    (detno = r_detno, edep = r_edep, pos = r_pos, thit = r_thit)
 end
 
 
-function cluster_detector_hits(table::TypedTables.Table, cluster_radius::RealQuantity)
+function cluster_detector_hits(
+    table::TypedTables.Table, 
+    cluster_radius::PT, 
+    cluster_time::TTT
+) where {PT <: RealQuantity, TTT <: RealQuantity}
```
fhagemann: Maybe here, ensure that `pos` is converted to `CartesianPoint` before being passed to `cluster_detector_hits`, so that this method does not throw an error if defined with other element types.

test/test_geant4.jl
```diff
@@ -66,6 +66,11 @@ end
     @test evts isa Table
     @test length(evts) == 100
 
+    # Cluster events by time and radius
+    @test_nowarn clustered_evts = SolidStateDetectors.cluster_detector_hits(evts, 10u"µm", 100u"s")
```
fhagemann: And excuse the failing tests, this should be

```suggestion
    clustered_evts = @test_nowarn SolidStateDetectors.cluster_detector_hits(evts, 10u"µm", 100u"s")
```

src/ChargeClustering/ChargeClustering.jl
fhagemann: And it also looks like we don't support `Clustering@0.14` since quite a while ago (#372),
so we can remove the method in line 5 that I introduced to support both `Clustering@0.14` (line 5) and `Clustering@0.15` (line 4) in #332

src/ChargeClustering/ChargeClustering.jl
sm2909: We could do this, but it would require a lot of work in the CartesianPoint type.

We need to implement:
getindex, length, iterate, arithmetic (+, /, *), dot, norm, Statistics.mean, matrix conversion
(all the changes are listed by chatGPT)

this is the bare minimum required to make the function work with CartesianPoint type without conversion.

Then also, it might not be most efficient, as SVector is best optimized at compiler level for the various functions that we are doing here.

src/ChargeClustering/ChargeClustering.jl
sm2909: Unitful.jl automatically handles that. When we do ustrip on eps_time, it is automatically converted to same units as thit before removing units.

Similar had been happening already for pos units as well, I have just copied those things as much as possible.

sm2909:
1. Added some more test: checking eltype of pos, sum of all edep, no NaN in output table
2. push! in previous has been moved to helper function to save allocations
3. pos type is converted to CartesianPoint in the function (with table input) to avoid errors.
4. Removed Clustering@0.14 supporter method as you suggested.

src/ChargeClustering/ChargeClustering.jl
oschulz: We purposefully removed support for arithmetic for points, since such operation don't exist in affine spaces in maths. GeometryBasics.jl still does such things, unfortunately, but modern packages like Meshes.jl also to these things "properly" now (can't use Meshes.jl as the basics for SolidStateDetector.jl currently though since they forbid unitless values).

Clustering itself is, of course, well defined for points, since it should be independent of the choice of origin. So if the clustering algorithm needs such operations, they actually should be applied to vectors from some arbitrary origin (typically just the point at zero).

sm2909: Thanks for the insight! I didn’t realize this had already been considered — really appreciate the clarification.

fhagemann:
Ok, sorry for making this chaotic, but in PR #558, I added code that would make sure that the OLD implementation of `cluster_detector_hits` would work and merged this into main.
This also included some clever way to deal with both `CartesianPoint` and `StaticVector` as `pos` column of the `Table`.
https://github.com/JuliaPhysics/SolidStateDetectors.jl/commit/d6a62f381b3e134060519d44b05a54be2a50fe72

Could you build your changes on top of this, and see if the tests require changes?

fhagemann:
And I think we might still need to tackle the issue with the runtime to avoid this from taking forever to run.
Check how fast it was before this change, after this change, and maybe with some default (`cluster_time = Inf * u"s"`) that should result in the same behavior as before, and how long that takes.

But we're getting there!

sm2909:
Benchmarks before change:

```
 @time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm")
  4.338386 seconds (6.55 M allocations: 384.269 MiB, 3.01% gc time, 96.09% compilation time)
```

Benchmarks after change (clustering by time as well):

```
@time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm", Inf *u"s")
  5.322641 seconds (8.62 M allocations: 483.518 MiB, 6.93% gc time, 97.33% compilation time)
```

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -1,64 +1,117 @@
...
-    ) where {TT <: RealQuantity, PT <: RealQuantity}
+    detno::AbstractVector{<:Integer},
+    edep::AbstractVector{TT},
+    pos::AbstractVector{CartesianPoint{PT}},
+    thit::AbstractVector{TTT},
+    cluster_radius::RealQuantity,
+    cluster_time::RealQuantity
+) where {TT<:RealQuantity, PT <: RealQuantity, TTT <: RealQuantity}
+
+    pos = [SVector(p.x, p.y, p.z) for p in pos]   # converting CartesianPoint to SVector
```
fhagemann: Just curious: is this line still needed?
Because if not, removing it might fix the failing tests.

fhagemann: --> It is not needed anymore and I removed it in 830eca8

src/ChargeClustering/ChargeClustering.jl
fhagemann: --> no need to convert if the base method `cluster_detector_hits` allows for both `CartesianPoint` and `StaticVector` (see a8c2666)

src/ChargeClustering/ChargeClustering.jl
fhagemann: --> Allow for both `CartesianPoint` and `StaticVector` (see a8c2666)

test/test_geant4.jl
```diff
@@ -63,31 +63,56 @@ end
...
     @test eltype(first(clustered_evts.pos)) <: CartesianPoint
 
-    # Generate waveforms
-    simulate!(sim, refinement_limits = [0.2,0.1,0.05,0.03,0.02])
-    wf = simulate_waveforms(evts, sim, Δt = 1u"ns", max_nsteps = 2000)
-    @test wf isa Table
-    @test :waveform in columnnames(wf)
-    @test length(wf) == length(evts) * sum(.!ismissing.(sim.weighting_potentials))
+    # Cluster events by radius, but split in time
+    clustered_evts = SolidStateDetectors.cluster_detector_hits(evts, 10u"µm", 0u"ns")
```
fhagemann: FYI: This was the test that revealed the typo bug (`d_hits` instead of `t_hits`)

fhagemann (dismissed):
I pushed some minor changes (a typo bugfix, and support for both `CartesianPoint` and `SVector` as type of the `pos` column of the table).

It might make sense to do a final check that the time splitting performs as intended (also looking at energy spectra etc.) and adding this to the tests if wanted.]
And we might also (in a separate PR) look into performance in a bit more detail to further decrease the runtime of this if possible.

But in general, this PR looks good to me now!

@sm2909 thanks for the initiative!

test/test_geant4.jl
Kunming-CN: Excellent! Thank you!

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -1,64 +1,115 @@
...
-        else
-            append!(r_detno, d_hits.detno)
-            append!(r_edep, d_hits.edep)
-            append!(r_pos, d_hits.pos)
         end
     end
 
-    (detno = r_detno, edep = r_edep, pos = r_pos)
+    (detno = r_detno, edep = r_edep, pos = r_pos, thit = r_thit)
 end
```
fhagemann: We could remove `time_groups` and `finalize_group!` by doing what the second `for`-loop does whenever you would `push!` something to `time_groups`:
```suggestion
function cluster_detector_hits(
    detno::AbstractVector{<:Integer},
    edep::AbstractVector{TT},
    pos::AbstractVector{<:Union{<:StaticVector{3,PT}, <:CartesianPoint{PT}}},
    thit::AbstractVector{TTT},
    cluster_radius::RealQuantity,
    cluster_time::RealQuantity
) where {TT<:RealQuantity, PT <: RealQuantity, TTT <: RealQuantity}

    unsorted = TypedTables.Table(detno = detno, edep = edep, pos = pos, thit = thit)
    sorting_idxs = sortperm(unsorted.detno)
    sorted = view(unsorted, sorting_idxs)
    grouped = TypedTables.Table(consgroupedview(sorted.detno, TypedTables.columns(sorted)))

    r_detno = similar(detno, 0)
    r_edep = similar(edep, 0)
    r_pos = similar(pos, 0)
    r_thit = similar(thit, 0)

    ustripped_cradius = ustrip(internal_length_unit, cluster_radius)

    for d_hits_nt in grouped

        d_hits = TypedTables.Table(d_hits_nt)

        # sort hits by time
        t_sort_idx = sortperm(d_hits.thit)
        d_hits = view(d_hits, t_sort_idx)
        t_vals = d_hits.thit

        current_group = 1
        @inbounds for i in eachindex(t_vals) .+ 1
            if i > lastindex(t_vals) || t_vals[i] - t_vals[current_group] > cluster_time
                t_hits = view(d_hits, current_group:i-1)
                if length(t_hits) > 3
                    d_detno = first(t_hits.detno)
                    @assert all(isequal(d_detno), t_hits.detno)
                    clusters = Clustering.dbscan(hcat((ustrip.(internal_length_unit, getindex.(t_hits.pos,i)) for i in 1:3)...)', 
                        ustripped_cradius, leafsize = 20, min_neighbors = 1, min_cluster_size = 1).clusters
                    
                    for c in clusters
                        idxs = vcat(c.boundary_indices, c.core_indices)
                        @assert length(idxs) == c.size
                        c_hits = view(t_hits, idxs)
                        esum = sum(c_hits.edep)

                        push!(r_detno, d_detno)
                        push!(r_edep, esum)
                        if esum ≈ zero(TT)
                            push!(r_pos, barycenter(c_hits.pos))
                            push!(r_thit, mean(c_hits.thit))
                        else
                            weights = ustrip.(Unitful.NoUnits, c_hits.edep .* inv(esum))
                            push!(r_pos, barycenter(c_hits.pos, StatsBase.Weights(weights)))
                            push!(r_thit, sum(c_hits.thit .* weights))
                        end
                    end
                else
                    append!(r_detno, t_hits.detno)
                    append!(r_edep, t_hits.edep)
                    append!(r_pos, t_hits.pos)
                    append!(r_thit, t_hits.thit)
                end
                current_group = i
            end
        end
    end

    (detno = r_detno, edep = r_edep, pos = r_pos, thit = r_thit)
end
```

sm2909: I have implemented these changes that you have suggested

`clusters = Clustering.dbscan(hcat((ustrip.(internal_length_unit, getindex.(t_hits.pos,i)) for i in 1:3)...)'`

in this line `internal_length_unit` was causing unit mismatch with cluster_radius, so I have used the `posunit` from cluster_radius while stripping units of `pos` and `cluster_radius` to maintain consistency

fhagemann: But what if we `ustrip` the `cluster_radius` to the same `internal_length_unit`?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -28,9 +23,6 @@ function cluster_detector_hits(
     posunit = unit(PT)
     ustripped_cradius = ustrip(posunit, cluster_radius)
```
fhagemann: No need to define `posunit` if we do everything in internal units (`m`):
```suggestion
    ustripped_cradius = ustrip(internal_length_unit, cluster_radius)
```

src/ChargeClustering/ChargeClustering.jl
fhagemann, line 1: Nice! Can you also give numbers on how long this takes to run (maybe compared with the previous implementation?)

sm2909: Benchmarks before change (clustering without time):

```
 @time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm")
  4.338386 seconds (6.55 M allocations: 384.269 MiB, 3.01% gc time, 96.09% compilation time)
```

Benchmarks after current change :

```
@time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm", 100u"s")
  5.116692 seconds (9.16 M allocations: 506.285 MiB, 8.49% gc time, 93.41% compilation time)
```

fhagemann: For a table with how many events initially? 
Same as here?
> 1000 rows
> 
> These are the exact commands that I ran:
> 
> ```
> using SolidStateDetectors, Geant4, Unitful
> sim = Simulation{Float32}(SSD_examples[:InvertedCoax])
> source = IsotopeSource(90, 228, 0.0, 0.0, CartesianPoint(0.05,0,0.05))
> app = G4JLApplication(sim, source)
> t = run_geant4_simulation(app, 1_000)
> @time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm", 100u"s")
> ```
 ?

And does that number change if `cluster_time` is small?

sm2909: 1000 initial events (same commands that I ran before)

It takes roughly same time and resources on all values of cluster_time
here I took for 100μs instead of 100s:

```
@time tc = SolidStateDetectors.cluster_detector_hits(t, 20u"μm", 100u"μs")
  4.951178 seconds (9.39 M allocations: 516.259 MiB, 3.76% gc time, 98.09% compilation time)
```

fhagemann: And what about something like `0u"ns"` or something negative, where more splitting might be happening?

sm2909: works same for 0u"ns"

negative time cases are not handled
output table is generated in same runtime, but it is wrong

sm2909: For the conflicts:

For ChargeClustering.jl, all the changes from my branch should be taken as the changes in the main codebase does not use clustering based on time.

You will need to look in the tests file. There is some conflict between the changes you made on this branch versus the recent new code in the main codebase

fhagemann: Do you want me to resolve all conflicts? :)

test/test_geant4.jl
```diff
@@ -110,15 +142,20 @@ end
...
     evts_static = Table(evts; pos = VectorOfVectors(broadcast.(p -> SVector{3}(p.x, p.y, p.z) * 1000u"mm", SolidStateDetectors.to_internal_units(evts.pos))))
-    clustered_evts_static = SolidStateDetectors.cluster_detector_hits(evts_static, 10u"µm")
+    clustered_evts_static = SolidStateDetectors.cluster_detector_hits(evts_static, 10u"µm", 1u"ns")
     @test length(clustered_evts_static) == length(evts_static)
     @test length(flatview(clustered_evts_static.pos)) <= length(flatview(evts_static.pos))
     @test eltype(first(clustered_evts_static.pos)) <: StaticVector{3}
 
     # Check clustering with cluster radius being unitless or with the wrong unit
-    clustered_evts_static_unitless = SolidStateDetectors.cluster_detector_hits(evts_static, 1e-5)
+    clustered_evts_static_unitless = SolidStateDetectors.cluster_detector_hits(evts_static, 1e-5, 1u"ns")
```
sm2909: In earlier version of tests, it still used the function where clustering was happening only based on space and not time. The function signature has changed now, so I have modified these tests here.
Also I added corresponding tests for incorrect units of cluster_time.

src/ChargeClustering/ChargeClustering.jl
fhagemann: I re-resolved the merge conflicts to account for this, let's see if that works now.

fhagemann:
Yes, this is why it's prefered to use `to_internal_units` on `CartesianPoint` (or `SVector`, depending on which it is).
I undid your merge commit and re-merged while resolving the merge conflicts.

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -77,19 +76,12 @@ function cluster_detector_hits(
...
-) where {PT <: RealQuantity, TTT <: RealQuantity}
-    @assert :pos in TypedTables.columnnames(table) "Table has no column `pos`"
-    @assert :thit in TypedTables.columnnames(table) "Table has no column `thit`"
-    @assert :edep in TypedTables.columnnames(table) "Table has no column `edep`"
-    @assert :detno in TypedTables.columnnames(table) "Table has no column `detno`"
+function cluster_detector_hits(table::TypedTables.Table, cluster_radius::RealQuantity, cluster_time::RealQuantity = Inf * u"s")
+    @assert is_detector_hits_table(table) "Table does not have the correct format"
     
     # Collect all time-clustered results
     all_results = []
```
fhagemann: We should avoid declaring `Arrays` with unknown element-type, as this will default to `Any` and not be type-stable.

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -80,7 +80,19 @@ function cluster_detector_hits(table::TypedTables.Table, cluster_radius::RealQua
     @assert is_detector_hits_table(table) "Table does not have the correct format"
     
     # Collect all time-clustered results
-    all_results = []
+    @assert !isempty(table) "Input table is empty"
+    firstrow = first(table)
```
fhagemann: We might not even need this, as the check `is_detector_hits_table` will check whether certain columns exist and that they have the correct type. Can you work with `table` instead of `firstrow` (i.e. inferring `eltypes` from `table` directly)?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -2,69 +2,123 @@
...
+    end
+    
+    # Build output table
+    result_columns = (
+        evtno = evtno_col,
+        detno = VectorOfVectors([r.detno for r in all_results]),
+        thit = VectorOfVectors([r.thit for r in all_results]),
+        edep = VectorOfVectors([r.edep for r in all_results]),
+        pos = VectorOfVectors([r.pos for r in all_results])
+    )
```
fhagemann: How do we deal with events that had more columns initially? Can we do the same thing we did for `evtno` to just clone the entries if a row was split into multiple rows by the clustering?
Maybe then, `ResultType` can just be the similar type as the original `Table` (?)
You can check how we used to do it for the original `Table` (merging the "old" Table with the new one)

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -77,13 +77,16 @@ end
...
 
 function cluster_detector_hits(table::TypedTables.Table, cluster_radius::RealQuantity, cluster_time::RealQuantity = Inf * u"s")
-    @assert is_detector_hits_table(table) "Table does not have the correct format"
     
-    # Collect all time-clustered results
-    all_results = []
-    evtno_col = Int[]
-    
-    for (evtno, evt) in enumerate(table)
+    @assert is_detector_hits_table(table)
```
fhagemann: I would like to keep the assertion message that we had before:
`"Table does not have the correct format"` (or also happy to discuss another message that's more descriptive)

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -93,20 +96,18 @@ function cluster_detector_hits(table::TypedTables.Table, cluster_radius::RealQua
...
+        end
     end
-    
-    # Build output table
-    result_columns = (
-        evtno = evtno_col,
-        detno = VectorOfVectors([r.detno for r in all_results]),
-        thit = VectorOfVectors([r.thit for r in all_results]),
-        edep = VectorOfVectors([r.edep for r in all_results]),
-        pos = VectorOfVectors([r.pos for r in all_results])
```
fhagemann: Do we need to convert the new columns into `VectorOfVectors` or is that automatically done by the `similar` when defining `col_vectors`? Do we test somewhere in the tests that the new column types are `VectorOfVectors`?

Issue #183, opened 2021-07-15
Proper handling of units in the user interface


fhagemann:
It would be nice to allow for units in the user interface, especially for plotting, e.g.
```julia
plot(simulation.electric_field, xlims = (-4u"cm", 4u"cm"))
```
This also comes with nice unit handling in configuration files (see #182).

oschulz:
We should take a look at [UnitfulRecipes.jl](https://github.com/jw3126/UnitfulRecipes.jl).

fhagemann:
I just realized this is a duplicate of #66. Closing #66 in favour of this.

lmh91:
Regarding the input and output units of the signal generation.

I think we should return the signal in units of (induced) charge and not "energy" anymore.
The conversion from charge to energy is done by the electronics which is not part of SSD.
The conversion from the energy of the hit table (input) to number of charge carriers 
should happen before the drift is simulated as the self-repulsion needs charge.

oschulz:
Sounds good.

fhagemann:
Adding [UnitfulRecipes.jl](https://github.com/jw3126/UnitfulRecipes.jl) as additional dependency fixes the plotting issue
(we can then just pass values with units to plot and they will automatically be added to the axes labels, quite neat!)

```julia
using Plots, Unitful, UnitfulRecipes
plot([1,2]u"m", [1,2]u"m", xlabel = "x", ylabel = "y", legend = false)
```
![Plot1](https://user-images.githubusercontent.com/30291312/135977722-41d38323-fdda-40db-afef-807442568346.png)
If you add more stuff with other units, it will keep the units from the first `plot` command
```julia
plot!([1200,1300]u"mm", [1000,1100]u"mm")
```
![Plot2](https://user-images.githubusercontent.com/30291312/135977763-f209e0e0-955a-4fd2-b652-30f66557a63c.png)
So, we can just add the internal units to all the objects/values when we plot them.

The question is how we want to deal with external units... Should we allow for a keyword argument in the plot recipe, something like `plot(..., length_unit = u"mm")`?

claudiaalvgar:
> [@claudiaalvgar](https://github.com/claudiaalvgar) could you take care of this?

Sure, I'll have a look at it

fhagemann:
[...]

PR #560, opened 2025-11-10
Create new datatype when saving `CartesianPoint` to HDF5 using `LegendHDF5IO`

claudiaalvgar:
This PR allows for ```CartesianPoint``` and ```CylindricalPoint``` with and without units from #489 to be passed to ```Event``` and ```NBodyChargeCloud``` ( issue in #183 ). Test code and modes allowed: 

```
# Passing 1 point in Cartesian or Cylindrical
pt = CylindricalPoint(2, deg2rad(20), 1)
e = 10u"keV"
evt1 = Event(pt)
evt2 = Event(pt, e)
println("1-Cyl no units: \n$(evt1), \n$(evt2)\n\n")

pt = CartesianPoint(1, 1.3, 1.2)
evt1 = Event(pt)
evt2 = Event(pt, e)
println("2-Cart no units: \n$(evt1), \n$(evt2)\n\n")

T=Float16
pt = CartesianPoint{T}(1, 1.3, 1.2)
evt1 = Event(pt)
evt2 = Event(pt, e)
println("3-Cart no units, Float16: \n$(evt1), \n$(evt2)\n\n")

pt = CylindricalPoint{T}(1, 1.3, 1.2)
evt1 = Event(pt)
evt2 = Event(pt, e)
println("3.1-Cyl no units, Float16: \n$(evt1), \n$(evt2)\n\n")

T=Float32
pt = CartesianPoint{T}(1, 1.3, 1.2)
evt1 = Event(pt)
evt2 = Event(pt, e)
println("4-Cart no units, Float32: \n$(evt1), \n$(evt2)\n\n")

pt = CylindricalPoint{T}(1, 1.3, 1.2)
evt1 = Event(pt)
evt2 = Event(pt, e)
println("4.1-Cyl no units, Float32: \n$(evt1), \n$(evt2)\n\n")


# Passing 1 point in Cartesian or Cylindrical with units
pt = CartesianPoint(1u"mm", 2u"mm", 3u"mm")
evt1 = Event(pt)
evt2 = Event(pt, e)
println("5-Cart units: \n$(evt1), \n$(evt2)\n\n")

pt = CylindricalPoint(5u"mm", 9u"rad", 2u"mm")
evt1 = Event(pt)
evt2 = Event(pt, e)
println("6-Cyl units: \n$(evt1), \n$(evt2)\n\n")

# Passing 1 vector in Cartesian or Cylindrical
pt1 =  CartesianPoint(1.0, 2.0, 3.0)
pt2 =  CartesianPoint(4.0, 5.0, 6.0)
energy1 = 1.0u"MeV"
energy2 = 2.0u"MeV"
evt1 = Event([pt1,pt2])
evt2 = Event([pt1,pt2], [energy1, energy2])
println("7-Cart no units: \n$(evt1), \n$(evt2)\n\n")

pt1 = CylindricalPoint(5, 9, 2)
pt2 = CylindricalPoint(1, 1.0, 0.0)
evt1 = Event([pt1,pt2])
evt2 = Event([pt1,pt2], [energy1, energy2])
println("8-Cyl no units: \n$(evt1), \n$(evt2)\n\n")

# Passing 1 vector in Cartesian or Cylindrical with units
pt1 =  CartesianPoint(1.0u"mm", 2.0u"mm", 3.0u"mm")
pt2 =  CartesianPoint(4.0u"mm", 5.0u"mm", 6.0u"mm")
evt1 = Event([pt1,pt2])
evt2 = Event([pt1,pt2], [energy1, energy2])
println("9-Cart with units: \n$(evt1), \n$(evt2)\n\n")

pt1 = CylindricalPoint(5u"mm", 9u"rad", 2u"mm")
pt2 = CylindricalPoint(1u"mm", 1.0u"rad", 0.0u"mm")
evt1 = Event([pt1,pt2])
evt2 = Event([pt1,pt2], [energy1, energy2])
println("10-Cyl with units: \n$(evt1), \n$(evt2)\n\n")

# Passing 1 vector of vectors in Cartesian or Cylindrical
pt1 = CartesianPoint(1.0, 2.0, 3.0)
pt2 = CartesianPoint(4.0, 5.0, 6.0)
pt3 = CartesianPoint(7.0, 8.0, 9.0)
pt4 = CartesianPoint(10.0, 11.0, 12.0)

locations = [
    [pt1, pt2],  # first "event"
    [pt3, pt4]   # second "event"
]
energy1 = 1.0u"MeV"
energy2 = 2.0u"MeV"
energy3 = 1.5u"MeV"
energy4 = 2.5u"MeV"

energies = [
    [energy1, energy2],
    [energy3, energy4]
]
evt1 = Event(locations)
evt2 = Event(locations, energies)
println("11-Cart no units: \n$(evt1), \n$(evt2)\n\n")

pt1 = CylindricalPoint(1.0, 2.0, 3.0)
pt2 = CylindricalPoint(4.0, 5.0, 6.0)
pt3 = CylindricalPoint(7.0, 8.0, 9.0)
pt4 = CylindricalPoint(10.0, 11.0, 12.0)

locations = [
    [pt1, pt2],  # first "event"
    [pt3, pt4]   # second "event"
]

energies = [
    [energy1, energy2],
    [energy3, energy4]
]
evt1 = Event(locations)
evt2 = Event(locations, energies)
println("12-Cyl no units: \n$(evt1), \n$(evt2)\n\n")

# Passing 1 vector of vectors in Cartesian or Cylindrical with units
pt1 = CartesianPoint(1.0u"mm", 2.0u"mm", 3.0u"mm")
pt2 = CartesianPoint(4.0u"mm", 5.0u"mm", 6.0u"mm")
pt3 = CartesianPoint(7.0u"mm", 8.0u"mm", 9.0u"mm")
pt4 = CartesianPoint(10.0u"mm", 11.0u"mm", 12.0u"mm")

locations = [
    [pt1, pt2],  # first "event"
    [pt3, pt4]   # second "event"
]

energies = [
    [energy1, energy2],
    [energy3, energy4]
]
evt1 = Event(locations)
evt2 = Event(locations, energies)
println("13-Cart units: \n$(evt1), \n$(evt2)\n\n")

pt1 = CylindricalPoint(1.0u"mm", 2.0u"rad", 3.0u"mm")
pt2 = CylindricalPoint(4.0u"mm", 5.0u"rad", 6.0u"mm")
pt3 = CylindricalPoint(7.0u"mm", 8.0u"rad", 9.0u"mm")
pt4 = CylindricalPoint(10.0u"mm", 11.0u"rad", 12.0u"mm")

locations = [
    [pt1, pt2],  # first "event"
    [pt3, pt4]   # second "event"
]

energies = [
    [energy1, energy2],
    [energy3, energy4]
]
evt1 = Event(locations)
evt2 = Event(locations, energies)
println("14-Cyl units: \n$(evt1), \n$(evt2)\n\n")

#NBodyChargeCloud point no units
center = CartesianPoint(0.2,0.1,0.5)
energy = 1460u"keV"
nbcc = NBodyChargeCloud(center, energy)
evt = Event(nbcc)
println("15-Cart no units: \n$(evt)\n\n")

center = CartesianPoint(0.2,0.1,0.5)
energy = 1460u"keV"
nbcc = NBodyChargeCloud(center, energy, 100)
evt = Event(nbcc)
println("16-Cart no units: \n$(evt)\n\n")

center = CylindricalPoint(0.2,0.1,0.5)
nbcc = NBodyChargeCloud(center, energy)
evt = Event(nbcc)
println("17-Cyl no units: \n$(evt)\n\n")

#NBodyChargeCloud point units
center = CartesianPoint(0.2u"mm",0.3u"mm",5.0u"mm")
nbcc = NBodyChargeCloud(center, energy)
evt = Event(nbcc)
println("18-Cart units: \n$(evt)\n\n")

center = CartesianPoint(0.2u"mm",0.3u"mm",5.0u"mm")
nbcc = NBodyChargeCloud(center, energy, 100)
evt = Event(nbcc)
println("19-Cart units: \n$(evt)\n\n")


center = CylindricalPoint(0.2u"mm",0.0u"rad",2.1u"mm")
nbcc = NBodyChargeCloud(center, energy)
evt = Event(nbcc)
println("20-Cyl units: \n$(evt)\n\n")
```

### DIFF
--- a/test/test_io.jl
+++ b/test/test_io.jl
@@ -64,51 +64,55 @@ end
     sim = Simulation{T}(SSD_examples[:InvertedCoax])
     timed_simulate!(sim, convergence_limit = 1e-5, device_array_type = device_array_type, refinement_limits = [0.2, 0.1], verbose = false)
 
-    evt_table = Table(
-        evtno = Int32[1], 
-        detno = Int32[1],
-        thit = VectorOfVectors([T[0] * u"s"]),
-        edep = VectorOfVectors([T[1] * u"eV"]),
-        pos = VectorOfVectors([[SVector{3, T}.(0.01, 0.01, 0.01) * u"m"]])
-    )
-
-    contact_charge_signals = timed_simulate_waveforms(      
-            evt_table,
-            sim,
-            max_nsteps = 4000, 
-            Δt = 1u"ns", 
-            number_of_carriers = 20,
-            number_of_shells = 2,
-            geometry_check = true,
-            verbose = false);
-    
-    @timed_testset "Regular simulate_waveforms" begin
-        signalsum = T(0)
-        for i in 1:length(contact_charge_signals.waveform)
-            signalsum += abs(ustrip(contact_charge_signals.waveform[i].signal[end]))
-        end
-        signalsum *= inv(ustrip(SolidStateDetectors._convert_internal_energy_to_external_charge(sim.detector.semiconductor.material)))
-        @test isapprox( signalsum, T(2), atol = 5e-3 )
-    end
-
-    @timed_testset "LegendHDF5IO simulate_waveforms" begin
-        timed_simulate_waveforms(      
-            evt_table,
-            sim,
-            ".",
-            chunk_n_physics_events = 1,
-            max_nsteps = 4000, 
-            Δt = 1u"ns", 
-            number_of_carriers = 20,
-            number_of_shells = 2,
-            geometry_check = true,
-            verbose = false
+    # test I/O for both SVector (with and without units) and CartesianPoint
+    for pos in ([SVector{3, T}.(10, 10, 10) * u"mm"], [SVector{3, T}.(0.01, 0.01, 0.01)], [CartesianPoint{T}(0.01, 0.01, 0.01)])
+        
+        evt_table = Table(
+            evtno = Int32[1], 
+            detno = Int32[1],
+            thit = VectorOfVectors([T[0] * u"s"]),
+            edep = VectorOfVectors([T[1] * u"eV"]),
+            pos = VectorOfVectors([pos])
         )
-        @info isfile("generated_waveforms_evts_1-1.h5")
-        @test contact_charge_signals == LegendHDF5IO.lh5open("generated_waveforms_evts_1-1.h5") do h5f
-            @test haskey(h5f, "generated_waveforms")
-            LegendHDF5IO.readdata(h5f.data_store, "generated_waveforms")
+
+        contact_charge_signals = timed_simulate_waveforms(      
+                evt_table,
+                sim,
+                max_nsteps = 4000, 
+                Δt = 1u"ns", 
+                number_of_carriers = 20,
+                number_of_shells = 2,
+                geometry_check = true,
+                verbose = false);
+        
+        @timed_testset "Regular simulate_waveforms" begin
+            signalsum = T(0)
+            for i in 1:length(contact_charge_signals.waveform)
+                signalsum += abs(ustrip(contact_charge_signals.waveform[i].signal[end]))
+            end
+            signalsum *= inv(ustrip(SolidStateDetectors._convert_internal_energy_to_external_charge(sim.detector.semiconductor.material)))
+            @test isapprox( signalsum, T(2), atol = 5e-3 )
         end
-        rm("generated_waveforms_evts_1-1.h5")
+
+        @timed_testset "LegendHDF5IO simulate_waveforms" begin
+            timed_simulate_waveforms(      
+                evt_table,
+                sim,
+                ".",
+                chunk_n_physics_events = 1,
+                max_nsteps = 4000, 
+                Δt = 1u"ns", 
+                number_of_carriers = 20,
+                number_of_shells = 2,
+                geometry_check = true,
+                verbose = false
+            )
+            @info isfile("generated_waveforms_evts_1-1.h5")
+            @test contact_charge_signals == LegendHDF5IO.lh5open("generated_waveforms_evts_1-1.h5") do h5f
+                @test haskey(h5f, "generated_waveforms")
+                LegendHDF5IO.readdata(h5f.data_store, "generated_waveforms")
+            end
+            rm("generated_waveforms_evts_1-1.h5")
+        end
     end
 end
--- a/ext/SolidStateDetectorsLegendHDF5IOExt.jl
+++ b/ext/SolidStateDetectorsLegendHDF5IOExt.jl
@@ -57,12 +57,12 @@ function SolidStateDetectors.simulate_waveforms( mcevents::AbstractVector{<:Name
         @info "Now simulating $(evtrange) and storing it in\n\t \"$ofn\""
         mcevents_sub = simulate_waveforms(mcevents[evtrange], sim; Δt, max_nsteps, diffusion, self_repulsion, number_of_carriers, number_of_shells, signal_unit, max_interaction_distance, end_drift_when_no_field, geometry_check, verbose)
 
-        # LH5 can't handle CartesianPoint, turn positions into CartesianVectors which will be saved as SVectors
-        pos_vec = VectorOfVectors([[CartesianPoint(p...) - cartesian_zero for p in ps] for ps in mcevents_sub.pos])
-        new_mcevents_sub = TypedTables.Table(merge(Tables.columns(mcevents_sub), (pos = pos_vec,)))
+        # # LH5 couldn't handle CartesianPoint, turn positions into CartesianVectors which will be saved as SVectors
+        # pos_vec = VectorOfVectors([[CartesianPoint(p...) - cartesian_zero for p in ps] for ps in mcevents_sub.pos])
+        # new_mcevents_sub = TypedTables.Table(merge(Tables.columns(mcevents_sub), (pos = pos_vec,)))
 
         LegendHDF5IO.lh5open(ofn, "w") do h5f
-            LegendHDF5IO.writedata(h5f.data_store, "generated_waveforms", new_mcevents_sub)
+            LegendHDF5IO.writedata(h5f.data_store, "generated_waveforms", mcevents_sub)
         end        
     end    
 end


src/ChargeCloudModels/ChargeCloudModels.jl
```diff
@@ -52,6 +52,24 @@ NBodyChargeCloud(center, energy, number_of_shells = 3, shell_structure = SolidSt
...
+                          kwargs...)
+    internal_center = to_internal_point(center)
+    return NBodyChargeCloud(internal_center, energy; kwargs...)
+end
+
+function NBodyChargeCloud(center::CylindricalPoint{<:SSDFloat}, energy::RealQuantity;
+                          kwargs...)
+    internal_center = to_internal_point(center)
+    return NBodyChargeCloud(internal_center, energy; kwargs...)
+end
```
fhagemann: This looks to me like it could be unified:

```suggestion
function NBodyChargeCloud(center::CylindricalPoint{Union{<:SSDFloat, <:Unitful.Quantity}}, energy::RealQuantity;
                          kwargs...)
    internal_center = to_internal_point(center)
    return NBodyChargeCloud(internal_center, energy; kwargs...)
end
```

or maybe even

```suggestion
function NBodyChargeCloud(center::CylindricalPoint, energy::RealQuantity;
                          kwargs...)
    internal_center = to_internal_point(center)
    return NBodyChargeCloud(internal_center, energy; kwargs...)
end
```

or

```suggestion
function NBodyChargeCloud(center::AbstractCoordinatePoint, energy::RealQuantity;
                          kwargs...)
    internal_center = to_internal_point(center)
    return NBodyChargeCloud(internal_center, energy; kwargs...)
end
```
?

src/ChargeCloudModels/ChargeCloudModels.jl
```diff
@@ -121,6 +139,24 @@ NBodyChargeCloud(center, energy, 200, number_of_shells = 3)
...
+
+function NBodyChargeCloud(center::CylindricalPoint{<:SSDFloat}, energy::RealQuantity, N::Integer;
+                          kwargs...)
+    internal_center = to_internal_point(center)
+    return NBodyChargeCloud(internal_center, energy; kwargs...)
+end
+
```
fhagemann: `N` doesn't seem to be passed... Is this intended?

Also, can these methods be combined?

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -244,6 +244,17 @@ struct CylindricalPoint{T} <: AbstractCylindricalPoint{T}
     CylindricalPoint{T}(r::Real, φ::Real, z::Real) where {T} = new(T(r), mod(T(φ),T(2π)), T(z))
 end
 
+# Handles Unitful quantities or plain numbers 
+function CylindricalPoint(r, φ, z)
+    r_val = isa(r, Unitful.Quantity) ? float(ustrip(uconvert(u"m", r))) : float(r)
+    φ_val = isa(φ, Unitful.Quantity) ? float(ustrip(uconvert(u"rad", φ))) : float(φ)
+    z_val = isa(z, Unitful.Quantity) ? float(ustrip(uconvert(u"m", z))) : float(z)
```
fhagemann: `ustrip(unit, x)` is the same as `ustrip(uconvert(unit,x))` and also use internal units where intended:
```suggestion
    r_val = isa(r, Unitful.Quantity) ? float(ustrip(internal_length_unit, r)) : float(r)
    φ_val = isa(φ, Unitful.Quantity) ? float(ustrip(internal_angle_unit, φ)) : float(φ)
    z_val = isa(z, Unitful.Quantity) ? float(ustrip(internal_length_unit, z)) : float(z)
```

fhagemann: Also, is `float` what we want?
If one of them is an `Int`, `float` will make this a `Float64`

I would argue, though, that if you pass `Int, Float32, Float32`, you would like your final point to be of type `Float32` rather than `Float64`.

src/Event/Event.jl
```diff
@@ -1,6 +1,30 @@
...
+        T = pt.r isa Unitful.Quantity ? float(typeof(ustrip(pt.r))) : float(typeof(pt.r))
+        r = pt.r isa Unitful.Quantity ? float(to_internal_units(pt.r)) : float(pt.r)
+        φ = pt.φ isa Unitful.Quantity ? float(ustrip(uconvert(u"rad", pt.φ))) : float(pt.φ)
+        z = pt.z isa Unitful.Quantity ? float(to_internal_units(pt.z)) : float(pt.z)
+        return CartesianPoint{T}(r * cos(φ), r * sin(φ), z)
+        
+     else
+        error("Unsupported point type $(typeof(pt)). Expected CartesianPoint or CylindricalPoint.")
+     end
+end
```
fhagemann: Split this into several methods?


```suggestion
function to_internal_point(pt::CartesianPoint}
    T = pt.x isa Unitful.Quantity ? float(typeof(ustrip(pt.x))) : typeof(pt.x)
    x = pt.x isa Unitful.Quantity ? float(to_internal_units(pt.x)) : pt.x
    y = pt.y isa Unitful.Quantity ? float(to_internal_units(pt.y)) : pt.y
    z = pt.z isa Unitful.Quantity ? float(to_internal_units(pt.z)) : pt.z
    return CartesianPoint{T}(x, y, z)
end

function to_internal_point(pt::CylindricalPoint)
    T = pt.r isa Unitful.Quantity ? float(typeof(ustrip(pt.r))) : float(typeof(pt.r))
    r = pt.r isa Unitful.Quantity ? float(to_internal_units(pt.r)) : float(pt.r)
    φ = pt.φ isa Unitful.Quantity ? float(ustrip(uconvert(u"rad", pt.φ))) : float(pt.φ)
    z = pt.z isa Unitful.Quantity ? float(to_internal_units(pt.z)) : float(pt.z)
    return CartesianPoint{T}(r * cos(φ), r * sin(φ), z)
end

function to_internal_point(pt::AbstractCoordinatePoint)
    error("Unsupported point type $(typeof(pt)). Expected CartesianPoint or CylindricalPoint.")
end
```

src/Event/Event.jl
```diff
@@ -19,30 +43,43 @@ mutable struct Event{T <: SSDFloat}
...
-    evt.locations = VectorOfArrays([[CartesianPoint(location)]])
-    evt.energies = VectorOfArrays([[T(to_internal_units(energy))]])
+function Event(location::AbstractCoordinatePoint{T}, energy::RealQuantity = 1.0) where {T <: RealQuantity}
+    pt_internal = to_internal_point(location)
+    T_ = typeof(pt_internal.x)
+    evt = Event{T_}()
+    evt.locations = VectorOfArrays([[pt_internal]])
+    evt.energies  = VectorOfArrays([[T_(to_internal_units(energy))]])
     return evt
 end
```
fhagemann: Looks to me like you don't need to define `T` here in the argument list and can then rename `T_` to `T`:
```suggestion
function Event(location::AbstractCoordinatePoint, energy::RealQuantity = 1.0)
    pt_internal = to_internal_point(location)
    T = typeof(pt_internal.x)
    evt = Event{T}()
    evt.locations = VectorOfArrays([[pt_internal]])
    evt.energies  = VectorOfArrays([[T(to_internal_units(energy))]])
    return evt
end
```

fhagemann: Do we have `eltype` and could we just do `T = eltype(pt_internal)`?

src/Event/Event.jl
```diff
@@ -19,30 +43,43 @@ mutable struct Event{T <: SSDFloat}
...
+    
+    if isnan(d)
+        evt.locations = VectorOfArrays(broadcast(pt -> [pt], pts_internal))
+        evt.energies  = VectorOfArrays(broadcast(E -> [T_(to_internal_units(E))], energies))
     else
-        evt.locations, evt.energies = group_points_by_distance(CartesianPoint.(locations), T.(to_internal_units.(energies)), d)
+        evt.locations, evt.energies = group_points_by_distance(pts_internal, T_.(to_internal_units.(energies)), d)
     end
     return evt
 end
```
fhagemann: Same here, you don't use `T`, so `T_` can become `T`:


```suggestion
function Event(locations::Vector{<:AbstractCoordinatePoint}, energies::Vector{<:RealQuantity} = fill(1.0, length(locations)); max_interaction_distance::Union{<:Real, <:LengthQuantity} = NaN)
    pts_internal = to_internal_point.(locations)
    T = typeof(first(pts_internal).x) # use eltype here if defined ?
    d::T = T(to_internal_units(max_interaction_distance))
    @assert isnan(d) || d >= 0

    evt = Event{T}()
    
    if isnan(d)
        evt.locations = VectorOfArrays(broadcast(pt -> [pt], pts_internal))
        evt.energies  = VectorOfArrays(broadcast(E -> [T(to_internal_units(E))], energies))
    else
        evt.locations, evt.energies = group_points_by_distance(pts_internal, T.(to_internal_units.(energies)), d)
    end
    return evt
end
```

src/Event/Event.jl
```diff
@@ -19,30 +43,43 @@ mutable struct Event{T <: SSDFloat}
...
+    locs_internal = [[to_internal_point(pt) for pt in group] for group in locations]
+    ens_internal  = [[to_internal_units(E) for E in group] for group in energies]
+    T_ = typeof(first(first(locs_internal)).x)
+    
+    locs_T = [[convert(CartesianPoint{T_}, pt) for pt in group] for group in locs_internal]
+    ens_T  = [[T_(E) for E in group] for group in ens_internal]
+
+    evt = Event{T_}()
+    evt.locations = VectorOfArrays(locs_T)
+    evt.energies  = VectorOfArrays(ens_T)
```
fhagemann: Rename `T_` to `T`

src/Event/Event.jl
```diff
@@ -103,10 +139,13 @@ function Event(
...
+    ) where {T <: RealQuantity, PT <: ParticleType}
+
+    locs_internal = [to_internal_point.(loc) for loc in locations]
+    
+    @assert eachindex(locs_internal) == eachindex(energies) == eachindex(radius)
+    events = map(i -> Event(locs_internal[i], energies[i], N; particle_type, number_of_shells, radius = radius[i]), eachindex(locations))
     
-    @assert eachindex(locations) == eachindex(energies) == eachindex(radius)
-    events = map(i -> Event(locations[i], energies[i], N; particle_type, number_of_shells, radius = radius[i]), eachindex(locations))
     Event(flatview.(getfield.(events, :locations)), flatview.(getfield.(events, :energies)))
```
fhagemann: `T` and `PT` are never used later

src/Event/Event.jl
```diff
@@ -386,4 +425,4 @@ function get_electron_and_hole_contribution(evt::Event{T}, sim::Simulation{T, S}
             hole_contribution = RDWaveform(range(zero(T) * u"ns", step = dt * u"ns", length = length(signal_h)), signal_h * calibration_factor))
 end
 
-export get_electron_and_hole_contribution
\ No newline at end of file
+export get_electron_and_hole_contribution
```
fhagemann, line 389: Can you avoid end-of-file changes like these? :sweat_smile:

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
fhagemann: And maybe, this can be a bit more specific than `Unitful.Quantity`?
e.g.
```suggestion
    r_val = isa(r, Unitful.Length) ? float(ustrip(internal_length_unit, r)) : float(r)
```

src/ChargeCloudModels/ChargeCloudModels.jl
```diff
@@ -121,6 +132,17 @@ NBodyChargeCloud(center, energy, 200, number_of_shells = 3)
...
+function NBodyChargeCloud(center::CartesianPoint{<:Unitful.Quantity}, energy::RealQuantity, N::Integer;
+                          kwargs...)
+    internal_center = to_internal_point(center)
+    return NBodyChargeCloud(internal_center, energy, N; kwargs...)
+end
+
+function NBodyChargeCloud(center::CylindricalPoint{T}, energy::RealQuantity, N::Integer; kwargs...) where T<:Union{SSDFloat, Unitful.Quantity}
+    internal_center = to_internal_point(center)
+    return NBodyChargeCloud(internal_center, energy, N; kwargs...)
+end
```
fhagemann: This can also be combined with the previous methods (without `N`) to one single expression: 
```suggestion
function NBodyChargeCloud(center::Union{<:CartesianPoint{<:Unitful.Quantity}, <:Union{SSDFloat, Unitful.Quantity}}, 
                          energy::RealQuantity, args...; kwargs...)
    internal_center = to_internal_point(center)
    return NBodyChargeCloud(internal_center, energy, args...; kwargs...)
end
```

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -244,6 +244,17 @@ struct CylindricalPoint{T} <: AbstractCylindricalPoint{T}
...
+# Handles Unitful quantities or plain numbers 
+function CylindricalPoint(r, φ, z)
+    r_val = isa(r, Unitful.Length) ? float(ustrip(internal_length_unit, r)) : float(r)
+    φ_val = isa(φ, Unitful.Quantity) ? float(ustrip(internal_angle_unit, φ)) : float(φ)
+    z_val = isa(z, Unitful.Length) ? float(ustrip(internal_length_unit, z)) : float(z)
+
+    T = promote_type(typeof(r_val), typeof(φ_val), typeof(z_val))
+    CylindricalPoint{T}(T(r_val), T(φ_val), T(z_val))
+
+end
```
fhagemann: This could probably also use either `to_internal_units` or `_parse_value`?

src/Event/Event.jl
```diff
@@ -1,6 +1,36 @@
...
+    return CartesianPoint{T}(x, y, z)
+end
+
+function to_internal_point(pt::CylindricalPoint)
+    T = pt.r isa Unitful.Quantity ? float(typeof(ustrip(pt.r))) : float(typeof(pt.r))
+    r = pt.r isa Unitful.Quantity ? float(to_internal_units(pt.r)) : float(pt.r)
+    φ = pt.φ isa Unitful.Quantity ? float(ustrip(uconvert(u"rad", pt.φ))) : float(pt.φ)
+    z = pt.z isa Unitful.Quantity ? float(to_internal_units(pt.z)) : float(pt.z)
+    return CartesianPoint{T}(r * cos(φ), r * sin(φ), z)
+end
```
fhagemann: I just looked a bit more into this:
Maybe this can be made more generic using already existing CSG methods like `_csg_get_promoted_eltype` and `promote_type`.

This is an example of how it's done for `Cone`:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a8518b65f08cfcf78c49091a1140d3e6995c1c97/src/ConstructiveSolidGeometry/VolumePrimitives/Cone.jl#L85-L98

src/Event/Event.jl
```diff
@@ -19,30 +49,44 @@ mutable struct Event{T <: SSDFloat}
...
+        evt.locations, evt.energies = group_points_by_distance(pts_internal, T.(to_internal_units.(energies)), d)
     end
     return evt
 end
 
-function Event(locations::Vector{<:Vector{<:AbstractCoordinatePoint{T}}}, energies::Vector{<:Vector{<:RealQuantity}} = [[one(T) for i in j] for j in locations])::Event{T} where {T <: SSDFloat}
+function Event(locations::Vector{<:Vector{<:AbstractCoordinatePoint}}, energies::Vector{<:Vector{<:RealQuantity}} = [[1.0 for _ in group] for group in locations])
+
+    locs_internal = [[to_internal_point(pt) for pt in group] for group in locations]
+    ens_internal  = [[to_internal_units(E) for E in group] for group in energies]
```
fhagemann: Seeing this, would it make sense to call `to_internal_point` also `to_internal_units`?

src/Event/Event.jl
```diff
@@ -19,30 +49,44 @@ mutable struct Event{T <: SSDFloat}
...
+    locs_T = [[convert(CartesianPoint{T}, pt) for pt in group] for group in locs_internal]
+    ens_T  = [[T(E) for E in group] for group in ens_internal]
+
     evt = Event{T}()
-    evt.locations = VectorOfArrays(broadcast(pts -> CartesianPoint{T}.(pts), locations))
-    evt.energies = VectorOfArrays(Vector{T}.(to_internal_units.(energies)))
+    evt.locations = VectorOfArrays(locs_T)
+    evt.energies  = VectorOfArrays(ens_T)
     return evt
 end
```
fhagemann: Could the determination of `T` be upfront to only create the nested array once instead of twice?

```suggestion
function Event(locations::Vector{<:Vector{<:AbstractCoordinatePoint}}, energies::Vector{<:Vector{<:RealQuantity}} = [[1.0 for _ in group] for group in locations])

    T = eltype(to_internal_point(first(first(locations)))
    locs_T = [[convert(CartesianPoint{T}, to_internal_point(pt)) for pt in group] for group in locations]
    ens_T  = [[T(to_internal_units(E)) for E in group] for group in energies]
    
    evt = Event{T}()
    evt.locations = VectorOfArrays(locs_T)
    evt.energies  = VectorOfArrays(ens_T)
    return evt
end
```

or maybe the old syntax even works (I didn't test this though)?


```suggestion
function Event(locations::Vector{<:Vector{<:AbstractCoordinatePoint}}, energies::Vector{<:Vector{<:RealQuantity}} = [[1.0 for _ in group] for group in locations])

    T = eltype(to_internal_point(first(first(locations)))
    locs = [[convert(CartesianPoint{T}, to_internal_point(pt)) for pt in group] for group in locations]
    ens  = [[T(to_internal_units(E)) for E in group] for group in energies]
    
    evt = Event{T}()
    evt.locations = VectorOfArrays(broadcast(pts -> CartesianPoint{T}.(to_internal_point.(pts)), locs))
    evt.energies  = VectorOfArrays(Vector{T}.(to_internal_units.(ens)))
    return evt
end
```

src/Event/Event.jl
```diff
@@ -19,30 +49,44 @@ mutable struct Event{T <: SSDFloat}
...
+    evt.energies  = VectorOfArrays([[T(to_internal_units(energy))]])
     return evt
 end
 
-function Event(locations::Vector{<:AbstractCoordinatePoint{T}}, energies::Vector{<:RealQuantity} = ones(T, length(locations)); max_interaction_distance::Union{<:Real, <:LengthQuantity} = NaN)::Event{T} where {T <: SSDFloat}
+function Event(locations::Vector{<:AbstractCoordinatePoint}, energies::Vector{<:RealQuantity} = fill(1.0, length(locations)); max_interaction_distance::Union{<:Real, <:LengthQuantity} = NaN)
+    pts_internal = to_internal_point.(locations)
+    T = eltype(first(pts_internal))
     d::T = T(to_internal_units(max_interaction_distance))
-    @assert isnan(d) || d >= 0 "Max. interaction distance must be positive or NaN (no grouping), but $(max_interaction_distance) was given."
```
fhagemann, line 31: Why did you remove the error message?

src/Event/Event.jl
```diff
@@ -19,30 +49,44 @@ mutable struct Event{T <: SSDFloat}
...
     evt = Event{T}()
-    evt.locations = VectorOfArrays([[CartesianPoint(location)]])
-    evt.energies = VectorOfArrays([[T(to_internal_units(energy))]])
+    evt.locations = VectorOfArrays([[pt_internal]])
+    evt.energies  = VectorOfArrays([[T(to_internal_units(energy))]])
     return evt
 end
 
-function Event(locations::Vector{<:AbstractCoordinatePoint{T}}, energies::Vector{<:RealQuantity} = ones(T, length(locations)); max_interaction_distance::Union{<:Real, <:LengthQuantity} = NaN)::Event{T} where {T <: SSDFloat}
+function Event(locations::Vector{<:AbstractCoordinatePoint}, energies::Vector{<:RealQuantity} = fill(1.0, length(locations)); max_interaction_distance::Union{<:Real, <:LengthQuantity} = NaN)
```
fhagemann: Maybe define the energies with `Ints` to avoid rounding errors if you already start with `Float64`:
```suggestion
function Event(locations::Vector{<:AbstractCoordinatePoint}, energies::Vector{<:RealQuantity} = fill(1, length(locations)); max_interaction_distance::Union{<:Real, <:LengthQuantity} = NaN)
```

src/Event/Event.jl
```diff
@@ -19,30 +49,44 @@ mutable struct Event{T <: SSDFloat}
...
+    pts_internal = to_internal_point.(locations)
+    T = eltype(first(pts_internal))
     d::T = T(to_internal_units(max_interaction_distance))
```
fhagemann: Side comment: I fear a bit that determining `T` here (and anywhere else) might create some trouble. If we require `T` of the `Event` to match `T` of the `Simulation` when drifting events, this might not be what we want..

```julia
simulate!(evt::Event{T}, sim::Simulation{T},..)
```
(the `T` need to match and I fear that this implementation might make it a bit unpredictable, especially because the function does not know the `T` of the output `Event{T}` before being called.

src/Event/Event.jl
```diff
@@ -1,6 +1,36 @@
...
+    z = pt.z isa Unitful.Quantity ? float(to_internal_units(pt.z)) : float(pt.z)
+    return CartesianPoint{T}(r * cos(φ), r * sin(φ), z)
+end
+
+function to_internal_point(pt::AbstractCoordinatePoint)
+    error("Unsupported point type $(typeof(pt)). Expected CartesianPoint or CylindricalPoint.")
+end
+
+function to_internal_point(p::SVector{3, <:Unitful.Quantity})
+    vals = ustrip.(uconvert.(u"m", p))   # convert to meters and strip units
```
fhagemann:
```suggestion
    vals = ustrip.(internal_length_unit, p))
```

src/Event/Event.jl
```diff
@@ -1,6 +1,36 @@
...
+    return SVector{3, typeof(vals[1])}(vals)  # preserves type
```
fhagemann: Do we need to explicitly say this or can this just be:

```suggestion
    return SVector{3}(vals)  # preserves type
```

test/test_charge_drift_models.jl
```diff
@@ -565,3 +565,48 @@ end
...
+    cartesian_unit = CartesianPoint(0.01u"m", 0.0u"m", 0.05u"m")
+    cartesian_float = CartesianPoint(0.01, 0.0, 0.05)
+    cylindrical_unit = CylindricalPoint(0.01u"m", π/4u"rad", 0.05u"m")
+    cylindrical_float = CylindricalPoint(0.01, π/4, 0.05)
+    Edep = 1u"MeV"
+    
+    # Variant 1: CartesianPoint with Units, energy
+    nb1 = NBodyChargeCloud(cartesian_unit, Edep)
+    @test all(x -> x isa CartesianPoint, nb1.locations)
+    @test isapprox(sum(nb1.energies), ustrip.(uconvert.(u"eV", Edep)))
```
fhagemann: Avoid using `ustrip.(uconvert.(unit, quantity)`, you can just do `ustrip.(unit, quantity)`

test/test_geant4.jl
```diff
@@ -66,11 +67,19 @@ end
     @test evts isa Table
     @test length(evts) == 100
 
+    # Temporary HDF5 file
+    mktemp() do tmpfile, io
+        LegendHDF5IO.lh5open(tmpfile, "w") do h5
+            LegendHDF5IO.writedata(h5, "geant4", evts)
+        end
+        @test isfile(tmpfile)
+    end
```
fhagemann, line 96: I would also be interested if this can be read back in and gives the same table:

```suggestion
    mktemp() do tmpfile, io
        LegendHDF5IO.lh5open(tmpfile, "w") do h5
            LegendHDF5IO.writedata(h5, "geant4", evts)
        end
        @test isfile(tmpfile)
        
        evts == LegendHDF5IO.lh5open(tmpfile) do h5
            LegendHDF5IO.readdata(h5, "geant4")
        end
    end
```

ext/SolidStateDetectorsGeant4Ext.jl
```diff
@@ -220,9 +220,12 @@ function SolidStateDetectors.run_geant4_simulation(app::G4JLApplication, number_
         evtno += 1
     end
 
-    evts_with_points = map(evt -> merge(evt, (pos = cartesian_zero + evt.pos,)), evts)
+    evts_with_points = map(evts) do evt
+        merge(evt, (pos = SolidStateDetectors.to_internal_point(evt.pos),))
+    end
```
fhagemann: The tests (Documentation) are failing because it expects `evts.pos` to have units of `m` that are stripped in the documentation here:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a8518b65f08cfcf78c49091a1140d3e6995c1c97/docs/src/tutorials/geant4_ssd_lit.jl#L92

fhagemann (dismissed):

ext/SolidStateDetectorsGeant4Ext.jl
```diff
@@ -220,9 +220,9 @@ function SolidStateDetectors.run_geant4_simulation(app::G4JLApplication, number_
         evtno += 1
     end
 
-    evts_with_points = map(evt -> merge(evt, (pos = cartesian_zero + evt.pos,)), evts)
-
-    return RadiationDetectorSignals.group_by_evtno(Table(evts_with_points))
+    #evts_with_points = map(evt -> merge(evt, (pos = cartesian_zero + evt.pos,)), evts) #This line triggers the error when using `writedata` function
+    
+    return RadiationDetectorSignals.group_by_evtno(Table(evts))
```
fhagemann: Just to have it written down somewhere: this will result in the column `pos` to be of eltype `SVector{3}` with units `u"mm"`

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
fhagemann: Do we need `float` before determining `T` in line 247 here?

I would just strip the units and pass this to `CylindricalPoint(r::Real, φ::Real, z::Real)` in line 252 then, to deal with the type promotion.
```suggestion
# Handles Unitful quantities or plain numbers 
function CylindricalPoint(r, φ, z)
    r_val::Real = to_internal_units(r)
    φ_val::Real = to_internal_units(φ)
    z_val::Real = to_internal_units(z)
    CylindricalPoint(r_val, φ_val, z_val)
end
```

Because right now I'm getting a different `T`, depending on whether the point has units or not, and I would argue that we do not want that:
Image

We could maybe also add a check that the arguments can only be lengths or angles.
```suggestion
# Handles Unitful quantities or plain numbers 
function CylindricalPoint(r::Union{<:Real, <:LengthQuantity}, φ::Union{<:Real, <:AngleQuantity}, z::Union{<:Real, <:LengthQuantity})
    r_val::Real = to_internal_units(r)
    φ_val::Real = to_internal_units(φ)
    z_val::Real = to_internal_units(z)
    CylindricalPoint(r_val, φ_val, z_val)
end
```

or

```suggestion
# Handles Unitful quantities or plain numbers 
function CylindricalPoint(r, φ, z)
    # Check for correct units
    (r isa Union{<:Real, <:LengthQuantity}) || throw(ArgumentError("Expected `r` to be a length, has unit of $(dimension(r))")
    (φ isa Union{<:Real, <:AngleQuantity}) || throw(ArgumentError("Expected `φ` to be an angle, has unit of $(dimension(φ))")
    (z isa Union{<:Real, <:LengthQuantity}) || throw(ArgumentError("Expected `z` to be a length, has unit of $(dimension(z))")
    
    # Strip units
    r_val::Real = to_internal_units(r)
    φ_val::Real = to_internal_units(φ)
    z_val::Real = to_internal_units(z)
    CylindricalPoint(r_val, φ_val, z_val)
end
```

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -303,4 +308,27 @@ Base.iszero(pt::CylindricalPoint) = iszero(pt.r) && iszero(pt.z)
...
+    return CartesianPoint(_r * cos(_φ), _r * sin(_φ), _z)
+end
+
+function to_internal_units(pt::AbstractCoordinatePoint)
+    error("Unsupported point type $(typeof(pt)). Expected CartesianPoint or CylindricalPoint.")
+end
+
+function to_internal_units(p::SVector{3, <:Unitful.Quantity})
+    vals = ustrip.(internal_length_unit, p)
+    return SVector{3, typeof(vals[1])}(vals)
```
fhagemann: Do you need `typeof(vals[1])` here?
```suggestion
    return SVector{3}(vals)
```

test/test_charge_drift_models.jl
```diff
@@ -565,3 +565,49 @@ end
         end
     end
 end
+
+@timed_testset "NBodyChargeCloud Units" begin
+
+    cartesian_unit = CartesianPoint(0.01u"m", 0.0u"m", 0.05u"m")
+    cartesian_float = CartesianPoint(0.01, 0.0, 0.05)
+    #cylindrical_unit = CylindricalPoint(0.01u"m", π/4u"rad", 0.05u"m")
```
fhagemann: Excluded because not supported?

fhagemann:
With my latest commit be4ec17, we should now be able to write/read `CartesianPoint` and `CylindricalPoint`, as well as Vectors containing those points using `LegendHDF5IO` -- both in the old `readdata`/`writedata` way and with the new `LH5Array`.

Tests failing are unrelated, same issue as before (with the documentation and Geant4 tests expecting `pos` to have eltype `CartesianPoint` instead of `SVector{3}`).

fhagemann:
I also reverted two of your commits to see if going back to `CartesianPoint` for the eltype in the Geant4 Table `pos` column fixes the tests again.

ext/SolidStateDetectorsGeant4Ext.jl
fhagemann: --> I reverted this change

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
fhagemann: Is this still used somewhere?

src/Event/Event.jl
claudiaalvgar: Yes, this is a worry but it's the only way I could think of if we want to pass units to Event()

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
claudiaalvgar: I changed the function ```CylindricalPoint(r, φ, z)``` and now I get this since ```ustrip``` always returns a float:
Screenshot 2025-11-14 at 10 32 37

The function ```to_internal_units(p::SVector{3, <:Unitful.Quantity})``` is not required anymore

test/test_charge_drift_models.jl
claudiaalvgar: I added this back

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -238,17 +238,34 @@ struct CylindricalPoint{T} <: AbstractCylindricalPoint{T}
...
-    r_val = isa(r, Unitful.Length) ? float(to_internal_units(r)) : float(r)
-    φ_val = isa(φ, Unitful.Quantity) ? float(to_internal_units(φ)) : float(φ)
-    z_val = isa(z, Unitful.Length) ? float(to_internal_units(z)) : float(z)
+    if !(r isa Real || r isa Unitful.Length)
+        throw(ArgumentError("Expected `r` to be a length or Real, got unit $(Unitful.unit(r))"))
+    end
 
-    T = promote_type(typeof(r_val), typeof(φ_val), typeof(z_val))
-    CylindricalPoint{T}(T(r_val), T(φ_val), T(z_val))
+    if !(φ isa Real || φ isa Unitful.Quantity)
```
fhagemann: Everything with units is a `Unitful.Quantity`, you might wanna check explicitly for something having no dimension:
```suggestion
    if !(φ isa Real || φ isa Unitful.Quantity{<:Real, NoDims})
```

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -238,17 +238,34 @@ struct CylindricalPoint{T} <: AbstractCylindricalPoint{T}
...
 
+    vals = (r_val, φ_val, z_val)
+    has_int   = any(v -> v isa Integer, vals)
+    has_float = any(v -> v isa AbstractFloat, vals)
+
+    if has_int && has_float
+        throw(ArgumentError("Mixed integer and float inputs, use all integers or all floats."))
+    end
+    
+    return CylindricalPoint(r_val, φ_val, z_val)
```
fhagemann: Do we really need this? If `r` is an `Int` `r_val` will be a `Rational{Int}`:
```julia
using Unitful
SolidStateDetectors.to_internal_units(4u"mm")
```
```
1//250
```

Can this just be
```suggestion
    return CylindricalPoint(r_val, φ_val, z_val)
```
and the rest is handled by the function in lines 269ff?

The `promote_type` method will make sure to use the `Float` types (if there is some), and only use `Float64` if only `Int` or `Rational` are passed as input.

With the current (but also with the old) code, I get this:
image

So, as long as we don't convert `r_val` etc. to `Float` in this method, we should be fine and the function should give us what we want.

fhagemann: Can you add the expected type conversion to the tests (I'm gonna shamelessly copy your picture):
image

Something like this:
```julia
@test CylindricalPoint(1u"m", 2u"rad", 1u"m") isa CylindricalPoint{Float64}
@test_throws ArgumentError CylindricalPoint(1u"m", 2u"m", 3u"m")
```

claudiaalvgar:
Yes, in principle, before this PR ```CylindricalPoint``` with units wasn't defined. I just added it this way in this PR because sometimes, it could be more convenient to pass the cylindrical coordinate to the ```Event``` if your detector is in those coordinates but we can talk about the best way of doing this. Are there any other changes to add?

test/ConstructiveSolidGeometry/CSG_coordinates.jl
```diff
@@ -36,21 +36,42 @@ import Unitful
...
+        @test CartesianPoint(1.0u"mm", 2.0u"cm", 3.0f0u"m") isa CartesianPoint{Float64}
+        @test CartesianPoint(1.0u"mm", 2.0f0u"cm", 3.0f0u"m") isa CartesianPoint{Float64}
+        @test CartesianPoint(1.0f0u"mm", 2.0f0u"cm", 3.0f0u"m") isa CartesianPoint{Float32}
+        @test CartesianPoint(1u"mm", 2u"cm", 3u"m") isa CartesianPoint{Float64}
+        
+        # test throwing errors with wrong units
+        @test_throws ArgumentError CartesianPoint(1u"m", 2u"rad", 3u"m")
+        @test_throws ArgumentError CartesianPoint(1u"s", 2u"m", 3u"m")
+        @test_throws ArgumentError CartesianPoint(1u"m", 2u"m", 3u"kg")
+        =#
```
claudiaalvgar: Why are the ```CartesianPoint``` tests commented out here?

fhagemann: Because they don’t work right now (`CartesianPoint` can have units, and in these tests, they would have the units stripped)

fhagemann:
> Didn't we want to leave units out for now?

That’s the question:
I would say, we should for now also strip the units from `CartesianPoint`, but still allow for the user to create them using units. The coordinates internally should still be floats for now, and let’s add support for units on the internal coordinates themselves in a future PR.

fhagemann:
@claudiaalvgar: could you remove the units from the `CartesianPoint` coordinates everywhere, such that `pt.x`, `pt.y` and `pt.z` are always unitless numbers in units of meters ?
Then, you should be able to also run the commented new tests for `CartesianPoint`.

fhagemann:
> Yes, I can add this but I still have the question if the code in Event.jl will stay as it is, 

I don't see why not. Many (all?) methods use `to_internal_units` which should work perfectly with unitless numbers, as long as those numbers are in meters.

> if we should delete the changes to account for the units or 

Ideally, all changes should still work, if we use values without units. I would keep the changes, but maybe restrict the type `T` of `AbstractCoordinatePoint{T}` to `AbstractCoordinatePoint{T <: Real}` or `AbstractCoordinatePoint{T <: SSDFloat}`.
We can then check, if the rest still works or not.

> if there are more changes to implement given this comment: 

If the event locations are of type `T` now, they should give the correct `T` for `Event{T}`.
We should add tests to ensure this though.

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -198,7 +214,8 @@ end
 @inline Base.zero(::CartesianZero{T}) where {T} = CartesianZero{T}()
 @inline Base.iszero(::CartesianZero) = true
 
-Base.:(*)(::CartesianZero{T}, u::Unitful.Units{<:Any,Unitful.𝐋}) where {T<:Real} = CartesianZero{Quantity{T, Unitful.𝐋, typeof(u)}}()
+#Base.:(*)(::CartesianZero{T}, u::Unitful.Units{<:Any,Unitful.𝐋}) where {T<:Real} = CartesianZero{Quantity{T, Unitful.𝐋, typeof(u)}}()
+Base.:*(z::CartesianZero, u::Unitful.Quantity) = z
```
fhagemann, line 218: Maybe add a `TODO` here to revert this once we also support units internally.

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -66,6 +66,22 @@ function CartesianPoint(x::TX, y::TY, z::TZ) where {TX<:Real,TY<:Real,TZ<:Real}
...
+    end
+    CartesianPoint(x::Unitful.Length, y::Unitful.Length, z::Unitful.Length)
+end
+
+function CartesianPoint(x::Unitful.Length, y::Unitful.Length, z::Unitful.Length)
+    vals = to_internal_units.( (x, y, z) )
+    T = all(v -> v isa Float32, vals) ? Float32 : Float64
+    CartesianPoint{T}(T.(vals)...)
+end
+
```
fhagemann: This right now doesn't cover the case `CartesianPoint(1u"mm", 0, 0)` (if only some of the arguments have units.
Consider making this equivalent to what we did for `CylindricalPoint`:
```suggestion
function CartesianPoint(x, y, z)
    for (name, pt) in zip((:x,:y,:z), (x, y, z))
        (pt isa Real || pt isa Unitful.Length) || throw(ArgumentError(
            "Expected `$(name)` to be a length or Real, got unit $(Unitful.unit(pt))"
        ))
    end
    x_val = to_internal_units(x)
    y_val = to_internal_units(y)
    z_val = to_internal_units(z)
    return CartesianPoint(x_val, y_val, z_val)
end
```

and let the `function CartesianPoint(x::TX, y::TY, z::TZ) where {TX<:Real,TY<:Real,TZ<:Real}` method deal with the type promotion.

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
fhagemann, line 1: I would still restrict `CartesianPoint{T}`
```julia
struct CartesianPoint{T <: AbstractFloat} <: AbstractCartesianPoint{T}
```
or
```julia
struct CartesianPoint{T <: Real} <: AbstractCartesianPoint{T}
```

(I think `SSDFloat` is not defined in `ConstructiveSolidGeometry`, though..)

The issue with `T <: Real` is that this would still allow for `CartesianPoint{Int}`..

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -66,6 +66,22 @@ function CartesianPoint(x::TX, y::TY, z::TZ) where {TX<:Real,TY<:Real,TZ<:Real}
...
+        ))
+    end
+    CartesianPoint(x::Unitful.Length, y::Unitful.Length, z::Unitful.Length)
+end
+
+function CartesianPoint(x::Unitful.Length, y::Unitful.Length, z::Unitful.Length)
+    vals = to_internal_units.( (x, y, z) )
+    T = all(v -> v isa Float32, vals) ? Float32 : Float64
+    CartesianPoint{T}(T.(vals)...)
+end
```
fhagemann: With this implementation, I'm getting unexpected type results:
```julia
CartesianPoint(1u"m", 2u"m", 3f0u"m") # CartesianPoint{Float64}(1.0, 2.0, 3.0)
```
is of type `Float64`, even though I only passed `Ints` and `Float32`.

We should add this case to the tests.

test/test_io.jl
```diff
@@ -69,7 +69,7 @@ end
         detno = Int32[1],
         thit = VectorOfVectors([T[0] * u"s"]),
         edep = VectorOfVectors([T[1] * u"eV"]),
-        pos = VectorOfVectors([[SVector{3, T}.(0.01, 0.01, 0.01) * u"m"]])
```
fhagemann, line 72: Why? The method with `SVectors` and units should still work.
Then, we just need to call `to_internal_units` or `_parse_value` somewhere, where we're calling `ustrip` now.
This is also what's making the Documentation build fail.

src/MCEventsProcessing/MCEventsProcessing.jl
```diff
@@ -63,6 +63,13 @@ simulate_waveforms(mcevents, sim, "output_dir", "my_basename", Δt = 1u"ns", ver
 !!! note
     Using values with units for `Δt` requires the package [Unitful.jl](https://github.com/JuliaPhysics/Unitful.jl).
 """
+function _get_unitless_positions_svec(point_vectors::AbstractVector{<:SVector{3, <:Quantity}})
+    converted_type = typeof(to_internal_units(point_vectors[1][1]))
+    [SVector{3, converted_type}(to_internal_units.(v)) for v in point_vectors]
+end
+_get_unitless_positions_svec(point_vectors::AbstractVector{<:SVector{3, <:Real}}) = point_vectors
+_get_unitless_positions_svec(point_vectors::AbstractVector{<:CartesianPoint{<:Real}}) = point_vectors
```
fhagemann: Doesn't this work with `to_internal_units`? 
```julia
_get_unitless_positions_svec(point_vectors) = to_internal_units.(point_vectors)
```
or no need to even define a function

src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -58,30 +58,37 @@ struct CartesianPoint{T} <: AbstractCartesianPoint{T}
...
 
 function CartesianPoint(x::Unitful.Length, y::Unitful.Length, z::Unitful.Length)
-    vals = to_internal_units.( (x, y, z) )
-    T = all(v -> v isa Float32, vals) ? Float32 : Float64
+    vals = to_internal_units.((x, y, z))
+    tys = map(typeof, vals)
+    T = any(T -> T === Float64, tys) ? Float64 :
+        any(T -> T === Float32, tys) ? Float32 :
+        Float64
+
```
fhagemann: This could be avoided if `CartesianPoint` in the struct were to be restriced to `AbstractFloat`
 or by defining a constructor in the struct (see `CylindricalPoint`):
```julia
struct CartesianPoint{T} <: AbstractCartesianPoint{T}
    x::T
    y::T
    z::T
    CartesianPoint{T}(x::Real, y::Real, z::Real) where {T} = new(T(x), T(y), T(z))
end
```
where then there should be no default `CartesianPoint(...)` method (without `{T}`) and it should then use the fallback method for the type promotion defined in line 84.

src/MCEventsProcessing/MCEventsProcessing.jl
```diff
@@ -76,6 +76,10 @@ function simulate_waveforms( mcevents_table::AbstractVector{<:NamedTuple}, sim::
                              geometry_check::Bool = false,
                              verbose::Bool = false ) where {T <: SSDFloat}
     mcevents = TypedTables.Table(mcevents_table)
+    #ToDo: Provisional until we support units in CartesianPoints
+    unitless_pos = [collect(to_internal_units.(v)) for v in mcevents.pos]
+    mcevents = TypedTables.Table((evtno = mcevents.evtno, detno = mcevents.detno, thit  = mcevents.thit, edep  = mcevents.edep, pos   = unitless_pos))
+
```
fhagemann: I'm not a fan of this, as it changes the `pos` column of the `Table` as output of this function. I can also see that we have a `_get_unitless_position` function beneath this function, can we not use this WITHIN this function, but still return the OLD table with the new column with waveforms at the end of this function?

src/ChargeClustering/ChargeClustering.jl
```diff
@@ -16,8 +16,7 @@ function cluster_detector_hits(
     r_edep = similar(edep, 0)
     r_pos = similar(pos, 0)
 
-    posunit = unit(PT)
-    ustripped_cradius = ustrip(posunit, cluster_radius)
+    ustripped_cradius = to_internal_units(cluster_radius)
```
fhagemann: We should either force `cluster_radius` to be `<:Union{Real, Unitful.Length}`:
```julia
function cluster_detector_hits(
        detno::AbstractVector{<:Integer},
        edep::AbstractVector{TT},
        pos::AbstractVector{<:Union{<:StaticVector{3,PT}, <:CartesianPoint{PT}}},
        cluster_radius::Union{<:Real, Unitful.Length} # <-- changed this
    ) where {TT <: RealQuantity, PT <: RealQuantity}

    unsorted = TypedTables.Table(detno = detno, edep = edep, pos = pos)
    sorting_idxs = sortperm(unsorted.detno)
    sorted = unsorted[sorting_idxs]
    grouped = TypedTables.Table(consgroupedview(sorted.detno, TypedTables.columns(sorted)))

    r_detno = similar(detno, 0)
    r_edep = similar(edep, 0)
    r_pos = similar(pos, 0)

    ustripped_cradius = to_internal_units(cluster_radius)
```
or explicitly write this to expect a length:

```julia
        function cluster_detector_hits(
        detno::AbstractVector{<:Integer},
        edep::AbstractVector{TT},
        pos::AbstractVector{<:Union{<:StaticVector{3,PT}, <:CartesianPoint{PT}}},
        cluster_radius::RealQuantity
    ) where {TT <: RealQuantity, T <: Real, PT <: RealQuantity{T}} # <-- changed this
    
    unsorted = TypedTables.Table(detno = detno, edep = edep, pos = pos)
    sorting_idxs = sortperm(unsorted.detno)
    sorted = unsorted[sorting_idxs]
    grouped = TypedTables.Table(consgroupedview(sorted.detno, TypedTables.columns(sorted)))
    
    r_detno = similar(detno, 0)
    r_edep = similar(edep, 0)
    r_pos = similar(pos, 0)

    ustripped_cradius = _parse_value(T, cluster_radius, internal_length_unit) # <-- and this
```

src/MCEventsProcessing/MCEventsProcessing.jl
claudiaalvgar: I already tried using the function ```_get_unitless_position``` before implementing my change but everything was failing, that's why I added the function ```_get_unitless_positions_svec``` since this was giving the right output... This change is the one that you suggested in the previous comments but if you have a suggestion on how to use the function ```_get_unitless_position``` and get the right output I can change it

fhagemann: I will have a look into this, don't worry and enjoy the weekend :)

src/ChargeClustering/ChargeClustering.jl
fhagemann: I chose the `_parse_value` implementation, see cfba8d4

src/MCEventsProcessing/MCEventsProcessing.jl
fhagemann: This is now moved to a function further down --> f6261ca

src/MCEventsProcessing/MCEventsProcessing.jl
fhagemann: Obsolete with f6261ca now

PR #562, opened 2025-11-14
Adding temperature dependence to drift model parameters

hervasa2:
I have now completely deprecated all old TemperatureModels, and removed the `scale_to_given_temperature` functionality, which would scale the longitudinal drift at each step in the charge drift. 

In its place with @fhagemann we propose to use,
```julia
scale_to_temperature(cdm::ADL2016ChargeDriftModel{T,M,N,PowerLawModel{T}}, Temp::RealQuantity)
```
which directly scales the drift parameters according to temperature.

The parametrization for the temperature dependence of drift velocity, as a function of the electric 
field strength, $E$, was proposed by [M.A. Omar and L. Reggiani](https://www.sciencedirect.com/science/article/pii/0038110187900633):
```math
\quad\mu_0(T) = A/T^P~, \quad V_s(T) = B\tanh^{1/2}(\theta/2T) 
```
Note that the saturation velocity, $V_s$, is related to $E_0$ via
```math
E_0(T) = V_s(T)/\mu_0(T)
```
The four parameters, $A$, $P$, $B$, $\theta$, are different for electrons and holes. 
This model can be used to scale the $\mu_0$ and $E_0$ parameters of the [`VelocityParameters`](@ref) struct directly (assuming these where measured at a reference temperature of 77K) as follows:
```math
\mu_0(T) = \mu_0(77K) \left(\frac{T}{77K}\right)^{-P}~, \quad E_0(T) = E_0(77K) \sqrt{\frac{\tanh(\theta/2T)}{\tanh(\theta/2 \cdot 77K)}} \frac{\mu_0(77K)}{\mu_0(T)}
```
This model is now implemented under the new `PowerLawModel`.

In [siggen](https://github.com/radforddc/icpc_siggen/blob/7afd11a567958c7bb7c1737247a352cfa9263eaf/fields.c#L413)  the M.A. Omar and L. Reggiani model is used to scale the longitudinal drift velocity, as was implemented in this now closed PR #361. The scaling as calculated, assumes that the velocities to be scaled are modeled with the `SquareRootModel` ($\beta = 2$ drift model). Nevertheless, this can lead to distortions in our case (and in siggen), since $\beta \ne 2$ in the base drift model which we are scaling (`ADL2016ChargeDriftModel`) . 

The question now is: do we use the original parameters from M.A. Omar and L. Reggiani (since these where calculated assuming $\beta = 2$) or something else? I see in @fhagemann's thesis that the original parameters (listed below) give a good agreement with his measurements at $E = 500$ V/cm. Is this true at other electric field strengths?
```yaml
temperature_dependence:
  reference_temperature: 77K
  model: PowerLaw
  parameters:
    e100:
      p: 1.68
      theta: 200K
    h100:
      p: 2.4
      theta: 200K
```

One last detail, note that setting the `temperature` in the drift config is no longer supported with this PR. Since this is set in the main config file of the simulation, having it in two places is confusing for the user. Under the implementation of this PR it has to be manually passed to the drift model as such
```julia
ADL2016ChargeDriftModel(temperature = sim.detector.semiconductor.temperature)
```
In general I don't like any variables being set in drift model configs, which is why I also support `phi110` being moved to the simulation config, and also being manually passed as
```julia
ADL2016ChargeDriftModel(temperature = sim.detector.semiconductor.temperature, phi110 = sim.detector.semiconductor.phi110)
```
or in some clever way. 

Below I compare the longitudinal drift velocity as calculated by #361 (siggen-like -> scaled $v_l$) and as calculated by this PR (-> scaled drift model).  The residuals correspond to these two quantities. $v_l$ at 77 K is shown for reference.

vl_animation

vl_animation_e

### DIFF
--- a/src/ChargeDriftModels/IsotropicChargeDriftModel/IsotropicChargeDriftModel.jl
+++ b/src/ChargeDriftModels/IsotropicChargeDriftModel/IsotropicChargeDriftModel.jl
@@ -13,6 +13,8 @@ struct IsotropicChargeDriftModel{T <: SSDFloat} <: AbstractChargeDriftModel{T}
     μ_h::T
 end
 
+IsotropicChargeDriftModel{T}(μ_e::Union{Quantity, AbstractString}, μ_h::Union{Quantity, AbstractString}) where {T <: SSDFloat} = IsotropicChargeDriftModel{T}(_parse_value.(T, (μ_e, μ_h), internal_mobility_unit)...)
+
 function _IsotropicChargeDriftModel(
         config::AbstractDict; 
         T::Type=Float32,
@@ -21,10 +23,10 @@ function _IsotropicChargeDriftModel(
         input_units::Union{Missing, NamedTuple} = missing,
         temperature::Union{Missing, Real, Unitful.Temperature} = missing
     )::IsotropicChargeDriftModel{T}
+
     mobility_unit = ismissing(input_units) ? internal_mobility_unit : input_units.length^2/(input_units.potential*internal_time_unit)
    IsotropicChargeDriftModel{T}(_parse_value.(T, (μ_e, μ_h), mobility_unit)...)
 end
-
 
 # Check the syntax of the IsotropicChargeDriftModel config file before parsing
 function IsotropicChargeDriftModel(config::AbstractDict, input_units::Union{Missing, NamedTuple} = missing; kwargs...)


src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawModel.jl
```diff
@@ -1,57 +1,88 @@
...
-    println("Fit function: p1 * T^(3/2)\n")
-    println("---Temperature settings---")
-    println("Crystal temperature:   \t $(tm.temperature)")
-    println("Reference temperature: \t $(tm.reftemperature)\n")
-
-    println("---Fitting parameters---")
-    println("   \te100      \te111      \th100      \th111")
-    println("p1 \t$(tm.p1e100)   \t$(tm.p1e111)   \t$(tm.p1h100)   \t$(tm.p1h111)")
-end
+function scale_to_temperature(cdm::ADL2016ChargeDriftModel{T,M,N,PowerLawModel{T}}, Temp::RealQuantity) where {T,M,N}
```
fhagemann: Can we add the same temperature dependence for the standard `ADLChargeDriftModel`, and maybe even the other ones (`ConstantMobilityChargeDriftModel` could just be scaled with the power law, and the saturation drift velocity T-dependence is ignored).

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawModel.jl
```diff
@@ -1,57 +1,88 @@
...
+* `theta_h100::T`: Characteristic temperature $\theta$ for holes along the <100> axis.
+
+"""
+
+mutable struct PowerLawModel{T <: SSDFloat} <: AbstractTemperatureModel{T}
+    reference_temperature::T
+    p_e100::T
+    theta_e100::T
+    p_h100::T
+    theta_h100::T
```
fhagemann: Remove `100`, as we apply this to all axes?

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawModel.jl
```diff
@@ -1,57 +1,88 @@
...
 end
 
 function PowerLawModel(config_file::AbstractDict; T::Type{<:AbstractFloat} = Float32)::PowerLawModel
     return PowerLawModel{T}(config_file::AbstractDict)
 end
 
-function PowerLawModel{T}(config_file::AbstractDict; temperature::Union{Missing, T} = missing)::PowerLawModel where {T <: SSDFloat}
+function PowerLawModel{T}(config_file::AbstractDict)::PowerLawModel where {T <: SSDFloat}
     config_file["temperature_dependence"]["model"] != "PowerLaw" ? error() : nothing
```
fhagemann: This might need some improved handling in checking whether all of these fields exist, and a sophisticated `ConfigFileError` with a message, rather than just `error`.

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawModel.jl
```diff
@@ -1,57 +1,88 @@
...
+    p_h100::T
+    theta_h100::T
 end
 
 function PowerLawModel(config_file::AbstractDict; T::Type{<:AbstractFloat} = Float32)::PowerLawModel
     return PowerLawModel{T}(config_file::AbstractDict)
 end
 
-function PowerLawModel{T}(config_file::AbstractDict; temperature::Union{Missing, T} = missing)::PowerLawModel where {T <: SSDFloat}
+function PowerLawModel{T}(config_file::AbstractDict)::PowerLawModel where {T <: SSDFloat}
```
fhagemann: Should we add another constructor with keyword arguments and some default values?
Such that e.g. `PowerLawModel{T}()` gives some default, just like `ADL2016ChargeDriftModel{T}()` for example.

src/ChargeDriftModels/ADLChargeDriftModel/ADL2016ChargeDriftModel.jl
```diff
@@ -108,32 +108,31 @@ function _ADL2016ChargeDriftModel(
...
-            elseif model == "PowerLaw"
-                temperaturemodel = PowerLawModel{T}(config, temperature = temperature)
-            elseif model == "Boltzmann"
-                temperaturemodel = BoltzmannModel{T}(config, temperature = temperature)
+            if model == "PowerLaw"
+                temperaturemodel = PowerLawModel{T}(config)
             else
                 temperaturemodel = VacuumModel{T}(config)
-                println("Config File does not suit any of the predefined temperature models. The drift velocity will not be rescaled.")
+                println("Config File does not suit any of the predefined temperature models. The drift parameters will not be rescaled.")
```
fhagemann: Make this `@info` or `@warn`.
Maybe also make this more of a deprecation warning, for people who might still be using "old" config files. 

Check here how we used to phrase old deprecation warnings:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/554

src/ChargeDriftModels/ADLChargeDriftModel/ADL2016ChargeDriftModel.jl
```diff
@@ -108,32 +108,31 @@ function _ADL2016ChargeDriftModel(
...
         if "model" in keys(config["temperature_dependence"])
             model::String = config["temperature_dependence"]["model"]
-            if model == "Linear"
-                temperaturemodel = LinearModel{T}(config, temperature = temperature)
-            elseif model == "PowerLaw"
-                temperaturemodel = PowerLawModel{T}(config, temperature = temperature)
-            elseif model == "Boltzmann"
-                temperaturemodel = BoltzmannModel{T}(config, temperature = temperature)
+            if model == "PowerLaw"
+                temperaturemodel = PowerLawModel{T}(config)
```
fhagemann: As we are redefining `PowerLaw`, we might need some deprecation warning if someone passes an "old" power-law temperature dependence with the old syntax.

src/ChargeDriftModels/ADLChargeDriftModel/ADL2016ChargeDriftModel.jl
```diff
@@ -41,8 +41,8 @@ function _ADL2016ChargeDriftModel(
         h111μ0::Union{RealQuantity,String} = config["drift"]["velocity"]["parameters"]["h111"]["mu0"], 
         h111β::Union{RealQuantity,String}  = config["drift"]["velocity"]["parameters"]["h111"]["beta"], 
         h111E0::Union{RealQuantity,String} = config["drift"]["velocity"]["parameters"]["h111"]["E0"],
-        material::Type{<:AbstractDriftMaterial} = HPGe,
-        temperature::Union{Missing, Real} = missing, 
+        material::Type{<:AbstractDriftMaterial} = HPGe, 
+        temperature::Union{Missing, RealQuantity} = missing,
```
fhagemann:
```suggestion
        temperature::Union{Missing, Real, Unitful.Temperature} = missing, 
```

src/ChargeDriftModels/ADLChargeDriftModel/ADLChargeDriftModel.jl
fhagemann, line 1: This is just removing the temperature dependence from `ADLChargeDriftModel` completely without allowing for the new method? Can you implement `scale_to_temperature` for `ADLChargeDriftModel` as well?

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawModel.jl
```diff
@@ -1,57 +1,88 @@
...
+    Eh::T = cdm.holes.axis100.E0 * sqrt(tanh(theta_h / (2*Temp)) / tanh(theta_h / (2 * reftemp))) * cdm.holes.axis100.mu0 / μh
+    h100 = SolidStateDetectors.VelocityParameters{T}(μh, βh, Eh, zero(T))
+    μh = cdm.holes.axis111.mu0 * (Temp / reftemp) ^ -p_h
+    βh = cdm.holes.axis111.beta
+    Eh = cdm.holes.axis111.E0 * sqrt(tanh(theta_h / (2*Temp)) / tanh(theta_h / (2 * reftemp))) * cdm.holes.axis111.mu0 / μh
+    h111 = SolidStateDetectors.VelocityParameters{T}(μh, βh, Eh, zero(T))
+    holes = SolidStateDetectors.CarrierParameters(h100, h111)
+    
+    ADL2016ChargeDriftModel{T,M,N,PowerLawModel{T}}(electrons, holes, cdm.crystal_orientation, cdm.γ, cdm.parameters, cdm.temperaturemodel) 
+end
```
fhagemann: This has some `SolidStateDetectors` in it that could be removed.

fhagemann: Also, could the temperature scaling on `VelocityParameters` be exported to a separate function that's being called here then:

```suggestion
# scale velocity parameters given PowerLawModel parameters
function scale_to_temperature_powerlaw(v::VelocityParameters{T}, Temp::T, reftemp::T, p::T, theta::T)
    μ0::T = v.mu0 * (Temp / reftemp) ^ -p
    β::T  = v.beta
    E0::T = v.E0 * sqrt(tanh(theta / (2*Temp)) / tanh(theta/ (2 * reftemp))) * v.mu0 / μe
    μn::T = μe / v.mu0 * v.mun
    return VelocityParameters{T}(μ0, β, E0, μn)
end

function scale_to_temperature(cdm::CDM, Temp::Union{<:Real,Unitful.Temperature}) where {T,M,N,CDM<:ADL2016ChargeDriftModel{T,M,N,PowerLawModel{T}}}
    Temp = _parse_value(T, Temp, u"K")
    reftemp = cdm.temperaturemodel.reference_temperature
    theta_e = cdm.temperaturemodel.theta_e100
    theta_h = cdm.temperaturemodel.theta_h100
    p_e = cdm.temperaturemodel.p_e100
    p_h = cdm.temperaturemodel.p_h100

    # scale electron parameters
    electrons = scale_to_temperature_powerlaw(cdm.electrons, Temp, reftemp, p_e, theta_e)

    # scale hole parameters (take <100> values and assume no anisotropy for now)
    h100 = scale_to_temperature_powerlaw(cdm.holes.axis100, Temp, reftemp, p_h, theta_h)
    h111 = scale_to_temperature_powerlaw(cdm.holes.axis111, Temp, reftemp, p_h, theta_h)
    holes = CarrierParameters(h100, h111)

    CDM(electrons, holes, cdm.crystal_orientation, cdm.γ, cdm.parameters, cdm.temperaturemodel) 
end  
```

test/test_charge_drift_models.jl
```diff
@@ -525,6 +525,18 @@ end
...
+    @testset "Test equivalence of longitudinal drift parameter implementation" begin
+    
+        # If a temperature is given to the ADL2016 model, the ChargeDriftModel is scaled. But at 77K (the reference temperature),
+        # the parameters should be identical to the unscaled ADL2016 model
+        
+        cdm_scaled = ADL2016ChargeDriftModel(temperature = 77u"K")
+        cdm = ADL2016ChargeDriftModel()
+        
+        @test cdm_scaled.electrons == cdm.electrons
+        @test cdm_scaled.holes     == cdm.holes
```
fhagemann: Maybe also add tests that at higher temperatures we expect lower values for `µ0` etc. ?

examples/example_config_files/ADLChargeDriftModel/drift_velocity_config.yaml
```diff
@@ -22,4 +22,14 @@ drift:
       h111:
         mu0: 61215cm^2/(V*s)
         beta: 0.662
-        E0: 182V/cm
\ No newline at end of file
+        E0: 182V/cm
+temperature_dependence:
+  reference_temperature: 77K
```
fhagemann: The initial publication from 2006 quotes 78K as reference temperature (see Tab.1):
https://doi.org/10.1016/j.nima.2006.08.130

examples/example_config_files/ADLChargeDriftModel/drift_velocity_config_powerlaw.yaml
```diff
@@ -1,37 +0,0 @@
...
-      h100:
-        mu0: 6.1824
-        beta: 0.942
-        E0: 18500
-      h111:
-        mu0: 6.1215
-        beta: 0.662
-        E0: 18200
-temperature_dependence:
-  model: PowerLaw
```
fhagemann, line 27: I'm worried that we're using the same name `"PowerLaw"` in the config files for the new model.

We could use `Omar1987`, in accordance to `Bruyneel2016` or `Bruyneel2006`.

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawTemperatureModel.jl
```diff
@@ -0,0 +1,136 @@
...
+    # scale electron parameters (take <100> values and assume no anisotropy)
+    e100 = scale_to_temperature_powerlaw(cdm.electrons.axis100, Temp, reftemp, p_e, theta_e)
+    e111 = scale_to_temperature_powerlaw(cdm.electrons.axis111, Temp, reftemp, p_e, theta_e)
+    electrons = CarrierParameters(e100, e111)
+    # scale hole parameters (take <100> values and assume no anisotropy)
+    h100 = scale_to_temperature_powerlaw(cdm.holes.axis100, Temp, reftemp, p_h, theta_h)
+    h111 = scale_to_temperature_powerlaw(cdm.holes.axis111, Temp, reftemp, p_h, theta_h)
+    holes = CarrierParameters(h100, h111)
+    CDM(electrons, holes, cdm.crystal_orientation, cdm.γ, cdm.parameters, cdm.temperaturemodel)
+end
```
fhagemann: One thing to note: people shouldn't call `scale_to_temperature` twice, because that will apply this thing twice.

One (ugly) way of avoiding this would be to explicitly remove the temperature model after having scaled the values:
```julia
CDM(electrons, holes, cdm.crystal_orientation, cdm.γ, cdm.parameters, VacuumTemperatureModel{T})
```

src/ChargeDriftModels/ADLChargeDriftModel/ADL2016ChargeDriftModel.jl
```diff
@@ -108,32 +108,31 @@ function _ADL2016ChargeDriftModel(
...
-        # println("No temperature dependence found in Config File. The drift velocity will not be rescaled.")
+        temperaturemodel = VacuumTemperatureModel{T}(config)
+    end
+    cdm = ADL2016ChargeDriftModel{T,material,length(γ),typeof(temperaturemodel)}(electrons, holes, crystal_orientation, γ, parameters, temperaturemodel)
+
+    if ismissing(temperature)
+        cdm
+    else
+        scale_to_temperature(cdm, temperature)
     end
```
fhagemann: This `if`-`else` could be avoided by defining an additional method
```julia
scale_to_temperature(cdm::AbstractChargeDriftModel, ::Missing) = cdm
```

test/test_charge_drift_models.jl
```diff
@@ -271,7 +271,7 @@ end
         @test cdm0.μ_e  == 0.1f0
         @test cdm0.μ_h == 0.1f0
 
-        cdm1 = IsotropicChargeDriftModel{T}(1000u"cm^2/(V*s)", 1000u"cm^2/(V*s)")
+        cdm1 = IsotropicChargeDriftModel{T}(μ_e = 1000u"cm^2/(V*s)", μ_h = 1000u"cm^2/(V*s)")
```
fhagemann, line 274: I would like to also allow for the old syntax.
Can this be achieved using something like
```julia
IsotropicChargeDriftModel(μ_e::Union{RealQuantity, AbstractString}, μ_h::Union{RealQuantity, AbstractString}, args...; kwargs...) = IsotropicChargeDriftModel(args...; µ_e, µ_h, kwargs...)
```
?

src/ChargeDriftModels/ADLChargeDriftModel/ADL2006ChargeDriftModel.jl
```diff
@@ -167,23 +119,55 @@ function _ADLChargeDriftModel(
...
-            elseif model == "Boltzmann"
-                temperaturemodel = BoltzmannModel{T}(config, temperature = temperature)
+            if model == "PowerLaw"
+                temperaturemodel = PowerLawTemperatureModel{T}(config)
             else
-                temperaturemodel = VacuumModel{T}(config)
-                println("Config File does not suit any of the predefined temperature models. The drift velocity will not be rescaled.")
+                temperaturemodel = VacuumTemperatureModel{T}(config)
+                @info "Config File does not suit any of the predefined temperature models. The drift parameters will not be rescaled."
             end
```
fhagemann: Add a more descriptive deprecation warning if people should use something that was defined before (`since 0.11, the temperature models .... have been discontinued. Find information on the new temperature models in the documentation [here](<link to docs>).`)

```suggestion
            else
                temperaturemodel = VacuumTemperatureModel{T}(config)
                if model in ("Linear", "Boltzmann")
                    @warn "<Descriptive deprecation warning>"
                else 
                    @info "Config File does not suit any of the predefined temperature models. The drift parameters will not be rescaled."
                end
            end
```

And because we are reusing `PowerLaw` as model string right now, we would also need to check if this is in the old format and also output a deprecation warning.

src/SolidStateDetector/Semiconductor.jl
```diff
@@ -56,13 +56,23 @@ function Semiconductor{T}(dict::AbstractDict, input_units::NamedTuple, outer_tra
...
+
     charge_drift_model = if haskey(dict, "charge_drift_model") && haskey(dict["charge_drift_model"], "model")
         model = Symbol(dict["charge_drift_model"]["model"])
         cdm = if isdefined(SolidStateDetectors, model) && getfield(SolidStateDetectors, model) <: AbstractChargeDriftModel
             if model == :InactiveLayerChargeDriftModel
                 InactiveLayerChargeDriftModel{T}(dict["charge_drift_model"], impurity_density_model, input_units)
             else
-                getfield(SolidStateDetectors, model){T}(dict["charge_drift_model"])
+                getfield(SolidStateDetectors, model){T}(dict["charge_drift_model"], temperature = temperature)
             end
```
fhagemann, line 76: Just a note seeing this in comparison with `InactiveLayerChargeDriftModel{T}`:
 do we completely ignore the `input_units` from the config file here?

docs/src/man/charge_drift.md
```diff
@@ -121,19 +160,49 @@ The `charge_drift_model` needs:
...
 ```julia
 T = SolidStateDetectors.get_precision_type(sim) # e.g. Float32
-charge_drift_model = ADLChargeDriftModel("<path_to_drift_velocity_config_boltzmann.yaml>", T = T, temperature = 100) 
+charge_drift_model = ADL2016ChargeDriftModel("<path_to_ADL_configuration_file>", T = T, temperature = 100u"K") 
 ```
 
 
-If no temperature is given as a parameter, the calculations will be performed at a default temperature of 78 K.
+If no temperature is given as a parameter, the calculations will be performed at a default temperature of 77 K. If no configuration file is given, the default config file `<package_directory>/example/example_config_files/ADLChargeDriftModel/drift_velocity_config_2016.yaml` is loaded. In charge drift configuration files the M.A. Omar and L. Reggiani parametrization can be used by adding the following `temperature_dependence` field like so:
```
fhagemann:
```suggestion
If no temperature is given as a parameter, the calculations will be performed at a default temperature of 77 K. If no configuration file is given, the default config file `<package_directory>/example/example_config_files/ADLChargeDriftModel/drift_velocity_config_2016.yaml` is loaded. In charge drift configuration files, the M.A. Omar and L. Reggiani parametrization can be used by adding the following `temperature_dependence` field like this:
```

docs/src/man/charge_drift.md
```diff
@@ -121,19 +160,49 @@ The `charge_drift_model` needs:
...
+      theta: 200K
+    h:
+      p: 2.4
+      theta: 200K
+```
 
+For convenience, this model can already be found in the `drift_velocity_config.yaml` and `drift_velocity_config_2016.yaml` example config files. 
+
+!!! note
+    The default parameters used in this model were calculated by M.A. Omar and L. Reggiani](https://www.sciencedirect.com/science/article/pii/0038110187900633) before the publication of the `ADLChargeDriftModel` parameters, with $\beta = 2$. Thus, the use of this model to scale the drift parameters in `ADLChargeDriftModel` or `ADL2016ChargeDriftModel` is considered experimental.
```
fhagemann:
```suggestion
    The default parameters used in this model were calculated by [M.A. Omar and L. Reggiani](https://www.sciencedirect.com/science/article/pii/0038110187900633) before the publication of the `ADLChargeDriftModel` parameters, with $\beta = 2$. Thus, the use of this model to scale the drift parameters in `ADLChargeDriftModel` or `ADL2016ChargeDriftModel` is considered experimental. 
```

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawTemperatureModel.jl
hervasa2: How about adding a `scaled::Bool` parametric type to `ADLChargeDriftModel` structs?
```julia
ADL2016ChargeDriftModel{T, M, N, TM, S <: Bool}
```

test/test_charge_drift_models.jl
hervasa2: We could. However I like to have constructor consistency amongst all models, so do we want to write this type of constructor for every model? Is there a specific use case for this constructor in particular?

src/SolidStateDetector/Semiconductor.jl
hervasa2: That has always been the case yes. It's a bit tricky to pass the `input_units` of the config file to the `ChargeDriftModel`.  It could happen that the same charge drift config file (if it has no explicit units in each field) is parsed differently if its included in config files with different `input_units`. This is also the case for any other included files, however since those are usually user created (as opposed to drift configs) the user is probably aware of the set units. In the case of included charge drift configs, the user could incorrectly assume that any given charge drift configs is treated exactly the same irrespective of where it's included. What do you think? 

I think we should pass the `input_units` anyhow. And write good documentation about that happens when a file is included in the config.

hervasa2: I now pass `input_units` and have included some tests

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/TemperatureModels.jl
```diff
@@ -1,4 +1,23 @@
...
+    else
+        "Vacuum"
+    end
+    model_set::String = if model in ("Omar1987", "Vacuum") 
+        model
+    else 
+        if model in ("Linear", "Boltzmann", "PowerLaw")
+            @warn "Since v0.11.0, temperature scaling is only supported at the drift parameter level. The now deprecated models, \"Linear\", \"Boltzmann\" and \"PowerLaw\", relied on the scaling of longitudinal drift velocity at every step of charge carrier drift. This scaling is no longer supported. Please use \"Omar1987\" for the `PowerLawTemperatureModel`, which directly scales charge drift parameters. \nThe drift parameters will not be rescaled."
+        else
+            @info "Config File does not suit any of the predefined temperature models. The drift parameters will not be rescaled."
```
fhagemann: Add to the warning which ones are supported?

test/test_charge_drift_models.jl
```diff
@@ -429,12 +429,24 @@ end
...
+        # charge drift model config has no units, internal units are assumed
         cdm_nounits = ADLChargeDriftModel(joinpath(get_path_to_example_config_files(), "ADLChargeDriftModel/drift_velocity_config_nounits.yaml"))
         @test cdm0.electrons == cdm_nounits.electrons
         @test cdm0.holes == cdm_nounits.holes
         @test cdm0.crystal_orientation ≈ cdm_nounits.crystal_orientation
 
+        # charge drift model config has no units, input units will be applied
+        cdm_nounits_inputunits = ADLChargeDriftModel(joinpath(get_path_to_example_config_files(), "ADLChargeDriftModel/drift_velocity_config_nounits.yaml"), input_units)
+        @test cdm_nounits.electrons != cdm_nounits_inputunits.electrons
+        @test cdm_nounits.holes != cdm_nounits_inputunits.holes
```
fhagemann, line 449: I see your concern with the input units...
Do our example files all have units to their values?

hervasa2: There is one example charge drift config purposefully without units: `drift_velocity_config_nounits`

src/ChargeDriftModels/ADLChargeDriftModel/TemperatureModels/PowerLawTemperatureModel.jl
hervasa2: The reference temperature of the temperature model is now updated when `scale_to_temperature` is applied. This allows for multiple scalings. Test were added to test equivalence of single step vs multi step scaling.

test/test_charge_drift_models.jl
fhagemann: Just to avoid breaking peoples code if they have been using this so far.

test/test_charge_drift_models.jl
fhagemann: We might wanna discourage defining this without units. 
I can see this going wrong when having arbitrary input units...

hervasa2: We could move many of the example drift config files to the test directory. In particular the one without units.

fhagemann: That sounds good. Especially the ones that are not used as default.

fhagemann: Can you check the files and move the unitless ones to the tests?
I think afterwards, this might be ready to go!

fhagemann (approved):

PR #570, opened 2025-12-01
Test Coverage for CartesianPoint: Part 1

claudiaalvgar:
This Pull Request addresses the first part of the list in #552 by adding tests to increase ```CartesianPoint``` coverage.
It also fixes the rotation implementation in ```Polygon.jl```, replacing ```CartesianPoint``` * rotation with ```CartesianVector``` * rotation, which is the correct mathematical operation.

### DIFF
--- a/src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl
+++ b/src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl
@@ -119,13 +119,11 @@ function in(pt::CartesianPoint{T}, p::Polygon{N,T}, csgtol::T = csg_default_tol(
         vs = vertices(p)
 
         pts2d = SVector{N+1, SVector{2,T}}(
-            SVector{2,T}((rot * (vs[i % N + 1] - cartesian_zero)).x,
-                         (rot * (vs[i % N + 1] - cartesian_zero)).y) for i in 0:N
+            [SVector{2,T}(view(rot * (vs[i % N + 1] - cartesian_zero), 1:2)) for i in 0:N]...
         )
-        
-        point2d = SVector{2,T}((rot * (pt - cartesian_zero)).x,
-                                (rot * (pt - cartesian_zero)).y)
 
+        rotated_pt = rot * (pt - cartesian_zero)
+        point2d = SVector{2,T}(view(rotated_pt, 1:2))
         PolygonOps.inpolygon(point2d, pts2d) != 0
     end
 end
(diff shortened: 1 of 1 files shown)


src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl
```diff
@@ -128,6 +128,10 @@ end
     CartesianPoint(pt.x - v[1], pt.y - v[2], pt.z - v[3])
 end
 
+@inline function Base.:*(r::RotationVec{T}, p::CartesianPoint{T}) where T
+    r * SVector{3,T}(p.x, p.y, p.z)
+end
```
fhagemann: Shouldn't this return a `CartesianPoint` as well instead of an `SVector`?

fhagemann: Or probably the mathematical correct way of doing this would not allow for this.
Rotating `CartesianVectors` should work, so wherever this is called, you can replace
```julia
rot * pt
```
with
```julia
cartesian_zero + rot * (pt - cartesian_zero)
````

Comment (claudiaalvgar) [reply src/ConstructiveSolidGeometry/PointsAndVectors/Points.jl]: I think the second option is probably the best one, I added the changes fixing the functions

File: src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl [hunk 2]
```diff
@@ -117,26 +117,34 @@ function in(pt::CartesianPoint{T}, p::Polygon{N,T}, csgtol::T = csg_default_tol(
...
+            SVector{2,T}((rot * (vs[i % N + 1] - cartesian_zero)).x,
+                         (rot * (vs[i % N + 1] - cartesian_zero)).y) for i in 0:N
         )
-        # PolygonOps.inpolygon -> in = 1, on = -1, out = 0)
-        return PolygonOps.inpolygon(view(rot * pt, 1:2), pts2d) != 0 
+        
+        point2d = SVector{2,T}((rot * (pt - cartesian_zero)).x,
+                                (rot * (pt - cartesian_zero)).y)
+
+        PolygonOps.inpolygon(point2d, pts2d) != 0
```
fhagemann, line 127: Does `view` not work on `CartesianVector`? 
I'm a bit worried that you calculate `rot * vs[...]` now twice for every point (once for `x` and once for `y`).
```suggestion
        pts2d = SVector{N+1, SVector{2,T}}(
            SVector{2,T}(
                        view(rot * (vs[i % N + 1] - cartesian_zero), 1:2) for i in 0:N
        )

        point2d = SVector{2,T}(view(rot * (pt - cartesian_zero), 1:2))

        PolygonOps.inpolygon(point2d, pts2d) != 0
```
?

src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl
```diff
@@ -117,26 +117,34 @@ function in(pt::CartesianPoint{T}, p::Polygon{N,T}, csgtol::T = csg_default_tol(
...
-        view(rot * vs[4], 1:2), 
-        view(rot * vs[1], 1:2), 
+
+    pts2d = SVector{5, SVector{2,T}}(
+        SVector{2,T}((rot * (vs[1] - cartesian_zero)).x, (rot * (vs[1] - cartesian_zero)).y),
+        SVector{2,T}((rot * (vs[2] - cartesian_zero)).x, (rot * (vs[2] - cartesian_zero)).y),
+        SVector{2,T}((rot * (vs[3] - cartesian_zero)).x, (rot * (vs[3] - cartesian_zero)).y),
+        SVector{2,T}((rot * (vs[4] - cartesian_zero)).x, (rot * (vs[4] - cartesian_zero)).y),
+        SVector{2,T}((rot * (vs[1] - cartesian_zero)).x, (rot * (vs[1] - cartesian_zero)).y),
     )
```
fhagemann: Can this be shortened by using the same syntax as above?
```suggestion
        pts2d = SVector{5, SVector{2,T}}(
            SVector{2,T}(view(rot * (vs[i % 4 + 1] - cartesian_zero), 1:2)) for i in 0:4
        )
```

src/ConstructiveSolidGeometry/SurfacePrimitives/Polygon.jl
```diff
@@ -117,26 +117,34 @@ function in(pt::CartesianPoint{T}, p::Polygon{N,T}, csgtol::T = csg_default_tol(
...
-    # PolygonOps.inpolygon -> in = 1, on = -1, out = 0)
-    return PolygonOps.inpolygon(view(rot * pt, 1:2), pts2d) != 0 
+
+    point2d = SVector{2,T}((rot * (pt - cartesian_zero)).x, (rot * (pt - cartesian_zero)).y)
```
fhagemann: Also here, check if the `view` method works to avoid calculating `rot * (pt - cartesian_zero)` twice.

PR #571, opened 2025-12-04
Test Coverage for CartesianPoint: Part 2

claudiaalvgar:
This Pull Request addresses the second part of the list in #552 by adding tests to increase ```CartesianPoint``` and ```CartesianVector``` coverage.

### DIFF
--- a/test/ConstructiveSolidGeometry/CSG_coordinates.jl
+++ b/test/ConstructiveSolidGeometry/CSG_coordinates.jl
@@ -129,12 +129,36 @@ using Unitful
 end
 
 @testset "CartesianVector" begin
-    v1 = CartesianVector(1, 2, 3)
-    @test v1.x == 1 && v1.y == 2 && v1.z == 3
-    @test v1 isa CartesianVector{Float64}
 
-    v2 = CartesianVector{Float32}(x=1.0, y=2.0, z=3.0)
-    @test v2.x == 1f0 && v2.y == 2f0 && v2.z == 3f0
+    @test CartesianVector(1, 2, 3) isa CartesianVector{Float64}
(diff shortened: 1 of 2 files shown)


test/ConstructiveSolidGeometry/CSG_coordinates.jl
```diff
@@ -127,3 +127,42 @@ import Unitful: @u_str
         @test_throws ArgumentError CylindricalPoint(1u"m", 2u"rad", 3u"rad")
     end
 end
+
+@testset "CartesianVector" begin
```
fhagemann, line 131: I see that `CartesianVector` can still have units.
Is this something we want to allow for 0.11? Or to restrict to `Real` and then maybe add back later, when also allowing for units in the coordinates of `CartesianPoint`?

claudiaalvgar: This I don't know... The units support in vectors comes from #489 so I guess we can choose if we support them or not

fhagemann: We currently cannot add `CartesianVectors` with units to `CartesianPoints`, so we should think about maybe either stripping the units also off `CartesianVector`, or making sure that units get propagated correctly
image

fhagemann: But we might need some `to_internal_unit` either when creating the `CartesianVector` or in its operations (but it might not only affect `+` but also other functions, so probably stripping the units when creating a `CartesianVector` is the simplest way to go now.)

claudiaalvgar: No problem, I added this modification to match ```CartesianPoint``` behaviour

PR #572, opened 2025-12-08
Adjust default `min_tick_distance` based on the grid dimensions

claudiaalvgar:
This PR introduces a new function ```compute_min_tick_distance``` as a possible solution to replace the previous hard-coded ```min_tick_distance``` in simulation. The new function computes a minimum grid spacing for both Cartesian and Cylindrical grids, taking into account the grid dimensions.
This function provides a finer default grid for detectors of O(100µm) size. The example introduced (BEGe_03.yaml) shows the default grid spacing provided by the simulation with the previous hard-coded method using a small detector:
BEGe_um_size_oldcode_1e-5
and using the new function:
BEGe_um_size_newchange

### DIFF
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -252,7 +252,7 @@ function Grid(sim::Simulation{T, Cylindrical};
         (T[], T[], T[])
     end
 
-    world_r_mid = (world.intervals[1].right + world.intervals[1].left)/2
+    world_r_mid = mean(world.intervals[1])
     if for_weighting_potential && world_Δφ > 0
         world_φ_int = SSDInterval{T, :closed, :open, :periodic, :periodic}(0, 2π)
         world_Δφ = width(world_φ_int)
@@ -904,20 +904,17 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
             if ismissing(min_tick_distance)
                 compute_min_tick_distance(grid)
             elseif min_tick_distance isa LengthQuantity
+                min_distance = T(to_internal_units(min_tick_distance)) 
                 if CS == Cylindrical
-                    world_r_mid = (sim.world.intervals[1].right + sim.world.intervals[1].left)/2 
-                    min_distance_z = min_distance_r = T(to_internal_units(min_tick_distance)) 
-                    min_distance_r, min_distance_z / world_r_mid, min_distance_z 
+                    world_r_mid = mean(sim.world.intervals[1])
+                    min_distance, min_distance / world_r_mid, min_distance 
                 else
-                    min_distance = T(to_internal_units(min_tick_distance))
-                    (min_distance, min_distance, min_distance)
+                    min_distance, min_distance, min_distance
                 end
             else
-                (
-                    T(to_internal_units(min_tick_distance[1])),
-                    T(to_internal_units(min_tick_distance[2])),
-                    T(to_internal_units(min_tick_distance[3]))
-                    )
+                T(to_internal_units(min_tick_distance[1])),
+                T(to_internal_units(min_tick_distance[2])),
+                T(to_internal_units(min_tick_distance[3]))
             end
         end
         


src/Simulation/Simulation.jl
```diff
@@ -849,34 +900,9 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
-                    T(to_internal_units(min_tick_distance[1])),
-                    T(to_internal_units(min_tick_distance[2])),
-                    T(to_internal_units(min_tick_distance[3]))
-                end
-            else
-                (T(1e-5), T(1e-5), T(1e-5))
-            end
-        end
+
+        new_min_tick_distance = compute_min_tick_distance(grid, T, CS)
```
fhagemann: Right now, this is not allowing anymore to set the `min_dick_distance` manually. We might want to add that support back in and just use this method if (and only if) `min_tick_distance` is `missing`.
(Cases in previous lines 864 and 877)

src/Simulation/Simulation.jl
```diff
@@ -812,6 +812,57 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+        y_min, y_max = T(first(y_vals)), T(last(y_vals))
+        z_min, z_max = T(first(z_vals)), T(last(z_vals))
+
+        Δx = clamp((x_max - x_min) * fraction, min_tick, max_tick)
+        Δy = clamp((y_max - y_min) * fraction, min_tick, max_tick)
+        Δz = clamp((z_max - z_min) * fraction, min_tick, max_tick)
+
+        return (Δx, Δy, Δz)
+    end
+end
```
fhagemann: Split this into two functions, because we can multiple dispatch (and we might not even need `CS` and `T` passed as arguments, as this is already saved in the `Grid`):

```suggestion
function compute_min_tick_distance(grid::CylindricalGrid{T}) where {T}
    min_tick = T(1e-12)
    max_tick = T(1e-5)
    fraction = T(1e-3)
  
    r_axis, phi_axis, z_axis = grid.axes

    r_vals = collect(r_axis)
    z_vals = collect(z_axis)

    r_min, r_max = T(first(r_vals)), T(last(r_vals))
    z_min, z_max = T(first(z_vals)), T(last(z_vals))
    r_mid = (r_min + r_max)/2
    safe_r_mid = max(abs(r_mid), T(1e-30))

    r_len = r_max - r_min
    z_len = z_max - z_min

    Δr = clamp(r_len * fraction, min_tick, max_tick)
    Δφ = clamp(Δr / safe_r_mid, min_tick, max_tick)
    Δz = clamp(z_len * fraction, min_tick, max_tick)

    return (Δr, Δφ, Δz)
end
   
function compute_min_tick_distance(grid::CartesianGrid3D{T}) where {T}
    min_tick = T(1e-12)
    max_tick = T(1e-5)
    fraction = T(1e-3)
    
    x_axis, y_axis, z_axis = grid.axes
    x_vals, y_vals, z_vals = collect(x_axis), collect(y_axis), collect(z_axis)
    x_min, x_max = T(first(x_vals)), T(last(x_vals))
    y_min, y_max = T(first(y_vals)), T(last(y_vals))
    z_min, z_max = T(first(z_vals)), T(last(z_vals))

    Δx = clamp((x_max - x_min) * fraction, min_tick, max_tick)
    Δy = clamp((y_max - y_min) * fraction, min_tick, max_tick)
    Δz = clamp((z_max - z_min) * fraction, min_tick, max_tick)

    return (Δx, Δy, Δz)
end
```

and you might not even need to work with `collect`, because you can calculate most of these quantities from `ax.interval`, e.g. `x_max - x_min = width(x_axis.interval)`

test/test_refine_existing_potential.jl
```diff
@@ -21,3 +21,40 @@ T = Float32
...
+        (0.0, 0.1),        # r: 0 → 100 mm
+        (0.0, 120),        # phi
+        (0.0, 0.1)         # z: 0 → 100 mm
+    ))
+    min_tick_large = SolidStateDetectors.compute_min_tick_distance(grid_large, T, CS)
+    @test min_tick_large[3] == 1.0e-5
+    
+    #Test for compute_min_tick_distance in simulation 
+
+    sim = Simulation{T}("test_config_files/BEGe_03.yaml")
```
fhagemann: Why do we need a new config file `BEGe_03.yaml`. Is this just updating the grid?
Maybe, we can make a file where we just change the length unit from `mm` to `µm` and check that we get similar results (with no impurity density, we shouldn't expect tiny to no differences between the two calculations, where right now they might differ).

claudiaalvgar: I added BEGe_03 to test the code in a small detector (plots in the PR description) so it changes the grid and detector size but yes, maybe I can just change the units in one of the existing yaml files (?)

fhagemann: Maybe you can even do that in code without the need of adding a new config file:
```julia
cf = SolidStateDetectors.parse_config_file("BEGe_01.yaml")
cf["units"]["length"] = "µm"
sim = Simulation{T}(cf)
```
(But this only works if we consistently do not provide units to any other quantity in the config file and if the impurity density profile is 0).

fhagemann: But I think it would be kind of nice to show that the results of the simulation does not depend on the length scale. Not sure how easy that is to implement in the tests though ^^

fhagemann: And this is testing it only for the `Cylindrical` case now. Maybe be can also add tests for `Cartesian` (and for the `2D` case)?

claudiaalvgar: The function uses 1e-5 as the default upper limit for the tick distance, and only goes smaller when the length scale of the axis is small enough that length * fraction < 1e-5. The min_tick_distance is scale-dependent by construction, so I don't quite understand what you mean by "the simulation does not depend on the length scale"

fhagemann: Ah sorry, let me rephrase: I would like to test that, roughly, the number of grid points in each dimension should not depend on the length scale, and also the final electric potential is "just scaled" onto a squeezed grid.

fhagemann:
That looks very nice! I can have a closer look later today.

src/Simulation/Simulation.jl
```diff
@@ -813,55 +813,60 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
-        z_min, z_max = T(first(z_vals)), T(last(z_vals))
-        r_mid = (r_min + r_max)/2
-        safe_r_mid = max(abs(r_mid), T(1e-30))
+    z_min = T(z_axis.interval.left)
+    z_max = T(z_axis.interval.right)
+    z_len = width(z_axis.interval)
 
-        r_len = r_max - r_min
-        z_len = z_max - z_min
+    r_mid = (r_min + r_max)/2
```
fhagemann: You can also rewrite this, such that you don't need to define `r_min`, `r_max`, `z_min` and `z_max` anymore.
```suggestion
    r_mid = mean(r_axis.interval)
```

src/Simulation/Simulation.jl
```diff
@@ -812,6 +812,62 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+    z_len = width(z_axis.interval)
+
+    r_mid = (r_min + r_max)/2
+    safe_r_mid = max(abs(r_mid), T(1e-30))
+
+    Δr = clamp(r_len * fraction, min_tick, max_tick)
+    Δφ = clamp(Δr / safe_r_mid, min_tick, max_tick)
+    Δz = clamp(z_len * fraction, min_tick, max_tick)
+
+    return (Δr, Δφ, Δz)
```
fhagemann: We should somewhere make sure that this is type-stable and returns a `Tuple` of type `T`:

```suggestion
    Δr::T = clamp(r_len * fraction, min_tick, max_tick)
    Δφ::T = clamp(Δr / safe_r_mid, min_tick, max_tick)
    Δz::T = clamp(z_len * fraction, min_tick, max_tick)

    return (Δr, Δφ, Δz)
```

src/Simulation/Simulation.jl
```diff
@@ -812,6 +812,62 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+
+    x_axis, y_axis, z_axis = grid.axes
+
+    x_len = width(x_axis.interval)
+    y_len = width(y_axis.interval)
+    z_len = width(z_axis.interval)
+
+    Δx = clamp(x_len * fraction, min_tick, max_tick)
+    Δy = clamp(y_len * fraction, min_tick, max_tick)
+    Δz = clamp(z_len * fraction, min_tick, max_tick)
```
fhagemann: Same thing here:

```suggestion
    Δx::T = clamp(x_len * fraction, min_tick, max_tick)
    Δy::T = clamp(y_len * fraction, min_tick, max_tick)
    Δz::T = clamp(z_len * fraction, min_tick, max_tick)
```

src/Simulation/Simulation.jl
fhagemann, line 845:
```suggestion
function compute_min_tick_distance(grid::CylindricalGrid{T}) where {T}
    min_tick = T(1e-12)
    max_tick = T(1e-5)
    fraction = T(1e-3)

    r_axis, phi_axis, z_axis = grid.axes
    
    r_len = width(r_axis.interval)
    z_len = width(z_axis.interval)
    safe_r_mid = max(abs(mean(r_axis_interval)), T(1e-30))

    Δr::T = clamp(r_len * fraction, min_tick, max_tick)
    Δφ::T = clamp(Δr / safe_r_mid, min_tick, max_tick)
    Δz::T = clamp(z_len * fraction, min_tick, max_tick)

    return (Δr, Δφ, Δz)
```

src/Simulation/Simulation.jl
```diff
@@ -849,34 +905,22 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
-                    T(to_internal_units(min_tick_distance[2])),
-                    T(to_internal_units(min_tick_distance[3]))
-                end 
+      
+        new_min_tick_distance::NTuple{3,T} = begin
+            if ismissing(min_tick_distance)
+                compute_min_tick_distance(grid)
+            elseif min_tick_distance isa LengthQuantity
+                min_distance = T(to_internal_units(min_tick_distance))
+                (min_distance, min_distance, min_distance)
```
fhagemann: This might not be applicable for Cylindrical coordinates..

test/test_refine_existing_potential.jl
```diff
@@ -21,3 +21,95 @@ T = Float32
...
+        "name" => "cylindrical",
+        "r"    => Dict("init" => 0, "gradient" => 0),
+        "z"    => Dict("init" => 0, "gradient" => 0)
+    )
+    sim_cart = Simulation{T}(cf_cart)
+    timed_calculate_electric_potential!(sim_cart)
+    
+    @test length(sim_cart.electric_potential.grid[1]) ≈ 60
+    @test length(sim_cart.electric_potential.grid[2]) ≈ 60
+    @test length(sim_cart.electric_potential.grid[3]) ≈ 62
```
fhagemann: Why are you using `≈` here? ^^
`==` should be fine for integers (?)

test/test_refine_existing_potential.jl
```diff
@@ -21,3 +21,95 @@ T = Float32
...
+    
+    # Cartesian Case
+    cf_cart = deepcopy(cf)
+    cf_cart["grid"]["coordinates"] = "cartesian"
+    cf_cart["grid"]["axes"] = Dict(
+        "x" => Dict("to" => 60),
+        "y" => Dict("to" => 60),
+        "z" => Dict("to" => 60)
+    )
+    cf_cart["units"]["length"] = "µm"
```
fhagemann, line 99: Would we get the same grid lengths when simulating this with `µm` compared to using `mm`? That would be a very nice test if that were to be the case.

test/test_refine_existing_potential.jl
```diff
@@ -105,11 +105,31 @@ end
...
+        "x" => Dict("to" => 60),
+        "y" => Dict("to" => 60),
+        "z" => Dict("to" => 60)
+    )
+    cf_cart_mm["units"]["length"] = "mm"
+    cf_cart_mm["detectors"][1]["semiconductor"]["impurity_density"] = Dict(
+        "name" => "cylindrical",
+        "r"    => Dict("init" => 0, "gradient" => 0),
+        "z"    => Dict("init" => 0, "gradient" => 0)
+    )
```
fhagemann: I like this test! Let's shorten it to 
```suggestion
    cf_cart_mm = deepcopy(cf_cart)
    cf_cart_mm["units"]["length"] = "mm"
```
so you can reuse what was already defined for `cf_cart`

src/Simulation/Simulation.jl
```diff
@@ -849,34 +899,28 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
-                if min_tick_distance isa LengthQuantity
-                    min_distance = T(to_internal_units(min_tick_distance))
-                    min_distance, min_distance, min_distance
+      
+        new_min_tick_distance::NTuple{3,T} = begin
+            if ismissing(min_tick_distance)
+                compute_min_tick_distance(grid)
+            elseif min_tick_distance isa LengthQuantity
+                if CS == Cylindrical
+                    world_r_mid = (sim.world.intervals[1].right + sim.world.intervals[1].left)/2
```
fhagemann:
```suggestion
                    world_r_mid = mean(sim.world.intervals[1])
```

src/Simulation/Simulation.jl
```diff
@@ -849,34 +899,28 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
-                    min_distance, min_distance, min_distance
+      
+        new_min_tick_distance::NTuple{3,T} = begin
+            if ismissing(min_tick_distance)
+                compute_min_tick_distance(grid)
+            elseif min_tick_distance isa LengthQuantity
+                if CS == Cylindrical
+                    world_r_mid = (sim.world.intervals[1].right + sim.world.intervals[1].left)/2 
+                    min_distance_z = min_distance_r = T(to_internal_units(min_tick_distance)) 
+                    min_distance_r, min_distance_z / world_r_mid, min_distance_z
```
fhagemann: A small style comment:
You are using the `min_distance_z` here for `phi`, whereas in `compute_min_tick_distance` you're using `Δr` instead. To be more consistent, I would expect here:

```suggestion
                    min_distance_z = min_distance_r = T(to_internal_units(min_tick_distance)) 
Comment view
                    min_distance_r, min_distance_r / world_r_mid, min_distance_z 
```
But also, you don't really need to define both variables `min_distance_r` and `min_distance_z`, so this can just become:

```suggestion
                    min_distance = T(to_internal_units(min_tick_distance)) 
                    min_distance, min_distance / world_r_mid, min_distance 
```

fhagemann:
It’s not an issue yet, and I don‘t think we need to change anything, but just like a minimum tick distance, there is also a maximum tick distance. I just want to cross-check that, after changing the minimum tick distance, we are not messing anything up by not adjusting the maximum tick distance (default).

PR #573, opened 2025-12-11
Redefinition of offsets in linear and cylindrical impurity/charge densities

fhagemann:
Right now, linear and cylindrical charge densities with gradients in all three dimensions also allowed for an offset in all three dimensions. It makes sense to combine the three offset values into one. 
This would be a breaking change, so I would like to include this in the 0.11 release.

It might make sense to still allow for the old syntax in the configuration file, but to also deprecate it and to push towards a new syntax, that disentangles the offsets from the coordinates.

Right now:
```yaml
impurity_density:
  name: linear
  x:
    init: 1.0e10
    gradient: 1.0e11
```

What I envision is something like this (inverting the logic, and removing `x`, `y` and `z` fields from `init`):
```yaml
impurity_density:
  name: linear
  init: 1.0e10 # (or call this offset?)
  gradient:
    x: 1.0e11
```

The new config file syntax/format is not yet part of this initial set of commits, but might be added within the course of this PR.


src/SolidStateDetectors.jl
```diff
@@ -28,6 +28,7 @@ using StaticArrays
 using StatsBase
 using Unitful
 using UnitfulAtomic
+using UnitfulParsableString
```
fhagemann: Related to #569: this might be needed to correctly write/read units to and from a file

oschulz:
What were the semantics when the offsets were different?

fhagemann:
I like that idea. I will look into this.

src/SolidStateDetectors.jl
fhagemann: @oschulz are you fine with adding this dependency?

oschulz: Hmmm, I'm not sure - that package seems to be built on major type piracy.

fhagemann: Do you have any other elegant way to parse a `Unitful.Quantity` to a string that is then also parsable afterwards?

fhagemann: I have solved this manually now in 2172f53, but this might still result in wrong results when writing a config dict with units to a file using `ssd_read` if the conversion to `Quantity` to a `String` is not parsable the other way around..

src/ImpurityDensities/CylindricalImpurityDensity.jl
```diff
@@ -5,54 +5,50 @@ Impurity density model which assumes a linear gradient in impurity density in ea
...
 function ImpurityDensity(T::DataType, t::Val{:cylindrical}, dict::AbstractDict, input_units::NamedTuple)
-    offsets, gradients = zeros(T,3), zeros(T,3)
     density_unit = input_units.length^(-3)
     density_gradient_unit = input_units.length^(-4)
-    if prod(map(k -> k in ["r","z"], collect(keys(dict)))) @warn "Only r and z are supported in the cylindrical impurity density model.\nChange the impurity density model in the config file or remove all other entries." end
-    if haskey(dict, "r")     
-        if haskey(dict["r"], "init")     offsets[1]   = _parse_value(T, dict["r"]["init"], density_unit) end
-        if haskey(dict["r"], "gradient") gradients[1] = _parse_value(T, dict["r"]["gradient"], density_gradient_unit) end
+    offset::T = _parse_value(T, get(dict, "offset", get(dict, "init", 0)), density_unit)
+    gradients::NTuple{3, T} = let g = get(dict, "gradient", get(dict, "gradients", Dict()))
```
fhagemann, line 36: I now allow for both `offset` and `init`, without a deprecation warning.
I also allow for both `gradient` and `gradients` for the gradients.

src/SolidStateDetectors.jl
oschulz: I think we should stick with the manual solution for now, then, the approach in UnitfulParsableString seems dicey to me.

Issue #567, opened 2025-11-24
Remove hardcoded values from `ThermalDiffusionLithiumDensity` and ` InactiveLayerChargeDriftModel`


hervasa2:
Could we please move the hardcoded values in `ThermalDiffusionLithiumDensity` and ` InactiveLayerChargeDriftModel` (for example in `_calculate_mobility_with_impurities`) to new example config files?

claudiaalvgar:
I think there's not a particular reason for this to be hardcoded, but @zpdg can probably answer this better than me since he implemented it this way

zpdg:
Yes, we can move these values to config files, just like the ADL setup. These values are either from experimental measurements or derived from physical constants, as detailed in the new DL simulation paper (arXiv: 2511.02513).

@claudiaalvgar Do you need me to handle this refactor? I can pick it up, but I'm short on time recently, so there would be a delay.

fhagemann:
@zpdg did you find time to look at this? 
If not, should someone else look into this?

fhagemann:
Maybe someone can comment on the format of the config file that we would envision?
@oschulz @zpdg @hervasa2

fhagemann:
I would say, we might wanna have the OPTION to pass them in the config file, but also have those values as default values if they are not specified in there. We would ideally want to get rid of all the hard-coded "magic" numbers in the source code if possible.

SimonePellegrini:
Ok, I tried a first refactor of ThermalDiffusionLithiumDensity.jl — you can find it in this PR:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/578
This is my first attempt, so feedback is very welcome. If the overall approach makes sense, I’d like to apply a similar refactoring strategy to InactiveLayerChargeDriftModel.jl as well.

PR #578, opened 2025-12-18
Refactor ThermalDiffusionLithiumDensity to remove hardcoded parameters

SimonePellegrini:
This PR addresses #567

### DIFF
--- a/examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yaml
+++ b/examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yaml
@@ -5,12 +5,12 @@ annealing_temperature_ranges:
   - T_min: 473K
     T_max: 873K
     D0: 25e-4cm^2/s
-    H: 11800cal
+    H: 11800cal/mol
   # https://doi.org/10.1103/PhysRev.91.193
   - T_min: 873K
     T_max: 1273K
     D0: 13e-4cm^2/s
-    H: 10700cal    
+    H: 10700cal/mol
   
 experimental_parameters:
   a: 21.27
--- a/src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
+++ b/src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
@@ -102,7 +102,7 @@ function _ThermalDiffusionLithiumParameters(
             _parse_value(T, r["T_min"], temperature_unit),
             _parse_value(T, r["T_max"], temperature_unit),
             _parse_value(T, r["D0"], diffusivity_unit),
-            T.(ustrip(u"cal", _parse_value(T, r["H"], u"cal") * internal_energy_unit)) # convert to cal internally
+            _parse_value(T, r["H"], internal_activation_energy_unit)
         )
         for r in temperature_ranges
     )
--- a/src/MaterialProperties/MaterialProperties.jl
+++ b/src/MaterialProperties/MaterialProperties.jl
@@ -5,7 +5,7 @@ const elementary_charge = Float64(ustrip(u"C", Unitful.q))
 const ϵ0 = Float64(ustrip(u"F/m", Unitful.ϵ0))
 const kB = Float64(ustrip(u"J/K", Unitful.k))
 const me = Float64(ustrip(u"kg", Unitful.me))
-const R_gas = Float64(ustrip(u"cal/K/mol", Unitful.R))
+const R_gas = Float64(ustrip(internal_activation_energy_unit / internal_temperature_unit, Unitful.R))
 
 const material_properties = Dict{Symbol, NamedTuple}()
 
--- a/src/Units.jl
+++ b/src/Units.jl
@@ -17,6 +17,7 @@ const internal_diffusion_unit = internal_length_unit ^ 2 / internal_time_unit
 const internal_charge_density_unit = internal_charge_unit / internal_length_unit ^ 3
 const internal_charge_density_gradient_unit = internal_charge_unit / internal_length_unit ^ 4
 const internal_temperature_unit = u"K"
+const internal_activation_energy_unit = u"cal/mol"
 
 const external_charge_unit  = u"e_au" # elementary charge - from UnitfulAtomic.jl
 
@@ -31,6 +32,7 @@ to_internal_units(x::Quantity{<:Real, dimension(internal_diffusion_unit)})  = us
 to_internal_units(x::Quantity{<:Real, dimension(internal_charge_density_unit)})  = ustrip(internal_charge_density_unit,  x)
 to_internal_units(x::Quantity{<:Real, dimension(internal_charge_density_gradient_unit)}) = ustrip(internal_charge_density_gradient_unit, x)
 to_internal_units(x::Quantity{<:Real, dimension(internal_temperature_unit)})     = ustrip(internal_temperature_unit,  x)
+to_internal_units(x::Quantity{<:Real, dimension(internal_activation_energy_unit)}) = ustrip(internal_activation_energy_unit,  x)
 
 from_internal_units(x::Real, unit::Unitful.Units{<:Any, dimension(internal_time_unit)})    = uconvert(unit, x * internal_time_unit)
 from_internal_units(x::Real, unit::Unitful.Units{<:Any, dimension(internal_voltage_unit)}) = uconvert(unit, x * internal_voltage_unit)


examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yml
```diff
@@ -0,0 +1,19 @@
...
+        H: 11800cal
+      high_T:
+        T_min: 873K
+        T_max: 1273K
+        D0: 1.3e-7m^2/s
+        H: 10700cal    
+      
+experimental_parameters:
+  a: 21.27
+  b: 2610
```
fhagemann: Not sure if I would call this field `impurity_density_profile`, because all of this would be in the config file under `impurity_density` already.
I think `annealing_temperature_ranges` could as well just because a field in the upper-most level here.


```suggestion
model: ThermalDiffusionLithiumDensity
material: HPGe
annealing_temperature_ranges:
    low_T:
      T_min: 473K
      T_max: 873K
      D0: 2.5e-7m^2/s
      H: 11800cal
    high_T:
      T_min: 873K
      T_max: 1273K
      D0: 1.3e-7m^2/s
      H: 10700cal  
experimental_parameters:
  a: 21.27
  b: 2610
```

fhagemann: Or we can also think if we want to allow for more than just exactly two temperatures (`low_T` and `high_T`) and pass the `annealing_temperature_ranges` as a list:
```suggestion
model: ThermalDiffusionLithiumDensity
material: HPGe
annealing_temperature_ranges:
    - T_min: 473K
      T_max: 873K
      D0: 2.5e-7m^2/s
      H: 11800cal
    - T_min: 873K
      T_max: 1273K
      D0: 1.3e-7m^2/s
      H: 10700cal  
experimental_parameters:
  a: 21.27
  b: 2610
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,24 +31,49 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
+    T_min::T
+    T_max::T
+    D0::T
+    H::T
+end
+
+struct LithiumDiffusionParameters{T <: SSDFloat}
+    lowT::LithiumDiffusionBranch{T}
+    highT::LithiumDiffusionBranch{T}
+end
```
fhagemann: This could then become a tuple with variable length `N`:

```suggestion
struct LithiumDiffusionParameters{T <: SSDFloat, N}
    parameters::NTuple{N, LithiumDiffusionBranch{T}}
end
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,24 +31,49 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
+
+struct LithiumSaturationParameters{T <: SSDFloat}
+    a::T
+    b::T
+end
+
+struct ThermalDiffusionLithiumDensityParameters{T <: SSDFloat}
+    diffusion::LithiumDiffusionParameters{T}
+    saturation::LithiumSaturationParameters{T}
+end
```
fhagemann: Or we might not need to define `LithiumDiffusionParameters` at all and this can just become

```suggestion
struct ThermalDiffusionLithiumDensityParameters{T <: SSDFloat,N}
    diffusion::NTuple{N, LithiumDiffusionBranch{T}}
    saturation::LithiumSaturationParameters{T}
end
```
(or maybe then rename `LithiumDiffusionBranch` to `LithiumDiffusionParameters`)

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,24 +31,49 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
+
+include("ThermalDiffusionLithiumDensityParameters.jl")
+
+function calculate_lithium_diffusivity_in_germanium(lithium_annealing_temperature::T,parameters::LithiumDiffusionParameters)::T where {T <: SSDFloat}
     # D0 [m^2*s^-1]
     # H [cal]
-    D0::T, H::T = ifelse(lithium_annealing_temperature <= 873, T.((2.5e-7, 11800)), T.((1.3e-7, 10700)))
+    D0::T, H::T = ifelse(lithium_annealing_temperature <= parameters.lowT.T_max, T.((parameters.lowT.D0, parameters.lowT.H)), T.((parameters.highT.D0, parameters.highT.H)))
```
fhagemann: This could then iterate through the list, assuming it's sorted by increasing temperatures (maybe we would need to add a check that the `Tmax` of the previous entry doesn't overlap with `Tmin` of the following one);

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,24 +31,49 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
+    exp10(parameters.a - parameters.b/lithium_annealing_temperature)
 end
 
-function ThermalDiffusionLithiumDensity{T}(
+function ThermalDiffusionLithiumDensity{T}( 
     lithium_annealing_temperature::T,
     lithium_annealing_time::T,
     contact_with_lithium_doped::G,
     inactive_contact_id::Int;
+    modelParameters::ThermalDiffusionLithiumDensityParameters{T} = ThermalDiffusionLithiumParameters(),
```
fhagemann: In Julia, it's more common to use variable names with snake case:

```suggestion
    model_parameters::ThermalDiffusionLithiumDensityParameters{T} = ThermalDiffusionLithiumParameters(), 
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,96 @@
...
+    highT_H::Union{RealQuantity, String} =
+        config["impurity_density_profile"]["annealing_temperature_ranges"]["high_T"]["H"],
+    a::Union{Real, String} =
+        config["experimental_parameters"]["a"],
+    b::Union{Real, String} =
+        config["experimental_parameters"]["b"],
+    input_units::Union{Missing, NamedTuple} = missing
+)
+
+    temperature_unit = !ismissing(input_units) ? input_units.temperature : u"K"
```
fhagemann:
```suggestion
    temperature_unit = !ismissing(input_units) ? input_units.temperature : internal_temperature_unit
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,96 @@
...
+    diffusivity_unit = !ismissing(input_units) ?
+        input_units.length^2 / internal_time_unit : u"m^2/s"
```
fhagemann:
```suggestion
    diffusivity_unit = !ismissing(input_units) ?
        input_units.length^2 / internal_time_unit : internal_length_unit^2 / internal_time_unit
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,96 @@
...
+
+    lowT = LithiumDiffusionBranch{T}(
+        _parse_value(T, lowT_Tmin, temperature_unit),
+        _parse_value(T, lowT_Tmax, temperature_unit),
+        _parse_value(T, lowT_D0,  diffusivity_unit),
+        _parse_value(T, lowT_H,   u"cal")
```
fhagemann: Should we add `u"cal"` to the internal units?

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,96 @@
...
+    return ThermalDiffusionLithiumDensityParameters{T}(
+        diffusion_params,
+        saturation_params
+    )
+end
+
+
+function ThermalDiffusionLithiumParameters(config::AbstractDict, input_units::Union{Missing, NamedTuple} = missing; kwargs...)
+    if !haskey(config, "impurity_density_profile") 
+        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'impurity_density_profile'."))
```
fhagemann: As noted above, I would remove `impurity_density_profile` as field from the config file.

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,96 @@
...
+
+    if !haskey(config, "experimental_parameters") 
+        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'experimental_parameters'.")) 
+    elseif !haskey(config["experimental_parameters"], "a") 
+        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'experimental_parameters/a'.")) 
+    elseif !haskey(config["experimental_parameters"], "b") 
+        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'experimental_parameters/b'.")) 
+    end
+
+    _ThermalDiffusionLithiumParameters(config; input_units = input_units, kwargs...)
```
fhagemann:
```suggestion
    _ThermalDiffusionLithiumParameters(config; input_units, kwargs...)
```

After the `;`, you can just pass `input_units = input_units` as `; input_units`, that's a nice thing about Julia ^^

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
fhagemann, line 57: You might also want to update this, if people were to pass those parameters using a config file.

examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yml
fhagemann, line 1: It might be that you need this file to be ending in `.yaml` instead of `.yml` in order for `parse_config_file` to work 

https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/7c4b3707b64f31ee3c8814dd2cc499bf89947d76/src/IO/ParseConfigFiles.jl#L27-L29

fhagemann: Yes, this also seems to be the source of the failing tests ^^

examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yml
SimonePellegrini: I think this would be a good option only if you plan to use more than two configurations in the future. Allowing more than two configurations provides greater flexibility, but it also makes error handling more complex at the moment, since you need to check that the list has the correct length, that each element contains the required parameters and so on.

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
SimonePellegrini: I think It is probably better to add it for consistency, but I didn’t know how to name it

examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yml
fhagemann: @zpdg Do you foresee cases in which we might want more than just two `annealing_temperature_ranges`?
How do we deal with things below `473K` and above `1273K`?

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
fhagemann: `(internal_)activation_energy_unit` ?

examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yml
fhagemann: > I think this would be a good option only if you plan to use more than two configurations in the future. Allowing more than two configurations provides greater flexibility, but it also makes error handling more complex at the moment, since you need to check that the list has the correct length, that each element contains the required parameters and so on.

Why would we need to check that list has correct length? My vision would have been to allow for arbitrary length (maybe having at least entry)

SimonePellegrini: What I meant is not that the list must have a fixed length, but that you must check it contains at least the minimum number of entries required by the program logic

zpdg: > @zpdg Do you foresee cases in which we might want more than just two `annealing_temperature_ranges`? How do we deal with things below `473K` and above `1273K`?

It would be nice to keep the options so that we could study the changes of DL in room temperature or study the impact of temperature on the charge collection in a wider range.

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
zpdg: > Should we add `u"cal"` to the internal units?

But we already have the internal "energy" unit. Can we just convert it to "eV"?

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,50 +31,54 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
     # H [cal]
-    D0::T, H::T = ifelse(lithium_annealing_temperature <= parameters.lowT.T_max, T.((parameters.lowT.D0, parameters.lowT.H)), T.((parameters.highT.D0, parameters.highT.H)))
+    if !(parameters[1].T_min ≤ lithium_annealing_temperature ≤ parameters[end].T_max)
+        throw(ArgumentError("Invalid lithium_annealing_temperature=$(lithium_annealing_temperature): expected $(parameters[1].T_min) ≤ lithium_annealing_temperature ≤ $(parameters[end].T_max)."))
+    end
+    for i in 1:length(parameters)
+        if lithium_annealing_temperature <= parameters[i].T_max
+            D0::T, H::T =  T.((parameters[i].D0, parameters[i].H))
+            break
+        end
```
fhagemann:
```suggestion    
    for parameter in parameters
        if lithium_annealing_temperature <= parameter.T_max
            D0::T, H::T =  T.((parameter.D0, parameter.H))
            break
        end
```

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -84,7 +88,8 @@ function ImpurityDensity(T::DataType, t::Val{:li_diffusion}, dict::AbstractDict,
     contact_with_lithium_doped = haskey(dict, "contact_with_lithium_doped") ? dict["contact_with_lithium_doped"] : nothing # you don't have to pass the geometry of doped contact only when the distance_to_contact is passed
     inactive_contact_id = get(dict, "doped_contact_id", -1)
     inactive_contact_id < 1 && error("Invalid doped_contact_id: missing or misspelled key")
-    ThermalDiffusionLithiumDensity{T}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, inactive_contact_id)
+    model_parameters = haskey(dict,"model_parameters") ? ThermalDiffusionLithiumParameters(dict["model_parameters"]) : ThermalDiffusionLithiumParameters()
+    ThermalDiffusionLithiumDensity{T}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, inactive_contact_id,model_parameters)
```
fhagemann:
```suggestion
    ThermalDiffusionLithiumDensity{T}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, inactive_contact_id, model_parameters)
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
-    )
+saturation_params = LithiumSaturationParameters{T}(
+    T(a),
+    T(b)
+)
+
+return ThermalDiffusionLithiumDensityParameters(
+    diffusion_params,
+    saturation_params
+)
```
fhagemann: Indent all of this by one tab

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
     )
+end
 
-    diffusion_params = LithiumDiffusionParameters{T}(lowT, highT)
+for i in 1:length(temp_ranges_parameters)
+    if(temp_ranges_parameters[i].T_max < temp_ranges_parameters[i].T_min)
+        throw(ConfigFileError("Invalid annealing temperature range $i: T_max must be ≥ T_min."))
+    end
+    if(i>1 && temp_ranges_parameters[i-1].T_max != temp_ranges_parameters[i].T_min)
+        throw(ConfigFileError("Annealing temperature ranges must be contiguous and increasing: range $i-1 ends at a different temperature than range $i starts."))
```
fhagemann:
```suggestion
        throw(ConfigFileError("Annealing temperature ranges must be contiguous and increasing: range $(i-1) ends at a different temperature than range $i starts."))
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
+            _parse_value(T, r["T_max"], temperature_unit),
+            _parse_value(T, r["D0"], diffusivity_unit),
+            _parse_value(T, r["H"], u"cal")
+        )
     )
+end
 
-    diffusion_params = LithiumDiffusionParameters{T}(lowT, highT)
+for i in 1:length(temp_ranges_parameters)
+    if(temp_ranges_parameters[i].T_max < temp_ranges_parameters[i].T_min)
```
fhagemann: It's more common to avoid the brackets
```suggestion
    if temp_ranges_parameters[i].T_max < temp_ranges_parameters[i].T_min
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
+        throw(ConfigFileError("Invalid annealing temperature range $i: T_max must be ≥ T_min."))
+    end
+    if(i>1 && temp_ranges_parameters[i-1].T_max != temp_ranges_parameters[i].T_min)
```
fhagemann:
```suggestion
    if i > 1 && temp_ranges_parameters[i-1].T_max != temp_ranges_parameters[i].T_min
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
+diffusion_params = Tuple(temp_ranges_parameters)
 
-    return ThermalDiffusionLithiumDensityParameters{T}(
-        diffusion_params,
-        saturation_params
-    )
+saturation_params = LithiumSaturationParameters{T}(
+    T(a),
+    T(b)
+)
```
fhagemann: This right now does not cover the case if they are strings.
You might wanna use `_parse_value` here:
```suggestion
saturation_params = LithiumSaturationParameters{T}(
    _parse_value(T, a, Unitful.NoUnits), _parse_value(T, b, Unitful.NoUnits)
)
```
(Or whatever the units of this might be)

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
-    elseif !haskey(config["impurity_density_profile"]["annealing_temperature_ranges"], "low_T")
-        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'impurity_density_profile/annealing_temperature_ranges/low_T'."))
-    elseif !haskey(config["impurity_density_profile"]["annealing_temperature_ranges"], "high_T")
-        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'impurity_density_profile/annealing_temperature_ranges/high_T'."))
-    end
+    if !haskey(config, "annealing_temperature_ranges") 
+        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'annealing_temperature_ranges'.")) 
+    elseif length(config["annealing_temperature_ranges"])<minimum_num_of_temp_ranges
+        throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs at least $minimum_num_of_temp_ranges entries in 'annealing_temperature_ranges'."))
+    end
```
fhagemann: What's wrong with having just one? (If it spans a wide temperature range)?
I would even suggest here to remove the `const minimum_num_of_temperature_ranges` and just check that it has at least one (valid) entry, i.e. `isempty(config[...])`)

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -1,81 +1,65 @@
...
-        if !haskey(config["impurity_density_profile"]["annealing_temperature_ranges"]["low_T"], axis)
-            throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'impurity_density_profile/annealing_temperature_ranges/low_T/$(axis)'."))
-        end
-        if !haskey(config["impurity_density_profile"]["annealing_temperature_ranges"]["high_T"], axis)
-            throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'impurity_density_profile/annealing_temperature_ranges/high_T/$(axis)'."))
+    for i in 1:length(config["annealing_temperature_ranges"])
+        for axis in ("T_min", "T_max", "D0", "H")
+            if !haskey(config["annealing_temperature_ranges"][i], axis)
+                throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'annealing_temperature_ranges[$i]/$(axis)'."))
+            end
```
fhagemann: Thinking out loud, but this should be in a follow up PR and definitely not in this one:
I would argue that one could also be able to skip `T_max` (except for the last one) and just use `T_min` from the next entry. (Or skip `T_min` and set it to `T_max` from the previous entry).

fhagemann:
```suggestion
    for i in eachindex(config["annealing_temperature_ranges"])
        for axis in ("T_min", "T_max", "D0", "H")
            if !haskey(config["annealing_temperature_ranges"][i], axis)
                throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'annealing_temperature_ranges[$i]/$(axis)'."))
            end
```
or
```suggestion
    for (i, temperature_range) in enumerate(config["annealing_temperature_ranges"])
        for axis in ("T_min", "T_max", "D0", "H")
            if !haskey(temperature_range, axis)
                throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'annealing_temperature_ranges[$i]/$(axis)'."))
            end
```

And should we also add a warning here if there are fields with other namings? Something like:
```suggestion
    for (i, temperature_range) in enumerate(config["annealing_temperature_ranges"])
        for axis in ("T_min", "T_max", "D0", "H")
            if !haskey(temperature_range, axis)
                throw(ConfigFileError("ThermalDiffusionLithiumDensity config file needs entry 'annealing_temperature_ranges[$i]/$(axis)'."))
            end
        end
        for k in keys(temperature_range)
            if !(k in ("T_min", "T_max", "D0", "H")
                @warn "Encountered unexpected key `$(k)` in 'annealing_temperature_ranges[$i]': Allowed keys are `T_min`, `T_max`, `D0` and `H`. Key `$(k)` will be ignored."
            end
        end
```

Project.toml
fhagemann, line 1: Can you undo these changes to the `Project.toml`?
You don't want `LegendHDF5IO` and `TimerOutputs` added here.
You should be fine using the `test/Project.toml` for this.

examples/example_config_files/ThermalDiffusionLithiumDensity/Thermal_Diffusion_Config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    T_max: 873K
+    D0: 2.5e-7m^2/s
+    H: 11800cal
+  - T_min: 873K
+    T_max: 1273K
+    D0: 1.3e-7m^2/s
+    H: 10700cal    
+  
+experimental_parameters:
+  a: 21.27
```
fhagemann, line 16: Maybe this is what is causing the failing tests: the parameter used to be 27.27, not 21.27.
Or is it 21.27 in the publication?

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,25 +31,56 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
+    for i in 1:length(parameters)
+        if lithium_annealing_temperature <= parameters[i].T_max
+            D0::T, H::T =  T.((parameters[i].D0, parameters[i].H))
+            break
+        end    
+    end
     D0 * exp(-H/(R_gas*lithium_annealing_temperature))
 end
-function calculate_lithium_saturated_density(lithium_annealing_temperature::T)::T where {T <: SSDFloat}
-    exp10(27.27 - 2610.0/lithium_annealing_temperature)
```
fhagemann, line 41: Talking about this line.

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
```diff
@@ -31,25 +31,56 @@ struct ThermalDiffusionLithiumDensity{T <: SSDFloat} <: AbstractImpurityDensity{
...
     contact_with_lithium_doped::G,
     inactive_contact_id::Int;
+    model_parameters::ThermalDiffusionLithiumDensityParameters{T,N} = ThermalDiffusionLithiumParameters(), 
     distance_to_contact::Function = pt::AbstractCoordinatePoint{T} -> ConstructiveSolidGeometry.distance_to_surface(pt, contact_with_lithium_doped),
-    lithium_density_on_contact::T = calculate_lithium_saturated_density(lithium_annealing_temperature),
-    lithium_diffusivity_in_germanium::T = calculate_lithium_diffusivity_in_germanium(lithium_annealing_temperature),
-) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}}
+    lithium_density_on_contact::T = calculate_lithium_saturated_density(lithium_annealing_temperature,model_parameters.saturation),
+    lithium_diffusivity_in_germanium::T = calculate_lithium_diffusivity_in_germanium(lithium_annealing_temperature,model_parameters.diffusion),
+) where {T <: SSDFloat,N, G <: Union{<:AbstractGeometry, Nothing}}
```
fhagemann: You can remove `N` from here, because it's not used in the function afterwards.
```suggestion
function ThermalDiffusionLithiumDensity{T}( 
    lithium_annealing_temperature::T,
    lithium_annealing_time::T,
    contact_with_lithium_doped::G,
    inactive_contact_id::Int;
    model_parameters::ThermalDiffusionLithiumDensityParameters{T} = ThermalDiffusionLithiumParameters(), 
    distance_to_contact::Function = pt::AbstractCoordinatePoint{T} -> ConstructiveSolidGeometry.distance_to_surface(pt, contact_with_lithium_doped),
    lithium_density_on_contact::T = calculate_lithium_saturated_density(lithium_annealing_temperature,model_parameters.saturation),
    lithium_diffusivity_in_germanium::T = calculate_lithium_diffusivity_in_germanium(lithium_annealing_temperature,model_parameters.diffusion),
) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}}`
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,80 @@
+function _ThermalDiffusionLithiumParameters(
+    config::AbstractDict;
+    T::Type = Float32,
```
fhagemann: This line seems to break this:
```julia
sim = Simulation{Float64}(SSD_examples[:TrueCoaxial])
```

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,80 @@
...
+
+temp_ranges_parameters = Vector{LithiumDiffusionParameters{T}}()
+
+for r in temperature_ranges
+    push!(temp_ranges_parameters,
+        LithiumDiffusionParameters{T}(
+            _parse_value(T, r["T_min"], temperature_unit),
+            _parse_value(T, r["T_max"], temperature_unit),
+            _parse_value(T, r["D0"], diffusivity_unit),
+            _parse_value(T, r["H"], u"cal")
```
fhagemann: So, what's causing the error is that `H` is converted to `eV` (internal energy units) but treated in the formula as if it was still in `cal`.

Image

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
fhagemann: --> fixed in 4030c66

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
fhagemann: I checked: it's 21.27 in Eq.(3) of https://arxiv.org/abs/2511.02513
@zpdg: can you confirm it should be 21.27 here

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
fhagemann: --> Fixed in 7f69bc8

Project.toml
fhagemann: --> reverted in 0dfbc85

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
fhagemann: --> storing it internally in units of `cal` now: 7f69bc8
If we switch to `eV`, we might need to update the definition of `R_gas` as well.

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
fhagemann: Never mind, I think I figured this out, see 8889f06.
The publication states `21.27` (to get the impurity density in units of `cm^-3`), but we had `27.27` to account for the fact that we expect the density in units of `m^-3`.
I now explicitly separated the 27.27 we had into `21.27` (From the config file) and `6` (for the order of magnitude) and added a comment to the code to avoid confusion.

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
SimonePellegrini: Yeah, you're right. When I implemented it, I was thinking about the original code having two ranges, but they are not necessarily required and are defined in the YAML file.

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,93 @@
...
+    for r in temperature_ranges
+        push!(temp_ranges_parameters,
+            LithiumDiffusionParameters{T}(
+                _parse_value(T, r["T_min"], temperature_unit),
+                _parse_value(T, r["T_max"], temperature_unit),
+                _parse_value(T, r["D0"], diffusivity_unit),
+                _parse_value(T, r["H"], u"cal") / ustrip(Unitful.NoUnits, one(T) * u"cal" / internal_energy_unit) # convert to cal internally
+            )
+        )
+    end
```
fhagemann: This is double now and should not be needed anymore.

src/ImpurityDensities/ThermalDiffusionLithiumDensityParameters.jl
```diff
@@ -0,0 +1,159 @@
...
+```
+
+## Parametric types
+* `T`: Precision type.
+    
+## Fields
+* `T_min::T`: Minimum temperature (in K) for which the parameters are valid.
+* `T_max::T`: Maximum temperature (in K) for which the parameters are valid.
+* `D0::T`: Diffusivity constant in m^2/s.
+* `H::T`: Activation energy in cal/mol.
```
fhagemann, line 16: One thing I've seen in the publication, is that the value for `H` is reported in `cal/mol`.
Should we change the code to expect `H` in units of `cal/mol` rather than `cal` (and then converting it to `eV` because both are energies)?

zpdg: > One thing I've seen in the publication, is that the value for `H` is reported in `cal/mol`. Should we change the code to expect `H` in units of `cal/mol` rather than `cal` (and then converting it to `eV` because both are energies)?

That would be better. We used to write "H" in unit of "cal" and "R" in "cal/K" following old references, but it seems that current standard units for H and R are with a "mol^-1".

fhagemann: I changed `H` to be `cal/mol` in feda000. I think this would be the most elegant solution.

PR #594, opened 2026-02-17
Surface Grid Refinement

claudiaalvgar:
This PR introduces surface-only grid refinement for electric potential simulations and fixes boundary-related indexing issues that could lead to segmentation faults at high refinement resolutions e.g for 1e-5m grid spacing. In addition, it also fixes an existent issue when calculating the electric potential with custom grids for small grid spacing. For custom small grid spacing (e.g. 2e-5m), the detector appeared to be undepleted while for bigger grid spacing (1e-4m) the behaviour was correct. This issue was not related to the new RCC layer model since this was also spotted for simulations not including the RCC layer model. See the issue below:
---> With RCC layer custom initial grid:
* 0.02mm grid spacing: wrong behaviour (I attach the code in case this issue has to be reproduced):
```
T = Float64 
calculate_electric_potential!(sim ,max_n_iterations = 10 ,grid= Grid(sim) ,verbose = true ,depletion_handling = true)
ax1, ax2, ax3 = g.axes
tick_dis=0.02*mm 
user_additional_ticks_ax1=sort(vcat(ax1.interval.left:tick_dis:ax1.interval.right)) 
user_additional_ticks_ax3=sort(vcat(ax3.interval.left:tick_dis:ax3.interval.right))[2:end] # only support even number of ticks in z-direction 
user_ax3 = typeof(ax3)(ax3.interval, user_additional_ticks_ax3) 
user_g = typeof(g)((user_ax1, ax2, user_ax3)) 
calculate_electric_potential!(sim, refinement_limits = [0.2, 0.1, 0.05, 0.02], use_nthreads=8, grid = user_g, depletion_handling = true)
```
InvertedCoax_PR3_tick0 02mm_refinements
* 0.1mm grid spacing custom initial grid: correct behaviour:
InvertedCoax_PR3_tick0 1mm_refinements

---> Without RCC layer but using custom grid: same issue was observed:
* 0.02mm custom grid spacing: wrong behaviour
InvertedCoax_noinactivelayer_PR3_tick0 02mm_refinements
* 0.1mm custom grid spacing: correct behaviour
InvertedCoax_noinactivelayer_PR3_tick0 1mm_refinements
The new implementation checks whether a given slice of point types contains any surface (inactive layer) points and builds a new refined grid across all three axes. Surface refinement is automatically triggered during the electric potential calculation when:

- A surface impurity model is present

- At least 3 refinement steps are used in the electric potential calculation

- Final refinement limit ≤ 0.05

A YAML entry was added to configure spacing_surface_refinement, that controls the biggest grid spacing allowed in the surface for the 3 axis. The new method identifies the surface regions and it just refines in this area as can be seen, for example for a 0.1mm spacing:
Having this ```spacing_surface_refinement: [1e-4, 1e-4, 1e-4]``` in the yaml file and running the default: ```calculate_electric_potential!(sim, use_nthreads=8, grid=g, verbose = true, depletion_handling = true)``` will now give the correct behaviour for the surface layer, or even: ```calculate_electric_potential!(sim, refinement_limits = [0.2, 0.1, 0.05, 0.02], use_nthreads=8, grid = user_g, depletion_handling = true)``` if we want to further refine the bulk:

DeadLayerGrid_norefinements_1e-4min_spacingdeadlayer_surfafter_3rdref_grids

DeadLayerGrid_norefinements_1e-4min_spacingdeadlayer_surfafter_3rdref_griddensity

DeadLayerGrid_norefinements_1e-4min_spacingdeadlayer_surfafter_3rdref_epot
The weighting potential is also well behaved even when passing the custom grid from the surface-refined electric potential:

DeadLayerGrid_norefinements_1e-4min_spacingdeadlayer_surfafter_3rdref_wp_customgrid_from_epot

This provides a faster calculation O(3min) than for the custom grid case O(10min) since the initial grid was refined from the very beginning, see old case behaviour for custom grid:

DeadLayerGrid_norefinements_custominitialgrid0 1mm_griddensity
Custom_grid_0 1mm_noref_nosurfref_epot
The surface refinement is applied after three initial bulk refinements (the default refinement setup for the electric potential calculation) because only at that stage does the surface behaviour become well-resolved. Therefore, this model requires the default bulk refinement setup in order to correctly apply the surface refinement:
Default:  refinement_limits = [0.2, 0.1, 0.05]) + surface refinement. 
Applying the surface refinement after the first or second refinement step does not fully capture the surface layer, leading to incomplete refinement, as you can see :

- Surface Refinement after 2nd refinement uses [0.2, 0.1, -surface ref- 0.05]) and min_spacing = (1e-4, 1e-4, 1e-4)m in surface layer
  
DeadLayerGrid_norefinements_1e-4min_spacingdeadlayer_surfafter_2ndref

- Initial surface refinement (this uses -surface ref- + [0.2, 0.1, 0.05]) and min_spacing = (1e-4, 1e-4, 1e-4)m in surface layer

DeadLayerGrid_norefinements_1e-4min_spacingdeadlayer_surfbefore
This PR implements the default behaviour (3 refinements) plus the surface refinement, ensuring accurate resolution near surfaces. Additionally, the model now supports refining surface intervals down to 1e-5m, preserving correct physical behaviour and avoiding the incorrect results that can occur with custom grids resulting in undepleted detectors.

- New method 0.02mm spacing: Correct behaviour

DeadLayerGrid_norefinements_2e-5min_spacingdeadlayer_surfafter_3rdref_grids

DeadLayerGrid_norefinements_2e-5min_spacingdeadlayer_surfafter_3rdref_griddensity
DeadLayerGrid_norefinements_2e-5min_spacingdeadlayer_surfafter_3rdref_epot

NOTE: this model has not been tested on cartesian grid defined detectors but I wanted to ask for inputs before. I wasn't sure neither if this should have been a Draft PR, so feel free to move it.


src/Simulation/Simulation.jl
```diff
@@ -38,6 +38,7 @@ mutable struct Simulation{T <: SSDFloat, CS <: AbstractCoordinateSystem} <: Abst
     electric_potential::Union{ElectricPotential{T}, Missing}
     weighting_potentials::Vector{Any}
     electric_field::Union{ElectricField{T}, Missing}
+    spacing_surface_refinement::Union{NTuple{3,T}, Nothing}
```
fhagemann: I'm not a fan of adding this as a field to `Simulation`.
Could this be something that could be added to the `World` in `sim.world`?

examples/example_config_files/ivc_inactivelayer.yaml
```diff
@@ -20,6 +20,7 @@ grid:
       boundaries:
         left: inf
         right: inf
+  spacing_surface_refinement: [1e-4, 1e-4, 1e-4] #m
```
fhagemann, line 23: If it's here semantically, I would also add this to the `World` that we create from it and save it to `sim.world`.

Not necessarily for this PR, but we could think about also allowing to put things like the refinement limits, max/min tick distance etc. in here, so that they do not have to be passed at runtime to the `calculate_potential!` methods.

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl
```diff
@@ -168,36 +167,50 @@ end
...
-        ϵ_r[ in3,  i2,  i1 ],
-        ϵ_r[  i3,  i2,  i1 ]
+        ϵ_r[in3_safe, in2_safe, in1_safe],
+        ϵ_r[i3_safe,  in2_safe, in1_safe],
+        ϵ_r[in3_safe,  i2_safe, in1_safe],
+        ϵ_r[i3_safe,   i2_safe, in1_safe],
+        ϵ_r[in3_safe, in2_safe,  i1_safe],
+        ϵ_r[i3_safe,  in2_safe,  i1_safe],
+        ϵ_r[in3_safe,  i2_safe,  i1_safe],
+        ϵ_r[i3_safe,   i2_safe,  i1_safe]
```
fhagemann: Why is this needed now? Was it needed before already (is this a bug fix)?
Or is this something required now with the things you have added in this PR?

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl
```diff
@@ -168,36 +167,50 @@ end
...
-        ϵ_r[  i1,  i2, in3 ],
-        ϵ_r[  i1,  i2,  i3 ]
+        ϵ_r[in1_safe, in2_safe, in3_safe],
+        ϵ_r[in1_safe, in2_safe,  i3_safe],
+        ϵ_r[in1_safe,  i2_safe, in3_safe],
+        ϵ_r[in1_safe,  i2_safe,  i3_safe],
+        ϵ_r[ i1_safe, in2_safe, in3_safe],
+        ϵ_r[ i1_safe, in2_safe,  i3_safe],
+        ϵ_r[ i1_safe,  i2_safe, in3_safe],
+        ϵ_r[ i1_safe,  i2_safe,  i3_safe]
```
fhagemann: Same thing here.

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,78 @@ end
...
 _extend_refinement_limits(rl::Real) = (rl, rl, rl )
 _extend_refinement_limits(rl::Tuple{<:Real,<:Real,<:Real}) = rl
+
+@inline function has_surface_points(slice::AbstractArray{PointType})::Bool
+    return any(is_in_inactive_layer, slice)
+end
+
+function _refine_axis_surface( ax::DiscreteAxis{T, <:Any, <:Any, ClosedInterval{T}}, surface_intervals::AbstractVector{Bool}, min_spacing::T;
+    extra_before::Int = 5,  # intervals to refine before first surface interval
+    extra_after::Int = 5    # intervals to refine after last surface interval
```
fhagemann, line 184: What do you mean by "intervals"?

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,78 @@ end
...
+    old_ticks = ax.ticks
+    n_int = length(surface_intervals)
+
+    # Find first and last surface intervals
+    first_surface = findfirst(surface_intervals)
+    last_surface  = findlast(surface_intervals)
+
+    if first_surface === nothing || last_surface === nothing
+        return ax # Nothing to refine
+    end
```
fhagemann:
```suggestion
    (isnothing(first_surface) || isnothing(last_surface)) && return ax
```

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,78 @@ end
...
+
+    # Create flags for all intervals to refine
+    refine_flags = copy(surface_intervals)
```
fhagemann: Do you need `copy` or `deepcopy` here?
I believe using `copy` would be equivalent to just having 
```julia
refine_flags = surface_intervals
```

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,78 @@ end
...
+
+    # Add extra intervals before first surface interval
+    start_idx = max(first_surface - extra_before, 1)
+    refine_flags[start_idx:first_surface-1] .= true
```
fhagemann, line 198: Again what is `interval`? Do you mean grid ticks or grid tick index? Or something else?

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,78 @@ end
...
+    i_tick = 1
+    for j in 1:n_int
+        ticks[i_tick] = old_ticks[j]
+        for k in 1:ns[j]
+            i_tick += 1
+            ticks[i_tick] = old_ticks[j] + k*sub_widths[j]
+        end
+        i_tick += 1
+    end
+    ticks[end] = old_ticks[end]
```
fhagemann, line 237: Is there any scenario in which some values in `ticks` might remain `undef`?

src/Simulation/Simulation.jl
```diff
@@ -178,13 +181,19 @@ function Simulation{T}(dict::AbstractDict)::Simulation{T} where {T <: SSDFloat}
...
     sim.world = if haskey(dict, "grid") && isa(dict["grid"], AbstractDict) && haskey(dict["grid"], "axes")
-            World(T, dict["grid"], sim.input_units)
+        World(T, dict["grid"], sim.input_units)
         else let det = sim.detector 
             world_limits = get_world_limits_from_objects(CS, det)
             World(CS, world_limits)
         end
     end
     sim.weighting_potentials = Missing[ missing for i in 1:length(sim.detector.contacts)]
+    sim.spacing_surface_refinement = if haskey(dict, "grid") && isa(dict["grid"], AbstractDict) &&
```
fhagemann:
```suggestion
    sim.world.spacing_surface_refinement = if haskey(dict, "grid") && isa(dict["grid"], AbstractDict) &&
```

src/Simulation/Simulation.jl
```diff
@@ -178,13 +181,19 @@ function Simulation{T}(dict::AbstractDict)::Simulation{T} where {T <: SSDFloat}
...
+                                       haskey(dict["grid"], "spacing_surface_refinement")
+        ntuple(i -> T(dict["grid"]["spacing_surface_refinement"][i]), 3)
```
fhagemann: Can you use `_parse_value` something equivalent here?
Then, you might get parsing values with units for free :)

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
     nothing
 end
 
+"""
+refine_surface!(sim::Simulation{T};
```
fhagemann:
```suggestion
    refine_surface!(sim::Simulation{T};
```
(Indent the line describing the function name, so it appears as code in the docstring)

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+                    min_spacing::NTuple{3,T} = (T(1e-4), T(1e-4), T(1e-4)),
+                    not_only_paint_contacts::Bool = true,
+                    paint_contacts::Bool = true,
+                    update_other_fields::Bool = true) where {T <: SSDFloat}
+
+Refine the simulation's electric potential grid **only in regions containing surface points**, keeping other regions unchanged.
+
+This function:
+
+1. Identifies which intervals along each axis contain surface points.
```
fhagemann, line 828: Interval = the range between two grid points?

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
     nothing
 end
 
+"""
+refine_surface!(sim::Simulation{T};
+                    min_spacing::NTuple{3,T} = (T(1e-4), T(1e-4), T(1e-4)),
```
fhagemann: According to the docstring, this would be a keyword argument, but it actually seems to be an argument. Move the `;` from before to behind `min_spacing`

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+4. Updates the simulation's `electric_potential` and optionally dependent fields (`imp_scale`, `q_eff_imp`, `q_eff_fix`, `ϵ_r`, `point_types`).
+
+Special behaviour:
+
+- In cylindrical coordinates, the φ-axis (2nd axis) is left unchanged except for closing the axis for interpolation purposes.
+- Handles both 2D and 3D grids.
+
+# Arguments
+- `sim::Simulation{T}`: The simulation object containing the electric potential to refine.
+- `min_spacing::NTuple{3,T}`: Minimum allowed spacing in the surface after refinement. Default is `(1e-4, 1e-4, 1e-4)m`.
```
fhagemann: Is this default spacing something that should depend on the dimensions of `grid.world`?

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+"""
+
+function refine_surface!(sim::Simulation{T}, min_spacing::NTuple{3,T} = (T(1e-4), T(1e-4), T(1e-4));
+                         not_only_paint_contacts::Bool = true,
+                         paint_contacts::Bool = true,
+                         update_other_fields::Bool = true) where {T <: SSDFloat}
+    
+    old_grid = sim.electric_potential.grid
+
+    # Enforce minimum and maximum spacing for surface refinement (10 µm < min_spacing < 1 mm)
```
fhagemann: Why `10µm` and `1 mm`?
Out of experience? Does this hold for any dimension of `grid.world`
(you might be able to test this by setting your ICPC length scale from `mm` to `µm` and check if everything still works as intended -- this might maybe require also tweaking the impurity density down by a factor to resemble the same problem, just on another length scale).

This would also be a great example for a test case to see that what SSD does shouldn't be scale-dependent, if one were to shrink down everything by a factor.

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+    for d in 1:3
+        if length(old_grid.axes[d].ticks) != 1
+            if min_spacing[d] < T(1e-5) || min_spacing[d] > T(1e-3)
+                @warn "min_spacing[$d] = $(min_spacing[d]) m is out of bounds (1e-5 m < min_spacing < 1e-3 m). Higher values don't require surface refinement"
```
fhagemann: What about lower values?

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+                view(sim.point_types.data, :, i, :)
+            else
+                view(sim.point_types.data, :, :, i)
+            end
+            has_surface_points(slice)
+        end, 1:n_int)
+    end, 3)
+
+    # Refine axes: skip phi-axis if Cylindrical
+    is_cyl = old_grid isa Grid{<:Any, 3, Cylindrical}
```
fhagemann: Why `<:Any`? Shouldn't this be the `T` of the `Simulation{T}`?

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+                          else
+                              _refine_axis_surface( old_grid.axes[i], surface_intervals[i], min_spacing[i],)
+                          end
+                      end, 3)
+    
+    if is_cyl
+        new_grid = Grid{T,3,Cylindrical}((new_axes[1], new_axes[2], new_axes[3]))
+    else
+        new_grid = Grid{T,3,Cartesian}((new_axes[1], new_axes[2], new_axes[3]))
+    end
```
fhagemann: If in the function call, you define the type of `sim` to be `sim::Simulation{T,CS} where {T,CS}`, then you can use the coordinate system type here:


```suggestion
    new_grid = Grid{T,3,CS}((new_axes[1], new_axes[2], new_axes[3]))
```

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+            if min_spacing[d] < T(1e-5) || min_spacing[d] > T(1e-3)
+                @warn "min_spacing[$d] = $(min_spacing[d]) m is out of bounds (1e-5 m < min_spacing < 1e-3 m). Higher values don't require surface refinement"
+                if min_spacing[d] < T(1e-5)
+                    @warn "Using min_spacing[$d] = 1e-5 m"
+                else
+                    @warn "Using min_spacing[$d] = 1e-3 m"  
+                end
+            end
+        end
+    end
```
fhagemann: How about avoiding throwing a warning for the phi-axis in cylindrical coordinates.

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+        new_grid = Grid{T,3,Cylindrical}((new_axes[1], new_axes[2], new_axes[3]))
+    else
+        new_grid = Grid{T,3,Cartesian}((new_axes[1], new_axes[2], new_axes[3]))
+    end
+
+    # Interpolate potential onto new grid
+    closed_pot = _get_closed_potential(sim.electric_potential)
+    new_data = Array{T,3}(undef, size(new_grid))
+    only2d = size(closed_pot.data, 2) == 1
+    int = interpolate_closed_potential(closed_pot, Val(only2d))
```
fhagemann: Would this perform correctly in Cartesian coordinates?
Would `only2d` also require some `is_cyl` check?

src/Simulation/Simulation.jl
```diff
@@ -812,6 +823,127 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+    end
+    
+    # Update electric potential in simulation
+    sim.electric_potential = _convert_to_original_potential(sim.electric_potential, new_data, new_grid)
+    
+    # Update dependent fields
+    if update_other_fields
+
+        pcs = PotentialCalculationSetup(sim.detector, sim.electric_potential.grid, sim.medium, sim.electric_potential.data,
+                                        not_only_paint_contacts = not_only_paint_contacts, paint_contacts = paint_contacts)
```
fhagemann: You could shorten this to

```suggestion
        pcs = PotentialCalculationSetup(sim.detector, sim.electric_potential.grid, sim.medium, sim.electric_potential.data;
                                        not_only_paint_contacts, paint_contacts)
```

src/Simulation/Simulation.jl
```diff
@@ -899,7 +1031,26 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
         else
             sor_consts = T.(sor_consts)
         end
-      
+        has_surface_model = isdefined(sim.detector.semiconductor.impurity_density_model, :surface_imp_model)
+
+        if has_surface_model    
+            if sim.spacing_surface_refinement === nothing
```
fhagemann: I usually use `isnothing` to check equality,
not sure if there is a preferred way:

```suggestion
            if isnothing(sim.spacing_surface_refinement)
```

src/Simulation/Simulation.jl
```diff
@@ -1015,6 +1170,23 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
                                                 not_only_paint_contacts = not_only_paint_contacts, 
                                                 paint_contacts = paint_contacts,
                                                 sor_consts = is_last_ref ? T(1) : sor_consts )
+                
+                if has_surface_model && iref==3 && sim.spacing_surface_refinement !== nothing
```
fhagemann:
```suggestion
                if has_surface_model && iref == 3 && !isnothing(sim.spacing_surface_refinement)
```

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl
claudiaalvgar: This bit of the code was giving a bound error when adding more grid points around the surface layer. When refining the grid near the surface, some index would fall just outside the valid array range. Previously this did not show up because the grid configurations we used did not hit these boundary cases, but with the refinement introduced in this PR it can occur and leads to a BoundsError. So the clamping would act as a safeguard to keep the indices within the range. In that sense this is effectively a bug fix for boundary handling, but it only became visible with the refinement changes added in this PR

src/PotentialCalculation/Refinement.jl
claudiaalvgar: The intervals are defined by the distance between two consecutive grid points

src/Simulation/Simulation.jl
claudiaalvgar: yes

src/Simulation/Simulation.jl
claudiaalvgar: ```min_spacing``` controls the minimum allowed cell size during the surface refinement step, so it doesn't describe the physical extent of the simulation domain. 
It is only relevant when a surface model is present; otherwise the parameter is effectively unused, so I don't quite see why it should be in ```world``` since it depends if a surface model is defined or not to be useful. The 1e-4m is a leftover code since if ```spacing_surface_refinement``` is not defined in the yaml file, it won't perform any surface refinement at all given this line:
``` if has_surface_model && iref==3 && sim.spacing_surface_refinement !== nothing``` and ```refine_surface``` would never be called. This is a question that I had since I wasn't sure if the default behaviour for the surface refinement should be a number or not performing it if the user don't request for it. I went for the second option

File: src/Simulation/Simulation.jl (hunk 16 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: The bounds are mainly practical safeguards rather than physical constraints. The surface refinement is intended to locally refine the grid close to surfaces. In practice:
-> Values > 1 mm do not really make sense for this step, since the normal grid refinement already produces spacings of that order or smaller. In that case the surface refinement would effectively do nothing.
-> Values < 10 µm lead to extremely small cells at the surface, which dramatically increases the number of grid points and slows down the simulation without providing meaningful benefits for the current surface models.

So the (10 µm, 1 mm) range is meant to prevent accidental misuse (e.g. choosing values that either have no effect or explode the grid size). The bounds here are more about keeping the refinement computationally reasonable than about enforcing a physical scale

File: src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl (hunk 3 above)
Comment (fhagemann) [reply src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl]: Would it make sense to work with the extended grid then, to also keep boundary conditions fulfilled?
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/f279b078b3aef2cf7aee3e54f41c34b87559bcb0/src/Axes/DiscreteAxis.jl#L92-L123

Comment (claudiaalvgar) [reply src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl]: This looks like a better approach. I’ll check whether we can use the extended grid here instead

File: src/Simulation/Simulation.jl (hunk 1 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: I implemented this

File: examples/example_config_files/ivc_inactivelayer.yaml (hunk 2 above)
Comment (claudiaalvgar) [reply examples/example_config_files/ivc_inactivelayer.yaml:23]: Done

File: src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl (hunk 4 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl]: I still need to implement this

File: src/PotentialCalculation/Refinement.jl (hunk 6 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/Refinement.jl]: Done

File: src/Simulation/Simulation.jl (hunk 10 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 12 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 14 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 18 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 19 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 22 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 23 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 24 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/PotentialCalculation/Refinement.jl (hunk 7 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/Refinement.jl]: I think ```copy``` is needed here because ```refine_flags``` is modified later in the function. Without it (```refine_flags = surface_intervals```), the input vector would also be modified right?

File: src/PotentialCalculation/Refinement.jl (hunk 8 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/Refinement.jl:198]: Yes, interval is defined by the distance between 2 ticks

File: src/PotentialCalculation/Refinement.jl (hunk 9 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/Refinement.jl:237]: I don’t think so, the loop structure guarantees that all entries are written, this just follows the same pattern as ``` _refine_axis``` function

File: src/Simulation/Simulation.jl (hunk 11 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Done

File: src/Simulation/Simulation.jl (hunk 21 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Yes, the 2D interpolation path is only intended for the cylindrical case when the φ dimension has size 1, so maybe it makes more sense to have ```only2d = is_cyl && size(closed_pot.data, 2) == 1```, I added it

File: src/PotentialCalculation/Refinement.jl (hunk 7 above)
Comment (fhagemann) [reply src/PotentialCalculation/Refinement.jl]: Did you verify that this is true? I would assume that you need `deepcopy` instead of `copy` here, but I might be wrong.

File: src/Simulation/Simulation.jl (hunk 17 above)
Comment (claudiaalvgar) [reply src/Simulation/Simulation.jl]: Lower values would make the simulation very slow and I think for surface current purposes 0.01mm was a safe low threshold but this is always changeable

File: src/PotentialCalculation/Refinement.jl (hunk 7 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/Refinement.jl]: Is ```deepcopy``` needed here? I think since the elements are ```Bool``` they are immutable so just ```copy``` would be enough to keep the input unchanged

File: src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl (hunk 3 above)
Comment (claudiaalvgar) [reply src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl]: ```get_extended_ticks``` is already used in ```ϵ_r``` and all the approaches I tried to fix this issue didn’t work but clamp only affects cases where the index would go out of bounds, so for example the very refined grid of 1e-5m, for all usual cases that previously ran without errors, it changes nothing, so I think it’s a safe fix for very fine grids or edge conditions

**Comment by claudiaalvgar** [comment]
I think I've now implemented all the changes, let me know if there's still something missing/ or to improve

File: src/Simulation/Simulation.jl (hunk 14 above)
Comment (fhagemann) [reply src/Simulation/Simulation.jl]: On my end, it still looks like this is not done (?)
```suggestion
                    min_spacing::NTuple{3,T} = (T(1e-4), T(1e-4), T(1e-4));
```

src/World/World.jl
```diff
@@ -55,18 +55,25 @@ Definition of the finite volume on which a [`Simulation`](@ref) is performed.
 """
 struct World{T <: SSDFloat, N, S} <: AbstractWorld{T, N} 
     intervals::NTuple{N, SSDInterval{T}}
+    spacing_surface_refinement::Union{NTuple{N,T}, Nothing}
```
fhagemann: I think it would be nice to have this type-stable.
Could we replace `nothing` by `(NaN, NaN, NaN)`, such that it's always an `NTuple{N,T}`?

src/World/World.jl
```diff
@@ -55,18 +55,25 @@ Definition of the finite volume on which a [`Simulation`](@ref) is performed.
...
 end
 
-
 function World(T, dict::AbstractDict, input_units::NamedTuple)::World
+    
+    spacing_surface_refinement = if haskey(dict, "spacing_surface_refinement")
+        ntuple(i -> _parse_value(T, dict["spacing_surface_refinement"][i], internal_length_unit), 3)
+    else
+        nothing
+    end
```
fhagemann, line 71: How do we deal with the fact that the second value for a cylindrical world should not be a "length"? Also: replace `nothing` by Tuple of NaNs?

```suggestion
    spacing_surface_refinement = if haskey(dict, "spacing_surface_refinement")
        ntuple(i -> _parse_value(T, dict["spacing_surface_refinement"][i], internal_length_unit), 3)
    else
        ntuple(i -> NaN, 3)
    end
```

src/World/World.jl
```diff
@@ -158,37 +165,37 @@ function get_cartesian_SSDInterval(T, dict::AbstractDict, input_units::NamedTupl
 end
 
 
-function CylindricalWorld(T, dict::AbstractDict, input_units::NamedTuple)::World
+function CylindricalWorld(T, dict::AbstractDict, input_units::NamedTuple, spacing_surface_refinement)::World
```
fhagemann: Here (and elsewhere): can you define the type of `spacing_surface_refinement`?

src/Simulation/Simulation.jl
fhagemann: I think it would be nice to make this depend on the dimensions of the world.
I wouldn't be surprised if smaller/larger worlds would be better off with smaller/larger values.

src/Simulation/Simulation.jl
fhagemann: See previous comment

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl
fhagemann: Just by looking at the code, I don't fully understand this?
Can you clarify in the most simple terms why we need this clamping? When refining surface intervals close to the boundaries of the world? If so, why don't we care about the boundary conditions applied to the boundaries of the world?

src/World/World.jl
claudiaalvgar: I had introduced the tab indentation and in my local code appeared correct, I added spaces now to see if that worked (?) and I also introduced ```spacing_surface_refinement``` just as an ```NTuple``` so this should be done now

src/World/World.jl
claudiaalvgar: The second value for the cylindrical case is never used for surface refinement. See:
```
new_axes = ntuple(i -> begin
                          if is_cyl && i == 2
                              # Keep φ-axis type but close it for interpolation                                                                                     
                              _get_closed_axis(old_grid.axes[i])
                          else
                              _refine_axis_surface( old_grid.axes[i], surface_intervals[i], min_spacing[i],)
                          end
                      end, 3)

```
in ```src/Simulation/Simulation.jl```

src/World/World.jl
claudiaalvgar: Done

claudiaalvgar:
The ```_safe``` indices in ```get_ϵ_of_oktant``` functions I needed to introduce to prevent out-of-bounds access. For very small surface refinements, the code probes points slightly beyond the limits, so some indices like in1, in2, in3 can fall outside the valid range, I don’t know why the code behaves like this but I tried other solutions and none of them worked.
Clamping just forces those indices back inside the grid so we can safely read ϵᵣ. The issue only appears for small refinements because for the rest of refinements the boundaries are within limits, you can reproduce this issue if you want by running for ```spacing_surface_refinement: [1e-5, 1e-5, 1e-5]``` in ```ivc_inactivelayer.yaml``` with the default ```get_ϵ_of_oktant``` functions (it’ll give a SegFault after approximately an hour since calculating the electric potential is slow for very small grids). I’m open for other solutions I just couldn’t find any other better given the logic of the code

src/World/World.jl
fhagemann: Yes, I'm just saying that the current implementation would through an error if the second argument of surface_refinement in the config_file would be anything other than a number of a length quantity.

Specifically, it someone were to pass `[1e-4u"m", 0.1u"°", 1e-4u"m"]` as `surface_refinement`, this would throw an error because `0.1u"°"` is not a length.

src/World/World.jl
```diff
@@ -55,18 +55,25 @@ Definition of the finite volume on which a [`Simulation`](@ref) is performed.
 """
 struct World{T <: SSDFloat, N, S} <: AbstractWorld{T, N} 
     intervals::NTuple{N, SSDInterval{T}}
+    spacing_surface_refinement::NTuple{N, T}
 end
 
 function World{T, N, S}(args...) where {T <: SSDFloat, N, S} 
     return World{T, N, S}(args)
 end
```
fhagemann: If you add another method here, you don't need to define the default tuple later for EVERY method:
```julia
function World{T, N, S}(args...; surface_refinement_limits::NTuple{N,T} = ntuple(i -> T(NaN), N)) where {T <: SSDFloat, N, S} 
    return World{T, N, S}(args, surface_refinement_limits)
end

File: src/World/World.jl [hunk 29]
```diff
@@ -157,38 +164,37 @@ function get_cartesian_SSDInterval(T, dict::AbstractDict, input_units::NamedTupl
...
 
-function CylindricalWorld(r_max::T, zl::T, zr::T)::World where {T <: SSDFloat}
+function CylindricalWorld(r_max::T, zl::T, zr::T; spacing_surface_refinement::NTuple{3,T} = ntuple(i->T(NaN),3))::World where {T <: SSDFloat}
     r_int = SSDInterval{T, :closed, :closed, :r0, :infinite}(T(0), abs(r_max * T(1.1)))
     φ_int = SSDInterval{T, :closed, :open, :periodic, :periodic}(T(0), T(2π))
     Δz::T = zr - zl
     z_int = SSDInterval{T, :closed, :closed, :infinite, :infinite}(zl - (Δz / 10), zr + (Δz / 10))
-    return World{T, 3, Cylindrical}( r_int, φ_int, z_int )
+    return World{T, 3, Cylindrical}( (r_int, φ_int, z_int), spacing_surface_refinement )
 end
```
fhagemann: Then, this can just become:
```suggestion
function CartesianWorld(xl::T, xr::T, yl::T, yr::T, zl::T, zr::T; kwargs...)::World where {T <: SSDFloat}
Δx::T = (xr - xl) / 10
Δy::T = (yr - yl) / 10
Δz::T = (zr - zl) / 10
x_int = SSDInterval{T, :closed, :closed, :infinite, :infinite}(xl - Δx, xr + Δx)
y_int = SSDInterval{T, :closed, :closed, :infinite, :infinite}(yl - Δy, yr + Δy)
z_int = SSDInterval{T, :closed, :closed, :infinite, :infinite}(zl - Δz, zr + Δz)
    return World{T, 3, Cartesian}( x_int, y_int, z_int; kwargs... )
end

function CylindricalWorld(r_max::T, zl::T, zr::T; kwargs...)::World where {T <: SSDFloat}
r_int = SSDInterval{T, :closed, :closed, :r0, :infinite}(T(0), abs(r_max * T(1.1)))
φ_int = SSDInterval{T, :closed, :open, :periodic, :periodic}(T(0), T(2π))
Δz::T = zr - zl
z_int = SSDInterval{T, :closed, :closed, :infinite, :infinite}(zl - (Δz / 10), zr + (Δz / 10))
    return World{T, 3, Cylindrical}( r_int, φ_int, z_int; kwargs... )
end
```
(I believe then you could also get rid of the brackets around the intervals again)

examples/example_config_files/ivc_inactivelayer.yaml
fhagemann: I think this is also a style question, but we might also think about if it would make sense to also put other simulation-related quantities here, like `refinement_limits`, `max/min_tick_distance` etc.

Not in this PR, but in general, moving towards having all simulation parameters also STORED somewhere in `sim` or `sim.world` for later reference.

fhagemann:
> The `_safe` indices in `get_ϵ_of_oktant` functions I needed to introduce to prevent out-of-bounds access. For very small surface refinements, the code probes points slightly beyond the limits, so some indices like in1, in2, in3 can fall outside the valid range, I don’t know why the code behaves like this but I tried other solutions and none of them worked. Clamping just forces those indices back inside the grid so we can safely read ϵᵣ. The issue only appears for small refinements because for the rest of refinements the boundaries are within limits, you can reproduce this issue if you want by running for `spacing_surface_refinement: [1e-5, 1e-5, 1e-5]` in `ivc_inactivelayer.yaml` with the default `get_ϵ_of_oktant` functions (it’ll give a SegFault after approximately an hour since calculating the electric potential is slow for very small grids). I’m open for other solutions I just couldn’t find any other better given the logic of the code

And this is because, instead of expecting maximum 1 tick going "outside" of the grid, you might have more than 1 now?

src/Simulation/Simulation.jl
fhagemann: Still missing..

examples/example_config_files/ivc_inactivelayer.yaml
claudiaalvgar: Yes, I agree this would be good for a new PR

src/World/World.jl
claudiaalvgar: Do we need to allow for passing angles even though they are not going to be used at all for surface refinement?

src/World/World.jl
claudiaalvgar: This should be done now

claudiaalvgar:
> And this is because, instead of expecting maximum 1 tick going "outside" of the grid, you might have more than 1 now?
>

The problem is that, from what I’ve observed, the indices used to access ```ϵ_r``` sometimes go one unit beyond their expected bounds. Specifically, for cylindrical coordinates, ```in1 > size(ϵ_r, 3), in2 >size(ϵ_r, 2),``` or ```in3 >size(ϵ_r, 1)``` and for Cartesian coordinates,``` in1 > size(ϵ_r, 1),  in2 > size(ϵ_r, 2)``` or ```in3 > size(ϵ_r, 3)```. That’s why the clamps in ```get_ϵ_of_oktant``` are necessary to avoid these ```out-of-bounds``` errors. I tried to address this issue in other ways in other functions e.g. ```calculate_weights_for_innerloop``` and ```calculate_sor_weights``` but they didn't work as well

fhagemann:
I just want to make sure that we expect accessing `ϵ_r` out of bounds of not..
Sure, this might fix the fact that we will not get an error, but maybe we shouldn't exceed the bounds of `ϵ_r` at all in the first place :man_shrugging:

src/Simulation/Simulation.jl
```diff
@@ -813,6 +815,125 @@ function refine!(sim::Simulation{T}, ::Type{WeightingPotential}, contact_id::Int
...
+    end, 3)
+
+    # Refine axes: skip phi-axis if Cylindrical
+    is_cyl = CS === Cylindrical
+    new_axes = ntuple(i -> begin
+                          if is_cyl && i == 2
+                              # Keep φ-axis type but close it for interpolation
+                              _get_closed_axis(old_grid.axes[i])
+                          else
+                              _refine_axis_surface( old_grid.axes[i], surface_intervals[i], min_spacing[i],)
```
fhagemann: Right now, there is no possibility to pass `extra_before` and `extra_After` to `_refine_axis_surface` when calling `refine_surface!`.
Is this intended?

fhagemann:
Do you have an example (some config file + some simulation settings when running `calculate_electric_potential!` that would cause the calculation to fail when not including these "safe" `clamps` for `ϵ_r`.
I might want to have a short look at that myself to see what's going on :)

src/Simulation/Simulation.jl
```diff
@@ -1019,6 +1159,23 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
                                                 paint_contacts = paint_contacts,
                                                 sor_consts = is_last_ref ? T(1) : sor_consts,
                                                 verbose = verbose )
+                
+                if has_surface_model && iref==3 && !any(isnan, sim.world.spacing_surface_refinement)
```
fhagemann: Quick question while looking at this: why use `iref==3`?
I would assume we might want to use `is_last_ref` defined in line 1138 (?)

fhagemann:
We tracked down the out-of-bounds error when accessing `ϵ` to the grid not having an even number of grid ticks.
I recall that either all, or just one axis needs to have an even number of ticks for the red-black division of the grid to work. Having an even number of ticks resolves this issue.

Original code:
Screenshot from 2026-04-15 07-35-00

Forcing the `z`-axis to have an even number of ticks:
Screenshot from 2026-04-15 07-35-56


I can look into implementing a fix for this later today or this week.

src/PotentialCalculation/Refinement.jl
```diff
@@ -226,7 +226,9 @@ function _refine_axis_surface( ax::DiscreteAxis{T, <:Any, <:Any, ClosedInterval{
         for i in r
             Δ = old_ticks[i+1] - old_ticks[i]
             if Δ > min_spacing
-                ns[i] = ceil(Int, Δ/min_spacing) - 1
+                _ns = ceil(Int, Δ/min_spacing)
+                # always add an even number of ticks
+                ns[i] = isodd(_ns) ? _ns - 1 : _ns
```
fhagemann: @claudiaalvgar Do you see any problem with this?
It might end up giving us slightly more ticks, and they might be slightly less than `min_spacing` apart. If that's an issue, we could change line 229
```suggestion
                _ns = floor(Int, Δ/min_spacing)
                # always add an even number of ticks
                ns[i] = isodd(_ns) ? _ns - 1 : _ns
```

claudiaalvgar: The main problem I see is that this might double number of grid ticks needed and hence, the computing time since this is done per grid interval being refined right? Is there not a way of checking that the overall grid is even after the refinement?

fhagemann: I could do that, but then the question is: where do we add the extra tick to make it even if the axis is odd?

claudiaalvgar: I don't know how feasible is this but maybe we could add an extra tick in the biggest grid interval or at the last interval of the grid (outside the surface) but I'm not so sure about this, I'll have to test it

fhagemann: I can also think about this

fhagemann: I updated this now: f43290c

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl
```diff
@@ -10,14 +10,13 @@
...
 ) where {T, S}
-    
 
     # pww1r        = pcs.geom_weights[3][1, in1]
     pww1r        = geom_weights_3[1, in1]
     pww1l        = geom_weights_3[2, in1]
     pwΔmp1       = geom_weights_3[3, in1]
     Δ1_ext_inv_l = geom_weights_3[4, in1]
-    Δ1_ext_inv_r = geom_weights_3[4, in1 + 1]
+    Δ1_ext_inv_r = in1 < size(geom_weights_3, 2) ? geom_weights_3[4, in1 + 1] : geom_weights_3[4, in1]
```
fhagemann: Is this still a leftover of the clamping?

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,85 @@ end
...
+                    tmp = (Δ % 1, i)
+                end
+            end
+        end
+    end
+
+    # Ensure that the number of ticks is even
+    if isodd(sum(ns))
+        ns[tmp[2]] -= 1
+    end
```
fhagemann, line 223: What's happening here:
`Δ` denotes the difference between two neighboring ticks in units of `min_spacing`.

In case that the number of added grid points should be odd, we might want to reduce `1` tick from the interval where we can most easily remove one.

How do we define that: 
If `Δ % 1` is large, e.g. `0.8` for `Δ = 3.8`, that means that we are not quite able to add 4 ticks, so we are adding 3. Here, we don't want to add only 2.
If `Δ % 1` is small, e.g. `0.1` for `Δ = 4.1`, that means that we are comfortably adding 4 ticks, but it would also be fine "just" adding 3.

So, with this routine, I am selecting the index of `ns` with the lowest value of `Δ % 1` that fulfills `Δ > 1`, which is the interval where we can most easily remove one tick.

src/PotentialCalculation/Refinement.jl
```diff
@@ -174,3 +174,85 @@ end
...
+
+    # Compute number of points to add per interval
+    ns = zeros(Int, n_int)
+    tmp = (1.0, -1)
+    for r in merged
+        for i in r
+            Δ = (old_ticks[i+1] - old_ticks[i])/min_spacing
+            if Δ > 1
+                ns[i] = ceil(Int, Δ) - 1
```
fhagemann, line 211: Also, I copied this from the original code in this PR, but I want to argue that this should be `floor` rather than `ceil` (and then also `Δ >= 2` instead of `Δ > 1`, which makes sense, because we only want to add a tick if the two old ticks are at least **TWO** `min_spacing` apart, such that the new tick can have a distance of `min_spacing` to each of them).

```suggestion
                ns[i] = floor(Int, Δ) - 1
```

When using `ceil`, the resulting intervals can end up being smaller than `min_spacing`:

Image

src/Simulation/Simulation.jl
```diff
@@ -1019,6 +1158,26 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
                                                 sor_consts = is_last_ref ? T(1) : sor_consts,
                                                 verbose = verbose )
+                
+                if has_surface_model && iref == 3 && !any(isnan, sim.world.spacing_surface_refinement)
+                    # perform surface refinement
+                    mark_bulk_bits!(sim.point_types.data)
+                    mark_undep_bits!(sim.point_types.data, sim.imp_scale.data)
+                    mark_inactivelayer_bits!(sim.point_types.data)
+                    # Maximum spacing between (refined) surface ticks (if min_spacing = 1e-4m, only surface intervals wider than 0.1mm get new ticks)
+                    verbose && println("Refining the grid close to the surface")
```
claudiaalvgar, line 1162: Just a comment: here I had the print statement after the surface refinement is finished since it looked like the surface refinement takes forever when in reality, is the electric potential calculation on refined surfaces that takes longer, so it could be misleading

claudiaalvgar:
The new changes look good to me and the code behaves as intended so I think this is good to go on my side

src/Simulation/Simulation.jl
fhagemann: That’s why I added two print statements: one before surface refinement, and one before the electric potential calculation

src/PotentialCalculation/Refinement.jl
fhagemann: @claudiaalvgar what is your take on this?

claudiaalvgar: But min_spacing is used as the maximum allowed spacing so while it’s true that ceil can lead to spacings smaller than min_spacing, this is expected since we are enforcing an upper bound on the spacing. Using floor would in general allow intervals larger than min_spacing, for example in this case it would be 5/6 = 0.833 > 0.8

fhagemann: But then the name min_spacing is a bit misleading, if it’s more of a max_spacing

claudiaalvgar: Yes, I think it would be better to change the name since in the end it is an upper bound

examples/example_config_files/ivc_inactivelayer.yaml
fhagemann: —> #600

src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl
fhagemann: Fixed by ensuring the grid axes have an even number of ticks

src/PotentialCalculation/Refinement.jl
fhagemann: This should be changed to add a tick rather than removing one —> I will work on this

fhagemann:
Okay, I will deal with min/max distance and also look into performance, but will merge this afterwards!

src/PotentialCalculation/Refinement.jl
```diff
@@ -179,23 +179,20 @@ _extend_refinement_limits(rl::Tuple{<:Real,<:Real,<:Real}) = rl
...
     # Find first and last surface intervals
-    first_surface = findfirst(surface_intervals)
-    last_surface  = findlast(surface_intervals)
+    first_surface = findfirst(refine_flags)
+    last_surface  = findlast(refine_flags)
 
     (isnothing(first_surface) || isnothing(last_surface)) && return ax
 
-    # Create flags for all intervals to refine
-    refine_flags = copy(surface_intervals)
```
fhagemann: `surface_intervals` are not used anymore after calling this functions, so we should be fine modifying this array instead of allocating more memory.

src/PotentialCalculation/Refinement.jl
```diff
@@ -204,27 +201,11 @@ function _refine_axis_surface( ax::DiscreteAxis{T, <:Any, <:Any, ClosedInterval{
...
-        end
-    end
-
     # Compute number of points to add per interval
     ns = zeros(Int, n_int)
     tmp = (zero(T), 1)
-    for r in merged
-        for i in r
+    for i in eachindex(refine_flags)
+        if refine_flags[i]
```
fhagemann: Instead of creating a `merged` array of unit ranges, we can merge those two `for`-loops into one and just perform interval refinement for the `i` that have `refine_flags` set.

src/PotentialCalculation/Refinement.jl
```diff
@@ -242,14 +223,14 @@ function _refine_axis_surface( ax::DiscreteAxis{T, <:Any, <:Any, ClosedInterval{
...
-    sub_widths = [(old_ticks[i+1] - old_ticks[i]) / (ns[i]+1) for i in 1:n_int]
     ticks = Vector{T}(undef, length(old_ticks) + sum(ns))
     i_tick = 1
     for j in 1:n_int
         ticks[i_tick] = old_ticks[j]
+        sub_width = (old_ticks[j+1] - old_ticks[j]) / (ns[j]+1)
         for k in 1:ns[j]
             i_tick += 1
-            ticks[i_tick] = old_ticks[j] + k*sub_widths[j]
+            ticks[i_tick] = old_ticks[j] + k*sub_width
```
fhagemann: No need to allocate memory for `sub_widths` if the values can just be calculated "on-the-fly" in the `for`-loop

src/Simulation/Simulation.jl
```diff
@@ -845,24 +845,19 @@ function refine_surface!(sim::Simulation{T,CS}, max_spacing::NTuple{3,T} = (T(1e
...
+        if val < min_allowed || val > max_allowed
+            @warn "max_spacing[$d] = $val m is out of bounds (min_allowed = $min_allowed, max_allowed = $max_allowed)."
         end
-    end
-    
-    max_spacing = ntuple(d -> length(old_grid.axes[d].ticks) == 1 ? T(0.0) : max_spacing[d], 3)
+        clamp(val, min_allowed, max_allowed)
+    else 
+        zero(T)
+    end, 3)
```
fhagemann, line 860: This avoids allocations by preventing creating a `Vector` for `max_spacing`, but letting it remain it a `Tuple` throughout the function.

fhagemann (approved):

claudiaalvgar:
Final comment: I tested the code with new implementations on CPU and GPU and the grid results are as expected, I leave them here for future reference:
* max_spacing = 1e-4 m IVCIlayer on GPU: Elapsed time: 47.767041562 seconds GRID: Grid{Float64, 3, Cylindrical}(0.0 .. 0.04 - length = 348, 0.0 .. 0.0 - length = 1, -0.01 .. 0.09 - length = 902)
ICVIlayer_1e-4_GPU_DeleteClampChanges
* max_spacing = 2.5e-5 m IVCIlayer on GPU: Elapsed time: 80.683083286 seconds GRID: Grid{Float64, 3, Cylindrical}(0.0 .. 0.04 - length = 1292, 0.0 .. 0.0 - length = 1, -0.01 .. 0.09 - length = 3438)
ICVIlayer_2 5e-5_GPU_DeleteClampChanges
* max_spacing = 1e-5 m IVCIlayer on GPU: Elapsed time: 275.473297317 seconds GRID: Grid{Float64, 3, Cylindrical}(0.0 .. 0.04 - length = 3152, 0.0 .. 0.0 - length = 1, -0.01 .. 0.09 - length = 8520)
ICVIlayer_1e-5_GPU_DeleteClampChanges

* max_spacing = 1e-4 m IVCIlayer on CPU Threads= 45: Elapsed time: 28.721281699 seconds GRID: Grid{Float64, 3, Cylindrical}(0.0 .. 0.04 - length = 348, 0.0 .. 0.0 - length = 1, -0.01 .. 0.09 - length = 902)

* max_spacing = 2.5e-5 m IVCIlayer on CPU Threads= 45: Elapsed time: 190.124053047 seconds GRID: Grid{Float64, 3, Cylindrical}(0.0 .. 0.04 - length = 1292, 0.0 .. 0.0 - length = 1, -0.01 .. 0.09 - length = 3438)

* max_spacing = 1e-5 m IVCIlayer on CPU Threads= 45: Elapsed time: 1213.737185377 seconds GRID: Grid{Float64, 3, Cylindrical}(0.0 .. 0.04 - length = 3152, 0.0 .. 0.0 - length = 1, -0.01 .. 0.09 - length = 8520)

PR #597, opened 2026-03-20
Save `depletion_handling` to `PointTypes` and improve warnings

hervasa2:
### This PR addresses:

- The redundant overshoot warning and not useful in most simulations when `depletion_handling = true` is used
- Addresses `is_depleted` and `get_active_volume` incorrect results when `depletion_handling = false` was set during simulation.
- Passes verbose correctly down the simulation chain

### Redundant overshoot warning
For depleted detectors the electric potential (EP) is known to be monotonic, thus any absolute overshoot could indicate the detector is undepleted. Thus far this check has only been  done when `depletion_handling = true`. I ran a test of this check for a depleted and an undepleted detector. 
```julia
sim.detector = SolidStateDetector(sim.detector, contact_id = 2, contact_potential = 3000)
calculate_electric_potential!(sim,refinement_limits = [0.2, 0.1, 0.05], depletion_handling = true)
```
```julia
┌ Warning: Detector seems to not be fully depleted at a bias voltage of 3000.0 V.
│ At least one grid point has a higher potential value (3000.0007 V)
│ than the maximum applied potential (3000.0 V). This should not be.
│ However, small overshoots could be due to numerical precision.
└ @ SolidStateDetectors ~/SolidStateDetectors.jl/src/Simulation/Simulation.jl:1047
```
```julia
sim.detector = SolidStateDetector(sim.detector, contact_id = 2, contact_potential = 500)
calculate_electric_potential!(sim,refinement_limits = [0.2, 0.1, 0.05], depletion_handling = true)
```
```julia
┌ Warning: Detector seems to not be fully depleted at a bias voltage of 500.0 V.
│ At least one grid point has a higher potential value (500.00003 V)
│ than the maximum applied potential (500.0 V). This should not be.
│ However, small overshoots could be due to numerical precision.
└ @ SolidStateDetectors ~/SolidStateDetectors.jl/src/Simulation/Simulation.jl:1047
```
This is a counterexample to this test as was implemented. There are very small overshoots in both cases, which cannot be used to check for depletion. Any check for a monotonous EP should be much more robust, for example by checking if the EP extrema is local and checking for . This could be done in a future PR.

I suggest for now to remove this test entirely. Users should use `depletion_handling = true` and use the `is_depleted` to check for depletion accurately.

### `is_depleted` and `get_active_volume` incorrect results
When `depletion_handling = false` is set `is_depleted` always returns `true`  and `get_active_volume` always returns the full detector volume. _This is in general incorrect_. I added the `depletion_handling_bit` to `point_types` to address this. Now point types store the information of depletion handling. 
When `depletion_handling = false` was used, a warning now always appears in `is_depleted` and `get_active_volume`. Note that the output of these functions has not changed.


src/ScalarPotentials/PointTypes.jl
```diff
@@ -110,9 +112,14 @@ depleted at the provided bias voltage.
...
 """
-is_depleted(point_types::PointTypes)::Bool = 
+function is_depleted(point_types::PointTypes)
+    if depletion_handling_bit & point_types[1] == 0
+        @warn """Electric potential calculation was not run with depletion handling enabled. The result of is_depleted may be inaccurate. 
+        Consider running the electric potential calculation with depletion_handling = true for accurate results."""
+    end
     !any(b -> (bulk_bit & b > 0) && (undepleted_bit & b > 0) &&
     (inactive_layer_bit & b == 0), point_types.data)
+end
```
fhagemann: Maybe that's just me, but I feel like using one bit in the `PointTypes` just for this seems a bit like a waste/overkill, especially if we only care about the bit in the first data point and because we apply depletion handling for all grid points, and not just a sub-selection.

How about adding this as parametric type to `PointTypes`?
```julia
struct PointTypes{T, 3, CS, D} # where D being depletion_handling (true / false)
```
Then we could dispatch the functions to throw these warnings on this `D`.

src/Simulation/Simulation.jl
```diff
@@ -1014,7 +1015,8 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
                                                 device_array_type = device_array_type,
                                                 not_only_paint_contacts = not_only_paint_contacts, 
                                                 paint_contacts = paint_contacts,
-                                                sor_consts = is_last_ref ? T(1) : sor_consts )
+                                                sor_consts = is_last_ref ? T(1) : sor_consts,
+                                                verbose = verbose )
```
fhagemann, line 1021: This was just missing?

src/Simulation/Simulation.jl
```diff
@@ -1028,27 +1030,8 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
-                    than the minimum applied potential ($(minimum_applied_potential) V). This should not be.
-                    However, small overshoots could be due to numerical precision."""
-                break
-            end
-            if sim.electric_potential.data[i] > maximum_applied_potential # n-type
-                @warn """Detector seems not to be not fully depleted at a bias voltage of $(bias_voltage) V.
-                    At least one grid point has a higher potential value ($(sim.electric_potential.data[i]) V)
-                    than the maximum applied potential ($(maximum_applied_potential) V). This should not be.
-                    However, small overshoots could be due to numerical precision."""
-                break
```
fhagemann, line 1051: What about still keeping this, but adding a larger tolerance to not warn about small overshoots? This was initially introduced to warn in case that the maximum number of iterations was not enough to capture the depletion behavior of the detector well?

I'm not a fan of removing this without having a clear plan on how we want to have this replaced? Or is someone working on a monotony check?

fhagemann (commented):
I agree with the point, that these overshoot warnings could be improved upon, but I am not a fan of dedicating a bit in the `PointTypes` just to be able to throw a warning..

src/Simulation/Simulation.jl
hervasa2: It was missing yes.

src/ScalarPotentials/PointTypes.jl
hervasa2: I agree it's overkill, but it's the easiest non-breaking solution I could think of. And also the warning is non-trivial I would say. I agree your solution is best if we can live with any (if any) resulting breaking changes.

src/Simulation/Simulation.jl
hervasa2: Then I think the message should be different, steering the user towards increasing the maximum number of iterations. In any case we could add this back with some tolerance. I think `5e7*csg_default_tol(T)` would be a good value based on my tests. I will think if we can make an easy but more complete monotony check.

fhagemann: `5e7*csg_default_tol(T)` is huge though and not just due to numerical precision..

hervasa2: I know its not, but these are common overshoots, regardless if the detector is depleted or not.

fhagemann: `0.5` for `Float32` (where `csg_default_tol(Float32) == 1.0f-8`)?

hervasa2: The 5e7 was determined for the Float64 -> ends up quite large for Float32. In general overshoots are only 10 times higher for Float32 than for Float64, whereas `csg_default_tol(Float64)/csg_default_tol(Float32) = 10000`. I haven't done a comprehensive test, but it would seem that `tol = 0.1` and `tol = 0.01` would be good values for Float32 and Float64 respectively, rather than trying to tie it to `csg_default_tol(T)`.

fhagemann: And is this the **absolute** tolerance for a detector with ~3000V (so `1.0f-4` (Float32) or `1e-5` (Float64) relative to its depletion voltage)?

hervasa2: `tol = 0.1u"V"` and `tol = 0.01u"V"` are absolute tolerances

src/ScalarPotentials/PointTypes.jl
```diff
@@ -112,15 +113,17 @@ depleted at the provided bias voltage.
 is_depleted(sim.point_types)
 ```
 """
-function is_depleted(point_types::PointTypes)
-    if depletion_handling_bit & point_types[1] == 0
-        @warn """Electric potential calculation was not run with depletion handling enabled. The result of is_depleted may be inaccurate. 
-        Consider running the electric potential calculation with depletion_handling = true for accurate results."""
-    end
+function is_depleted(point_types::PointTypes{T, N, S, AT, true}) where {T, N, S, AT}
```
fhagemann: If you're not using `T`, `N`, `S`, `AT` in the function, this could just be
```julia
::PointTypes{<:Any, <:Any, <:Any, <:Any, true}
```
?

fhagemann (commented):
Thanks!

Some comments to keep compatibility with old simulations and reading/writing to HDF5:
- If we read a simulation from an HDF5 file, will `D` default to `true` or `false`? Probably the most elegant solution would be to set it to `false` if all grid points are marked
depleted, and to `true` if there is any undepleted grid point.
- How will the information about `depletion_handling` be written to the HDF5 file? We might want to add a new field to the `NamedTuple` when creating it, and some method to recover it from that `NamedTuple`.
- If it was simulated with `depletion_handling = true` but had no undepleted grid points, there should be a function (something like `set_point_type_depletion_handling!(sim)`, that would flip this parametric type for an old simulation. We could include this in the warning (" if the simulations were run with depletion_handling = true, please update the point_types using ...").
- We should then also check, if simulation files created with the updates in this PR and written to HDF5 files can still be read by older versions of SSD without problems.

hervasa2:
> If we read a simulation from an HDF5 file, will D default to true or false? Probably the most elegant solution would be to set it to false if all grid points are marked
depleted, and to true if there is any undepleted grid point.

I agree. See below.

> How will the information about depletion_handling be written to the HDF5 file? We might want to add a new field to the NamedTuple when creating it, and some method to recover it from that NamedTuple.

This is already done, see: 

https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/a17432b61eb3c4c4410950e940ee490fcebe0514/src/ScalarPotentials/PointTypes.jl#L238-L254

(I will take care of shitching to the `<:Any`'s here too) However this will beak when loading an HDF5 file from an older version of SSD. I suggest the following which will also take care of your first comment:
```julia
function PointTypes(nt::NamedTuple)
    grid = Grid(nt.grid)
    T = typeof(grid.axes[1].ticks[1])
    S = get_coordinate_system(grid)
    N = get_number_of_dimensions(grid)
    D = if hasproperty(nt, :depletion_handling)
        nt.depletion_handling
    else
        any(b -> (bulk_bit & b > 0) && (undepleted_bit & b > 0) &&
    (inactive_layer_bit & b == 0), nt.values)
    end
    PointTypes{T, N, S, typeof(grid.axes), D}( nt.values, grid )
end
```

> If it was simulated with depletion_handling = true but had no undepleted grid points, there should be a function (something like set_point_type_depletion_handling!(sim), that would flip this parametric type for an old simulation. We could include this in the warning (" if the simulations were run with depletion_handling = true, please update the point_types using ...").

Probably something like 
```julia
function set_point_type_depletion_handling!(sim::Simulation, depletion_handling::Bool)
    sim.point_types = PointTypes(sim.point_types.data, sim.point_types.grid, depletion_handling)
    nothing
end
```
Should then this live in `src/Simulation/Simulation.jl`?

> We should then also check, if simulation files created with the updates in this PR and written to HDF5 files can still be read by older versions of SSD without problems.

Could you help me out with testing a file (created with the code in this PR -- I will send it separately) with an older ssd version?

src/ScalarPotentials/PointTypes.jl
```diff
@@ -73,6 +76,7 @@ function in(pt::AbstractCoordinatePoint{T}, point_types::PointTypes{T, 3, S})::B
     return point_type & bulk_bit > 0
 end
 
+has_depletion_handling(::PointTypes{<:Any,<:Any,<:Any,<:Any,D}) where {D} = D
```
fhagemann, line 79: I added a new function called `has_depletion_handling` that can retrieve the information stored in `D` quickly.

fhagemann:
I think I have added the missing things based on your last comment. Please have a look!
Do you plan / have time to add tests for running a simulation with/without depletion_handling and then checking if the warning is thrown correctly?

src/Simulation/Depletion.jl
fhagemann, line 1: I renamed `src/Simulation/DepletionVoltage.jl` to `src/Simulation/Depletion.jl` and put `set_point_type_depletion_handling!` here.

src/ScalarPotentials/ScalarPotential.jl
fhagemann, line 37: One thing that came up when reviewing the change in `ScalarPotential`:
do we actually want this for `PointTypes`. When interpolating `PointTypes` onto another grid, I expect the interpolated results to make no sense..

tdixon97:
Thanks @hervasa2 for working on this, that warning is very annoying, but would'nt a more practical approach just be to only emit the warning with a 1V threshold. In practice, you never really care about small fractions of a volt.

fhagemann:
It makes sense to include some information about the depletion handling in the final `PointTypes` to avoid wrong results in `is_depleted` and `get_active_volume` in cases where `depletion_handling` was set to `false` (which is the faster default)

src/ScalarPotentials/ScalarPotential.jl
hervasa2: True. Is this ever done in the simulation chain?

hervasa2:
Thanks Felix! Looks good to me. I also realized that the monotony check should only be done in the detector. To remedy this, I added the overshoot check back but with a `bulk_bit` check, and changed the warning text. Now the warning is gone for all my tests. I checked and overshoots only seem to happen outside the detector.

src/ScalarPotentials/ScalarPotential.jl
fhagemann: I would have to check carefully to make sure this is never used

src/Simulation/Simulation.jl
```diff
@@ -1037,7 +1037,26 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
...
+            if sim.electric_potential.data[i] > maximum_applied_potential && sim.point_types.data[i] & bulk_bit > 0 # n-type
+                @warn """At least one grid point in the detector has a higher potential value ($(sim.electric_potential.data[i]) V)
+                    than the maximum applied potential ($(maximum_applied_potential) V). 
+                    A fully depleted detector should not have local extrema in a converged electric potential.
+                    However, small overshoots can occur due to numerical precision."""
+                break
+            end
+        end
+    end
     if isEP mark_bulk_bits!(sim.point_types.data) end
```
fhagemann: Correct me if I'm wrong, but this looks to me like you're checking for bulk bits  (lines 1040-1059) BEFORE they are being set (line 1060)?
I would expect this warning to never appear, just because of the fact that the bulk bits are not set yet...

hervasa2: Good catch, meant to put it after. But the same results still hold.

fhagemann: I would be interested to see if you’re still getting no warning then, or if it’s back then.

src/ScalarPotentials/ScalarPotential.jl
hervasa2: Could you please create an issue and we can follow up there?

src/Simulation/Simulation.jl
fhagemann: With the latest changes, are you still seeing no warnings in your test cases?

src/ScalarPotentials/ScalarPotential.jl
fhagemann: --> https://github.com/JuliaPhysics/SolidStateDetectors.jl/issues/599

src/Simulation/Simulation.jl
hervasa2: No warnings in all cases. As I said I had previously checked where all overshoots happen, and they were outside the detector for all my test cases.

src/ScalarPotentials/PointTypes.jl
```diff
@@ -110,9 +114,16 @@ depleted at the provided bias voltage.
...
-    (inactive_layer_bit & b == 0), point_types.data)
+function is_depleted(point_types::PointTypes)::Bool
+    return if has_depletion_handling(point_types)
+        !any(b -> (bulk_bit & b > 0) && (undepleted_bit & b > 0) &&
+        (inactive_layer_bit & b == 0), point_types.data)
+    else 
+        @warn """Electric potential calculation was not run with depletion handling enabled. The result of is_depleted may be inaccurate. 
+        Consider running the electric potential calculation with depletion_handling = true for accurate results."""
+        true
+    end
```
fhagemann, line 128: The warning now claims that the simulation was not run with `depletion_handling == true` even though that might not be the case if it was run with a previous SSD version. We might also want to add some note that if the user knows that the simulation was run with `depletion_handling == true` in previous versions of SSD, that there is a method to update this.
```julia
set_point_type_depletion_handling!(sim, true)
```

fhagemann (commented):
One final thing, but otherwise this should be good to go (except for if you check for local extrema only in points being bulk bits, we could also implement that quick fix for now and wait with this PR until the next minor version 0.12.0, having in mind that there is a slight inconvenience when reading old, fully-depleted simulations, where `depletion_handling` is guessed to be `false`).

src/Simulation/Simulation.jl
fhagemann: Is this comment obsolete, i.e. we don't seem to need some tolerances, as long as we restrict the search for local extrema to the grid points in the bulk?

hervasa2: Yes the need for tolerances and the comment is now obsolete. 

> We don't seem to need some tolerances, as long as we restrict the search for local extrema to the grid points in the bulk

src/ScalarPotentials/PointTypes.jl
hervasa2: How about:

> Electric potential calculation was not run with depletion handling enabled. The result of `is_depleted` may be inaccurate. Consider running the electric potential calculation with `depletion_handling = true` for accurate results. When loading simulations from <v0.11.x, the  depletion handling state is ambiguous if no `undepleted_bit`'s were found, in this case use `set_point_type_depletion_handling!(sim, true) as needed`

hervasa2:
I would say there is no need to split this. But I would not wait until v0.12, could be a while...

src/ScalarPotentials/PointTypes.jl
fhagemann: Sounds good to me

PR #601, opened 2026-04-13
Exclude inactive layer when searching for potential overshoots

fhagemann:
Follow-up of #597: the warning still appears for simulations with small overshoots in the inactive layer.
```julia
using SolidStateDetectors, CUDA
T = Float64
sim = Simulation{T}(SSD_examples[:IVCIlayer]) # setting spacing_surface_refinement = [1e-5, 1e-5, 1e-5]
g = Grid(sim)
apply_initial_state!(sim, ElectricPotential, g)

t = @elapsed begin
    calculate_electric_potential!(
        sim,
        device_array_type = CuArray,
        grid = g,
        verbose = true,
        depletion_handling = true
    )
end
println("Elapsed time: $(t) seconds")
```

Before this PR:
```
┌ Warning: At least one grid point in the detector has a higher potential value (3500.0000000000014 V)
│ than the maximum applied potential (3500.0 V). 
│ A fully depleted detector should not have local extrema in a converged electric potential.
│ However, small overshoots can occur due to numerical precision.
└ @ SolidStateDetectors /volumes/mobius/users/felix/test/SolidStateDetectors.jl/src/Simulation/Simulation.jl:1216
```

After this PR: no warning

### DIFF
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -1048,14 +1048,18 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
             maximum_applied_potential = maximum(broadcast(c -> c.potential, sim.detector.contacts))
             minimum_applied_potential = minimum(broadcast(c -> c.potential, sim.detector.contacts))
             @inbounds for i in eachindex(sim.electric_potential.data)
-                if sim.electric_potential.data[i] < minimum_applied_potential && sim.point_types.data[i] & bulk_bit > 0 # p-type
+                if sim.electric_potential.data[i] < minimum_applied_potential && # p-type
+                    sim.point_types.data[i] & inactive_layer_bit == 0 # exclude grid points in the inactive layer
+                    sim.point_types.data[i] & bulk_bit > 0 # exclude grid points outside of the bulk
                     @warn """At least one grid point in the detector has a smaller potential value ($(sim.electric_potential.data[i]) V)
                         than the minimum applied potential ($(minimum_applied_potential) V). 
                         A fully depleted detector should not have local extrema in a converged electric potential.
                         However, small overshoots can occur due to numerical precision."""
                     break
                 end
-                if sim.electric_potential.data[i] > maximum_applied_potential && sim.point_types.data[i] & bulk_bit > 0 # n-type
+                if sim.electric_potential.data[i] > maximum_applied_potential && # n-type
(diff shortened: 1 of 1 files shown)


fhagemann:
This is done already in `is_depleted`, but we also might want to add that here

fhagemann:
Excluding inactive layer grid points:
```julia
extrema(sim.electric_potential.data[
        (sim.point_types.data .& SolidStateDetectors.bulk_bit .> 0) .&&
        (sim.point_types.data .& SolidStateDetectors.inactive_layer_bit .== 0)])
```
```
(8.601215921073825, 3499.9999979657564)
```

Including inactive layer grid points:
```julia
extrema(sim.electric_potential.data[
        (sim.point_types.data .& SolidStateDetectors.bulk_bit .> 0)])
```
```
(8.601215921073825, 3500.0000000000014) # causing overshoot warning
```

hervasa2 (approved):
Looks great! Thanks for following up

src/Simulation/Simulation.jl
```diff
@@ -1048,14 +1048,18 @@ function _calculate_potential!( sim::Simulation{T, CS}, potential_type::UnionAll
             maximum_applied_potential = maximum(broadcast(c -> c.potential, sim.detector.contacts))
             minimum_applied_potential = minimum(broadcast(c -> c.potential, sim.detector.contacts))
             @inbounds for i in eachindex(sim.electric_potential.data)
-                if sim.electric_potential.data[i] < minimum_applied_potential && sim.point_types.data[i] & bulk_bit > 0 # p-type
+                if sim.electric_potential.data[i] < minimum_applied_potential && # p-type
+                    sim.point_types.data[i] & inactive_layer_bit == 0 # exclude grid points in the inactive layer
+                    sim.point_types.data[i] & bulk_bit > 0 # exclude grid points outside of the bulk
```
fhagemann, line 1053: This code is actually not correct and should be



I pushed the fix directly to `main` --> eb2b33e
```suggestion
                if sim.electric_potential.data[i] < minimum_applied_potential && # p-type
                    sim.point_types.data[i] & inactive_layer_bit == 0 && # exclude grid points in the inactive layer
                    sim.point_types.data[i] & bulk_bit > 0 # exclude grid points outside of the bulk
```

fhagemann:
Releasing this as v0.11.1 now:
https://github.com/JuliaRegistries/General/pull/153265

Issue #603, opened 2026-05-21
Errors with latest `SplitApplyCombine.jl`


tdixon97:
I noticed recently some errors when running SSD simulations of LEGEND detectors. These appeared without changing any code just updating the environment.

Here is a minimal example of the error:

```julia
T = Float64

tdata = legend_test_data_path() 
dets = normpath(joinpath(tdata, "data/legend/metadata/hardware/detectors/germanium/"))

meta = readprops(normpath(dets,"diodes/V99000A.yaml"))
xtal = readprops(normpath(dets,"crystals/V99000.yaml"))

sim = Simulation{T}(
    LegendData,
    meta,
    xtal,
    HPGeEnvironment("LAr", 87.0*u"K"),
    operational_voltage = 4000.0*u"V"
)

```
This gives the following error:
```console
1-element ExceptionStack:
LoadError: MethodError: no method matching haskey(::Float64, ::String)
The function `haskey` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  haskey(::LLVM.ModuleFlagDict, ::String)
   @ LLVM ~/.julia/packages/LLVM/eBGq5/src/core/module.jl:462
  haskey(::LLVM.ModuleFunctionSet, ::String)
   @ LLVM ~/.julia/packages/LLVM/eBGq5/src/core/module.jl:433
  haskey(::LLVM.ExecutionEngineFunctionSet, ::String)
   @ LLVM ~/.julia/packages/LLVM/eBGq5/src/execution.jl:284
  ...

Stacktrace:
  [1] parse_r_of_primitive(::Type{Float32}, dict::OrderedCollections.OrderedDict{String, Any}, unit::Unitful.FreeUnits{(mm,), 𝐋, nothing}, ::Type{SolidStateDetectors.ConstructiveSolidGeometry.Cone})
    @ SolidStateDetectors.ConstructiveSolidGeometry ~/.julia/packages/SolidStateDetectors/Oce2T/src/ConstructiveSolidGeometry/IO.jl:128
[...]

PR #604, opened 2026-05-21
Bug fix in `parse_r_of_primitive`

fhagemann:
Possible fix for #603:
In `parse_r_of_primitive` we used to check for keys of `dictr` without explicitly checking first if `dictr` is a dictionary first.
This worked because of some type piracy in `SplitApplyCombine` that was recently removed.


fhagemann:
Ok, tests seem to fail with the newest version `SplitApplyCombine@1.3.0` but not after this fix, so I think this might fix #603

PR #605, opened 2026-05-28
Allow `ThermalDiffusionLithiumDensity` to be built using multiple contacts

fhagemann:
Addressing issue #595:
People can now pass either a number OR an array of segment contacts to `doped_contact_id` to create `ThermalDiffusionLithiumDensity` or `PNJunctionImpurityDensity`.
Internally, the contact IDs (even if just one) will be stored as `Tuple` and then the contact geometry is defined as a `CSGUnion` of all these geometries.

### DIFF
--- a/src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
+++ b/src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
@@ -16,7 +16,7 @@ function PtypePNJunctionImpurityDensity{T}(
     lithium_annealing_temperature::T,
     lithium_annealing_time::T,
     contact_with_lithium_doped::G,
-    inactive_contact_id::Int,
+    inactive_contact_id::NTuple{<:Any,Int},
     bulk_imp_model::AbstractImpurityDensity{T},
     distance_to_contact::Function = pt::AbstractCoordinatePoint{T} -> ConstructiveSolidGeometry.distance_to_surface(pt, contact_with_lithium_doped)
     ) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}}
@@ -28,8 +28,8 @@ function ImpurityDensity(T::DataType, t::Val{:PtypePNjunction}, dict::AbstractDi
     lithium_annealing_temperature = _parse_value(T, get(dict, "lithium_annealing_temperature", 623u"K"), input_units.temperature)
     lithium_annealing_time = _parse_value(T, get(dict, "lithium_annealing_time", 18u"minute"), internal_time_unit)
     contact_with_lithium_doped = get(dict, "contact_with_lithium_doped", nothing)
-    inactive_contact_id = get(dict, "doped_contact_id", -1)
-    inactive_contact_id < 1 && error("Invalid doped_contact_id: missing or misspelled key")
+    inactive_contact_id = format_contact_id(get(dict, "doped_contact_id", -1))
+    any(i -> i < 1, inactive_contact_id) && throw(ConfigFileError("Invalid doped_contact_id: missing or misspelled key"))
     bulk_imp_model = haskey(dict, "bulk_impurity_density") ? ImpurityDensity(T, dict["bulk_impurity_density"], input_units) : ConstantImpurityDensity{T}(-1e16)
     PtypePNJunctionImpurityDensity{T}(lithium_annealing_temperature, lithium_annealing_time, contact_with_lithium_doped, inactive_contact_id, bulk_imp_model)
 end
--- a/src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
+++ b/src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
@@ -1,22 +1,30 @@
 """
-struct PtypePNJunctionImpurityDensity{T <: SSDFloat} <: AbstractImpurityDensity{T}
+struct PtypePNJunctionImpurityDensity{T <: SSDFloat, N} <: AbstractImpurityDensity{T}
 
 A PN junction impurity model based on lithium thermal diffusion and constant bulk impurity density.
 The surface lithium density is at the saturation level. Ref: [Dai _et al._ (2023)](https://doi.org/10.1016/j.apradiso.2022.110638)
+
+## Parametric types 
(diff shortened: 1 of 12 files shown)


fhagemann:
When simulating the `:BEGe` with `doped_contact_id: [2,3,4]`, I seem to get reasonable results, but we should definitely cross-check that all functionality from the old implementation still works.

image
image

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
claudiaalvgar, line 1: Maybe in this file we could add in the documentation that PtypePNJunctionImpurityDensity struct includes the `inactive_contact_id` field that stores which contacts are doped

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
```diff
@@ -16,7 +16,7 @@ function PtypePNJunctionImpurityDensity{T}(
     lithium_annealing_temperature::T,
     lithium_annealing_time::T,
     contact_with_lithium_doped::G,
-    inactive_contact_id::Int,
+    inactive_contact_id::NTuple{<:Any,Int},
```
claudiaalvgar: Could we also have ```inactive_contact_id::NTuple{N,Int}``` and then ```) where {T <: SSDFloat, G <: Union{<:AbstractGeometry, Nothing}, N}``` ?

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
claudiaalvgar, line 1: Here in the doc I would also update the struct to ```struct ThermalDiffusionLithiumDensity{T <: SSDFloat, N} <: AbstractImpurityDensity{T}``` and maybe write that the type parameter `N` is the number of contacts associated with the lithium-doped surface

src/SolidStateDetector/SolidStateDetector.jl
```diff
@@ -123,31 +127,34 @@ function SolidStateDetector{T}(config_file::AbstractDict, input_units::NamedTupl
...
-            imp_doped_contact_id = config_detector["semiconductor"]["impurity_density"]["doped_contact_id"]
-            if !in(imp_doped_contact_id, getfield.(contacts, :id))
-                ArgumentError("The imp_doped_contact_id is not defined in the configuration file.")
+            if isnothing(config_detector["semiconductor"]["impurity_density"]["doped_contact_id"])
+                throw(ConfigFileError("The imp_doped_contact_id is not defined in the configuration file.")) 
+            end 
+            imp_doped_contact_id = format_contact_id(config_detector["semiconductor"]["impurity_density"]["doped_contact_id"])
+            if !all(i -> in(i, getfield.(contacts, :id)), imp_doped_contact_id)
+                ids = findall(i -> !in(i, getfield.(contacts, :id)), imp_doped_contact_id)
+                throw(ConfigFileError("The imp_doped_contact_ids $(ids) are not defined in the configuration file."))
```
claudiaalvgar: Here I would have this rather than the position: 
```
invalid_ids = filter(i -> !in(i, getfield.(contacts, :id)), collect(imp_doped_contact_id))
throw(ConfigFileError("The imp_doped_contact_ids $(invalid_ids) are not defined in the configuration file."))```

File: src/SolidStateDetector/SolidStateDetector.jl [hunk 3]
```diff
@@ -123,31 +127,34 @@ function SolidStateDetector{T}(config_file::AbstractDict, input_units::NamedTupl
...
             end
         elseif doped_geometry_for_drift && haskey(config_detector["semiconductor"]["charge_drift_model"]["surface_impurity_density"], "doped_contact_id")
-            drift_doped_contact_id = config_detector["semiconductor"]["charge_drift_model"]["surface_impurity_density"]["doped_contact_id"]
-            if !in(drift_doped_contact_id, getfield.(contacts, :id))
-                ArgumentError("The drift_doped_contact_id is not defined in the configuration file.")
+            drift_doped_contact_id = format_contact_id(config_detector["semiconductor"]["charge_drift_model"]["surface_impurity_density"]["doped_contact_id"])
+            if !all(i -> in(i, getfield.(contacts, :id)), drift_doped_contact_id)
+                ids = findall(i -> !in(i, getfield.(contacts, :id)), drift_doped_contact_id)
+                throw(ConfigFileError("The drift_doped_contact_ids $(ids) are not defined in the configuration file."))
```
claudiaalvgar: Same here:
```
invalid_ids = filter(i -> !in(i, getfield.(contacts, :id)), collect(drift_doped_contact_id))
throw(ConfigFileError("The drift_doped_contact_ids $(invalid_ids) are not defined in the configuration file."))
```

claudiaalvgar (commented):
Other than the comments I made, I tested my previous code and it all seems to work fine!

src/SolidStateDetector/SolidStateDetector.jl
fhagemann: That's a very good point, I'm gonna update this!

src/ImpurityDensities/PtypePNJunctionImpurityDensity.jl
fhagemann: Ok, I actually now changed the code such that you can pass either an `Integer` a `Tuple` of `Integers` or a `Vector` of `Integers` as `inactive_contact_id`, and that it will be formatted in the constructor.

Also, the constructor will not require to define `N`, so all previous code should still work (not passing `N` as parametric type, and still being allowed to pass the `inactive_contact_id` as a sole `Integer`).
Only breaking change would be how the `inactive_contact_id` is stored in the field of the struct.

src/ImpurityDensities/ThermalDiffusionLithiumDensity.jl
fhagemann: I've updated the docs, but still kept the "old" constructors.

src/SolidStateDetector/SolidStateDetector.jl
claudiaalvgar: Nice, so this is good to go on my side!

PR #611, opened 2026-06-22
Temperature handling for charge drift and trapping models

claudiaalvgar:
This PR fixes the temperature handling across charge drift and charge trapping models to ensure that model temperatures are inherited consistently from the semiconductor configuration rather than being independently specified 
Updated semiconductor model construction to propagate the detector temperature to:
    * InactiveLayerChargeDriftModel
    * BoggsChargeTrappingModel
    * CombinedChargeTrappingModel
Updated charge trapping tests to construct models using the detector semiconductor temperature.
Removed tests that relied on model-specific temperature values in configuration files.
This change eliminates duplicated temperature definitions across detector models and prevents inconsistencies between semiconductor and trapping/drift temperatures.

### DIFF
--- a/examples/example_config_files/ivc_inactivelayer.yaml
+++ b/examples/example_config_files/ivc_inactivelayer.yaml
@@ -36,7 +36,6 @@ detectors:
         value: -1e10cm^-3
     charge_drift_model:
       model: InactiveLayerChargeDriftModel
-      temperature: 90K
       neutral_impurity_density:
         name: constant
         value: 5.6769e15cm^-3
--- a/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
+++ b/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
@@ -74,7 +74,7 @@ function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
             throw(ConfigFileError(
                 "Temperature mismatch: InactiveLayerChargeDriftModel defines temperature = $(cdm_temperature) K, " *
                 "but the semiconductor temperature is $(temperature) K. " *
-                "Remove `temperature` from the charge drift model and define it only in the semiconductor."
+                "Define `temperature` in the semiconductor and either remove it from the charge drift model or make sure that the temperatures match."
             ))
         end
     end
--- a/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
+++ b/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
@@ -63,11 +63,14 @@ function _calculate_mobility_with_impurities(
 end
 
 function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
-        imp_model::AbstractImpurityDensity, input_units::NamedTuple,
+        imp_model::AbstractImpurityDensity, input_units::NamedTuple, temperature::RealQuantity
     ) where {T <: SSDFloat}
 
-    temperature = _parse_value(T, get(config, "temperature", 90u"K"), input_units.temperature)
+    temperature::T = _parse_value(T, temperature, input_units.temperature)
 
+    if temperature < 50 || temperature > 150
+        @warn "Temperature = $(temperature) K is outside the typical validated range (50–150 K)."
+    end
     neutral_imp_model = if haskey(config, "neutral_impurity_density")
         ImpurityDensity(T, config["neutral_impurity_density"], input_units)
     else
--- a/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
+++ b/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
@@ -66,11 +66,11 @@ function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
         imp_model::AbstractImpurityDensity, input_units::NamedTuple, temperature::RealQuantity
     ) where {T <: SSDFloat}
 
-    temperature::T = _parse_value(T, temperature, input_units.temperature)
+    temp::T = _parse_value(T, temperature, input_units.temperature)
 
     if haskey(config, "temperature")
         cdm_temperature::T = _parse_value(T, config["temperature"], input_units.temperature)
-        if cdm_temperature != temperature
+        if cdm_temperature != temp
             throw(ConfigFileError(
                 "Temperature mismatch: InactiveLayerChargeDriftModel defines temperature = $(cdm_temperature) K, " *
                 "but the semiconductor temperature is $(temperature) K. " *
@@ -79,8 +79,8 @@ function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
         end
     end
 
-    if temperature < 50 || temperature > 150
-        @warn "Temperature = $(temperature) K is outside the typical validated range (50–150 K)."
+    if temp < 50 || temp > 150
+        @warn "Temperature = $(temp) K is outside the typical validated range (50–150 K)."
     end
     neutral_imp_model = if haskey(config, "neutral_impurity_density")
         ImpurityDensity(T, config["neutral_impurity_density"], input_units)
@@ -103,7 +103,7 @@ function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
     else
         throw(ConfigFileError("There is no surface impurity density profile provided."))
     end
-    InactiveLayerChargeDriftModel{T}(temperature, neutral_imp_model, bulk_imp_model, surface_imp_model)
+    InactiveLayerChargeDriftModel{T}(temp, neutral_imp_model, bulk_imp_model, surface_imp_model)
 end
 
 @fastmath function getVe(fv::SVector{3, T}, cdm::InactiveLayerChargeDriftModel{T}, current_pos::CartesianPoint{T} = zero(CartesianPoint{T})) where {T <: SSDFloat}
--- a/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
+++ b/src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
@@ -68,6 +68,17 @@ function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
 
     temperature::T = _parse_value(T, temperature, input_units.temperature)
 
+    if haskey(config, "temperature")
+        cdm_temperature::T = _parse_value(T, config["temperature"], input_units.temperature)
+        if cdm_temperature != temperature
+            throw(ConfigFileError(
+                "Temperature mismatch: InactiveLayerChargeDriftModel defines temperature = $(cdm_temperature) K, " *
+                "but the semiconductor temperature is $(temperature) K. " *
+                "Remove `temperature` from the charge drift model and define it only in the semiconductor."
+            ))
+        end
+    end
+
     if temperature < 50 || temperature > 150
         @warn "Temperature = $(temperature) K is outside the typical validated range (50–150 K)."
     end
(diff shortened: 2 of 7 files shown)


src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
```diff
@@ -63,11 +63,14 @@ function _calculate_mobility_with_impurities(
...
-        imp_model::AbstractImpurityDensity, input_units::NamedTuple,
+        imp_model::AbstractImpurityDensity, input_units::NamedTuple, temperature::RealQuantity
     ) where {T <: SSDFloat}
 
-    temperature = _parse_value(T, get(config, "temperature", 90u"K"), input_units.temperature)
+    temperature::T = _parse_value(T, temperature, input_units.temperature)
 
+    if temperature < 50 || temperature > 150
+        @warn "Temperature = $(temperature) K is outside the typical validated range (50–150 K)."
+    end
```
fhagemann, line 84: Looking at maximum compatibility: how about still being able to define the temperature in the charge drift model, especially for the case where no temperature was explicitly defined for the semiconductor (and maybe throwing an error here in case two non-matching temperatures are defined in the semiconductor and the charge drift model)?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -124,7 +123,7 @@ end
 
 
 BoggsChargeTrappingModel(args...; T::Type{<:SSDFloat}, kwargs...) = BoggsChargeTrappingModel{T}(args...; kwargs...)
-function BoggsChargeTrappingModel{T}(config_dict::AbstractDict = Dict(); temperature::RealQuantity = T(78)) where {T <: SSDFloat}
+function BoggsChargeTrappingModel{T}(temperature::RealQuantity, config_dict::AbstractDict = Dict()) where {T <: SSDFloat}
```
fhagemann: This looks like a breaking change.. 
Any way that we can keep `config_dict` as the first argument?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -347,7 +346,7 @@ function CombinedChargeTrappingModel{T}(config_dict::AbstractDict = Dict(); temp
             if isnothing(model)
                 throw(ConfigFileError("`model` is defined but empty in config. Remove it or provide a valid name."))
         elseif model == "Boggs"
-                BoggsChargeTrappingModel{T}(config_dict; temperature)
+                BoggsChargeTrappingModel{T}(temperature, config_dict)
```
fhagemann:
```suggestion
                BoggsChargeTrappingModel{T}(config_dict, temperature)
``` 
?

test/test_charge_drift_models.jl
fhagemann, line 1: Some of the changes you are proposing would be breaking changes, requiring a new minor version `0.12`..
Is there any way this can be avoided? Is it because the temperature needs a default value?
Maybe set it to `nothing` and deal with it in the functions (if it is `nothing`, use the one from the config. If it has a value, it was passed from the semiconductor or by the user -- or something like that)?

src/ChargeTrapping/ChargeTrapping.jl
fhagemann: Maybe give `temperature` a default value `missing`, and only read it from the config file if it is `missing` (?)

examples/example_config_files/ivc_inactivelayer.yaml
```diff
@@ -36,7 +36,6 @@ detectors:
         value: -1e10cm^-3
     charge_drift_model:
       model: InactiveLayerChargeDriftModel
-      temperature: 90K
```
fhagemann, line 39: Maybe still allow for this syntax (to make this non-breaking), but check that it matches the temperature of the semiconductor, if any was given (?)

claudiaalvgar:
Just a follow up on the last commits: In the Charge Drift Model (CDM) the temperature is always taken from the semiconductor since ```_drift_charge!``` reads ```det.semiconductor.temperature``` for diffusion independently of the CDM temperature, so allowing a different temperature in the CDM config (as it was before this PR) would silently use inconsistent temperatures for mobility and diffusion. The semiconductor temperature always has a default value (77 K for HPGe, 293 K otherwise) so it is never undefined. A temperature key in the CDM config is still accepted (so that this PR is not a breaking change) but it must match the semiconductor temperature, otherwise an error is thrown.

For the Charge Trapping Model the temperature can come from either the argument or from temperature inside the parameters block of the config. If both are present they must agree; if only one is provided that value is used; if neither is provided an error is thrown. This is safe because no parallel code path reads ```det.semiconductor.temperature``` for trapping.
Now existing configs that have a matching temperature in the CDM continue to work but I still think we should delete this temperature from the yaml examples.

src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
```diff
@@ -68,6 +68,17 @@ function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
 
     temperature::T = _parse_value(T, temperature, input_units.temperature)
 
+    if haskey(config, "temperature")
+        cdm_temperature::T = _parse_value(T, config["temperature"], input_units.temperature)
+        if cdm_temperature != temperature
+            throw(ConfigFileError(
+                "Temperature mismatch: InactiveLayerChargeDriftModel defines temperature = $(cdm_temperature) K, " *
+                "but the semiconductor temperature is $(temperature) K. " *
+                "Remove `temperature` from the charge drift model and define it only in the semiconductor."
```
fhagemann: I mean: people could still define them both in the semiconductor and in the charge drift model, as long as they match. Maybe reword this accordingly?
```suggestion
                "Define `temperature` in the semiconductor and either remove it from the charge drift model or make sure that the temperatures match."
```
or something like this?

src/ChargeTrapping/ChargeTrapping.jl
```diff
@@ -123,19 +123,34 @@ end
...
-    temperature::T = _parse_value(T, temperature, internal_temperature_unit)
 
     if haskey(config_dict, "model") && !haskey(config_dict, "parameters")
         throw(ConfigFileError("`BoggsChargeTrappingModel` does not have `parameters`"))
     end
 
     parameters = haskey(config_dict, "parameters") ? config_dict["parameters"] : config_dict
-    
+
+    temp::Union{T, Missing} = temperature === missing ? missing : _parse_value(T, temperature, internal_temperature_unit)
```
fhagemann: looking at this, maybe a better default value than `missing` would be `NaN`, because `NaN` is of type `RealQuantity` and `T` already and would need these `Unions` then.
```suggestion
    temp::T = isnan(temperature) ? T(NaN) : _parse_value(T, temperature, internal_temperature_unit)
```

Even though, `missing` might be semantically the right choice here..

src/ChargeDriftModels/InactiveLayerChargeDriftModel/InactiveLayerChargeDriftModel.jl
```diff
@@ -63,11 +63,25 @@ function _calculate_mobility_with_impurities(
 end
 
 function InactiveLayerChargeDriftModel{T}(config::AbstractDict,
-        imp_model::AbstractImpurityDensity, input_units::NamedTuple,
+        imp_model::AbstractImpurityDensity, input_units::NamedTuple, temperature::RealQuantity
     ) where {T <: SSDFloat}
 
-    temperature = _parse_value(T, get(config, "temperature", 90u"K"), input_units.temperature)
+    temperature::T = _parse_value(T, temperature, input_units.temperature)
```
fhagemann: Maybe also call this `temp` to avoid the naming clash with the newly added argument `temperature` of the function.

src/SolidStateDetector/Semiconductor.jl
```diff
@@ -70,7 +70,7 @@ function Semiconductor{T}(dict::AbstractDict, input_units::NamedTuple, outer_tra
         model = Symbol(dict["charge_drift_model"]["model"])
         cdm = if isdefined(SolidStateDetectors, model) && getfield(SolidStateDetectors, model) <: AbstractChargeDriftModel
             if model == :InactiveLayerChargeDriftModel
-                InactiveLayerChargeDriftModel{T}(dict["charge_drift_model"], impurity_density_model, input_units)
+                InactiveLayerChargeDriftModel{T}(dict["charge_drift_model"], impurity_density_model, input_units, temperature)
```
fhagemann: This can result in the following:
If NO temperature is defined in the semiconductor, and e.g. 80K in the charge drift model, then here we default to 77K (for germanium) in the semiconductor and would throw an error. Is this what we want
(Especially because 77K is nowhere defined in the configuration file, but just set as default value in the code here).

test/test_charge_drift_models.jl
```diff
@@ -119,8 +119,30 @@ end
...
+            "model" => "Boggs",
+            "parameters" => Dict("nσe" => "0.001cm^-1", "nσh" => "0.0005cm^-1", "temperature" => "100K")
+        )
+        @test_throws SolidStateDetectors.ConfigFileError Simulation{T}(config_dict)
+    end
+    @testset "Temperature from config only (no argument)" begin
+        ctm = @test_nowarn BoggsChargeTrappingModel{T}(Dict(
+            "parameters" => Dict("nσe" => "0.001cm^-1", "temperature" => "78K")
+        ))
+        @test ctm.temperature == T(78)
```
fhagemann, line 145: Does this ever happen when you build this from a config file?
Or are we always passing the semiconductor temperature (or the default value of 77K for germanium, if nothing is defined)?

src/SolidStateDetector/Semiconductor.jl
fhagemann, line 69: Just asking for opinions, not asking to change:
where is this used? Would it make sense to set this also to `NaN` and evaluate default values somewhere else, just to not lose the information that no temperature was defined in the semiconductor? Where is this used?

fhagemann (commented):
Almost there :)
I would also like to have something added to the documentation about which temperature takes priority and where those temperatures should (preferably) be defined!

src/SolidStateDetector/Semiconductor.jl
claudiaalvgar: Yes, that is intentional. If no temperature is defined in the semiconductor, it defaults to 77K, and if the CDM defines 80 K, the two would be inconsistent: diffusion would use 77K while mobility would use 80K and this was a silent inconsistency before. The ConfigFileError forces the user to remove temperature from the CDM config parameters (and use the default semiconductor temperature of 77K) or add temperature: 80K to the semiconductor block (and if wanted keep the also the 80K in the CDM config parameters)

test/test_charge_drift_models.jl
claudiaalvgar: No, when building from a config file, the semiconductor temperature always has a value that is passed as an argument. The case where temperature comes only from the parameters config dict only arises in direct uses like in the test, never via Simulation{T}(config_file)

src/SolidStateDetector/Semiconductor.jl
claudiaalvgar: This I don't really mind, I did it this way to preserve current implementation of the default semiconductor temperature. We can probably keep the physical defaults (77 K for HPGe, 293 K otherwise) to preserve backwards compatibility, but add a @warn when no temperature is defined in the semiconductor config so the user is explicitly informed that a default is being used?

claudiaalvgar:
I'll add the temperature info to the documentation once we have a final picture of what we want to do with the default semiconductor temperature

src/SolidStateDetector/Semiconductor.jl
fhagemann: I like that idea. This warning might then also encourage people to add a temperature to the semiconductor.

src/SolidStateDetector/Semiconductor.jl
fhagemann: I see.
It's just that the error message might then say "you defined 77K in the semiconductor", even though the user might have never defined 77K in the semiconductor and it was just the default value.
But I guess that's fine.

src/SolidStateDetector/Semiconductor.jl
claudiaalvgar: Yes, I agree. I'd be happy for you to add the info messages and documentation for the new changes wherever you think would be most helpful for the user. I'm not entirely sure where the best place is to add them

fhagemann (approved):
This looks great. I might reformat the "notes" in the documentation to appear as actual notes and will merge this afterwards 😄

PR #612, opened 2026-06-22
Add functions to analytically adjust electric potential to match depletion/bias 

hervasa2:
## Summary

Added `adapt_to_depletion_and_bias_voltage!`. 

`adapt_to_depletion_and_bias_voltage!(sim, dep, bias)` adapts a `Simulation` in place to a target depletion voltage `dep` and operating (bias) voltage `bias` by rescaling the impurity density component of the electric potential (`f = dep / dep_sim`) and swapping in the bias contact potential via superposition — without re-solving the field. Note that the actual `sim.detector.semiconductor.impurity_density_model` and `sim.detector.contacts[id].potential` fields are also modified in place to reflect the change.

Tests check that this is performed correctly, and that after `adapt_to_depletion_and_bias_voltage!` is run on a simulation, a subsequent `calculate_electric_potential!` returns compatible results.
This is then compared to the manual way a user would have to do things: initializing a fresh simulation, modifying the `sim.detector.semiconductor.impurity_density_model` and `sim.detector.contacts[id].potential` fields and then running `calculate_electric_potential!`. 

This function is useful when the user wants to modify the simulation on the fly to reflect real-world conditions: a different measured depletion voltage and changing operating voltages without having to recalculate the electric potential. Note that in this process the weighting potential will be calculated and populated in the simulation.

Note that this only works on depleted simulations. This check is done when `estimate_depletion_voltage` is called internally.

## Changes

- **Export** `adapt_to_depletion_and_bias_voltage!`.
- **Input validation**: `dep` and `bias` must share the same (non-zero) sign, and `abs(bias) > abs(dep)` (the detector must be over-depleted at the operating voltage). Both polarities are supported.
- **Docstring** describing the impurity-rescaling / superposition mechanism, arguments, keywords, and an example.
- **Tests** in `test/test_depletion.jl`:
  - round-trip: after adapting, the bias contact sits at `bias` and the re-estimated depletion voltage ≈ `dep`;
  - an in-place re-solve still matches;
  - a freshly built equivalent simulation (same adapted impurity model + `bias`) reproduces the same depletion voltage.

## Testing

`test/test_depletion.jl` passes locally (11/11).


fhagemann:
Can you let me know once you’re done force pushing for me to look at this PR? 😅

hervasa2:
Sorry Im done. Just some last minute naming changes for consistency

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,76 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+adapt_to_depletion_and_bias_voltage!(sim, 1000u"V", 1500u"V")
+```
+
+!!! note
+    The accuracy of the result depends on the precision of the initial simulation, since the
+    impurity density is rescaled rather than re-solved.
+
+See also [`estimate_depletion_voltage`](@ref).
+"""
+function adapt_to_depletion_and_bias_voltage!(sim::Simulation{T}, dep::RealQuantity, bias::RealQuantity;
```
fhagemann: I feel like we can probably find a better name to make clear that ONLY the electric potential (but e.g. not the electric field) is updated.

- Can we avoid the verb `adapt` to avoid confusion with `adapt` from [Adapt.jl](https://github.com/JuliaGPU/Adapt.jl)? I know we used `adapt` for the internal function to map the grids of electric and weighting potentials, but for an exported function I would like to be a bit more careful with wording.
- It would also be nice to have a method that doesn't expect a `bias` and just lets the `bias` remain whatever it was before. Same thing if someone were to just change the bias without planning on changing the depletion voltage. Would it make sense to split this into two functions (one called something like `update_electric_potential_to_match_depletion!` and the other one `update_electric_potential_to_match_bias!` -- these are probably not the best names to choose from), which are then called by this overall function that allows to change both if wanted? Not sure what's the best way to go here, so I'm open for suggestions.

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,76 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+    contact_id::Int = determine_bias_voltage_contact_id(sim.detector),
+    verbose::Bool = true,
+    kwargs...) where {T <: AbstractFloat}
+    if ismissing(sim.weighting_potentials[contact_id]) || sim.weighting_potentials[contact_id].grid != sim.electric_potential.grid
+        _adapt_weighting_potential_to_electric_potential_grid!(sim, contact_id)
+    end
+    dep::T = _parse_value(T, dep, internal_voltage_unit)
+    bias::T = _parse_value(T, bias, internal_voltage_unit)
+    @assert dep * bias > 0 "The depletion voltage ($(dep)$(internal_voltage_unit)) and operating voltage ($(bias)$(internal_voltage_unit)) must have the same (non-zero) sign."
+    @assert abs(bias) > abs(dep) "The operating voltage ($(bias)$(internal_voltage_unit)) must exceed the depletion voltage ($(dep)$(internal_voltage_unit)) in magnitude (the detector must be over-depleted)."
```
fhagemann: Can we use something different than `@assert`? Maybe `ArgumentError` or `ConfigFileError`?

test/test_depletion.jl
```diff
@@ -32,4 +32,29 @@ T = Float32
...
     @test_logs (:info,) (:info,) (:warn, r".*not in the specified range.*") estimate_depletion_voltage(sim, U_est/3, 0)
+
+    # `adapt_to_depletion_and_bias_voltage!` rescales the impurity density and swaps in a new
+    # contact potential so the simulation matches a target depletion voltage `dep` and
+    # bias voltage `bias`, without re-solving the field. Check the round-trip: after adapting,
+    # the estimated depletion voltage should be ≈ `dep` and the bias contact should sit at `bias`.
+    dep_target = 2000u"V"
+    bias_target  = 2500u"V"
+    imp_model_before = sim.detector.semiconductor.impurity_density_model
+    adapt_to_depletion_and_bias_voltage!(sim, dep_target, bias_target, check_for_depletion = false, verbose = false)
```
fhagemann: Can you also add tests to show that the errors are thrown correctly?

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,76 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+    dep_sim = _parse_value(T, estimate_depletion_voltage(sim; contact_id = contact_id, verbose = verbose, kwargs...), internal_voltage_unit)
```
fhagemann: Can we keep this short, or is there any particular reason why this wouldn't work?
```suggestion
    dep_sim = _parse_value(T, estimate_depletion_voltage(sim; contact_id, verbose, kwargs...), internal_voltage_unit)
```

fhagemann: Would it make sense to allow the user to pass the expected depletion voltage also to the function? I can't judge how long `estimate_depletion_voltage` takes, especially if `sim.electric_potential` and `sim.weighting_potentials[contact.id]` are already defined on the same grid.

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,76 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+    impurity density is rescaled rather than re-solved.
+
+See also [`estimate_depletion_voltage`](@ref).
+"""
+function adapt_to_depletion_and_bias_voltage!(sim::Simulation{T}, dep::RealQuantity, bias::RealQuantity;
+    contact_id::Int = determine_bias_voltage_contact_id(sim.detector),
+    verbose::Bool = true,
+    kwargs...) where {T <: AbstractFloat}
+    if ismissing(sim.weighting_potentials[contact_id]) || sim.weighting_potentials[contact_id].grid != sim.electric_potential.grid
+        _adapt_weighting_potential_to_electric_potential_grid!(sim, contact_id)
```
fhagemann, line 374: In `_adapt_weighting_potential_to_electric_potential_grid!`, we run `update_till_convergence!` and essentially change the weighting potential (grid + update till convergence).
This contradicts the docstring saying that it is not re-solving anything. Maybe make this clearer?

fhagemann (commented):
This looks like a very useful function and I think it would be a nice addition to SSD.
I have some comments on the name of the function, the requirement of having to pass BOTH a target depletion and bias voltage, the docstring, and error handling in the tests :)

src/Simulation/Depletion.jl
hervasa2: I am in favor of splitting it into `update_electric_potential_to_match_depletion!` and `update_electric_potential_to_match_bias!`. But I would not add a third function that calls both. I am happy with the names you chose

fhagemann: I'm not sure I like the verb `update`, because that's usually the verb we use to actually run the SOR, and this is explicitly not doing that, though :sweat_smile:

fhagemann: So if there's any suggestion other than `update` or `adapt`, I'd be happy to discuss :)

tdixon97:
@hervasa2 this is of course very useful. But I do not understand this part about the fields not being relculated? Doesn't this have to happen? Maybe it can be at the end, since the fields are needed to get the right charge drift

src/Simulation/Depletion.jl
tdixon97: Isn't what is really happening here is that the impurities are rescaled. I think the current function name is a bit obtuse and doesn't really tell you that (regardless of the update / adapt discussion).

fhagemann:
My understanding is that the electric **POTENTIAL** is recalculated, but NOT using the whole SOR relaxation algorithm, but just a linear combination of the old electric potential and the weighting potential of the contact that you're applying the bias voltage to (which allows you to skip the costly SOR algorithm step).

However, it is true that you need to recalculate the electric **FIELD** still would need to be recalculated with `calculate_electric_field!` in order for the charge drift to be based on the updated electric field.

src/Simulation/Depletion.jl
fhagemann: The data in `sim.electric.potential.data` is also updated in this function, but not using the standard way of using the SOR algorithm, but by calculating it from a linear combination of the electric potential and the weighting potentials on the same grid.

fhagemann: [Screencast from 2026-06-24 13-48-57.webm](https://github.com/user-attachments/assets/37e2bba7-94d0-48c6-99c3-321162f49819)

Electric potential from impurity density + Electric potential from bias voltage = Final electric potential

So, here, it's subtracting the contribution from the bias voltage, rescaling it to match the depletion voltage, and adding the electric potential with an updated bias voltage back onto it.

tdixon97: I wonder if in practice youd want to recompute the electric_potiential in the standard way before running sims (even if a bit unnecessary)

tdixon97:
Yeah then the docs need to reflect that?

src/Simulation/Depletion.jl
fhagemann: @hervasa2 One could think about adding an (optional) extra convergence step afterwards.

src/Simulation/Depletion.jl
hervasa2: Added `U_min` and `U_max` as `args` which passes the range to `estimate_depletion_voltage`

hervasa2:
**Reworked the API into two analytic functions** *(updated since the original comment to reflect later renames — `05140d8c`, `b9f9dc7d`)*

`adapt_to_depletion_and_bias_voltage!` is replaced by two focused, exported functions that each adjust the field analytically via superposition (no full cold re-solve):

- **`scale_electric_potential_impurity_to_match_depletion!(sim, dep)`** — rescales the impurity density by `f = dep / dep_sim` so the detector's depletion voltage matches `dep`, updating `sim.electric_potential` by scaling its impurity contribution (`f·ϕ + V·(1−f)·ϕV`). **Returns** the applied scaling factor `f`.
- **`shift_electric_potential_to_match_bias!(sim, bias)`** — shifts the bias contact potential, updating the field by `+= (bias − V_old)·ϕV`.

To match both a target depletion and bias, call them in sequence (depletion first, then bias with `check_against_depletion_voltage = false`).

**Other changes:**

- **`reconverge_electric_potential` (default `true`)** — after the analytical superposition, the field is relaxed to convergence with `update_till_convergence!` and the `bulk` / `undepleted` / `inactive-layer` point-type markings are restored (these are normally applied only by `calculate_electric_potential!`, not by a bare `update_till_convergence!`). Implemented in the helper `_update_electric_potential_till_convergence_and_mark_bits!` (in `Simulation.jl`). Set `false` to keep the purely analytical result.
- **Validation** — both functions throw `ArgumentError` unless the target and reference voltages share a (non-zero) sign and the detector is over-depleted (`abs(reference) > abs(target)`).
- **`estimate_depletion_voltage`** now **throws `ArgumentError`** (previously warned) when the result falls outside the search range, so callers can catch it rather than silently get an edge value.
- **`update_till_convergence!` and `refine!`** now default `depletion_handling` to the sim's current state (`has_depletion_handling(sim.point_types)`) instead of `false`, so re-relaxing an already-depletion-handled sim keeps that handling.
- Docstrings updated throughout; tests reworked for the split API with added error-path coverage. Local `test_depletion.jl`: 15/15 passing.

**Note:** this changes the public API — `adapt_to_depletion_and_bias_voltage!` is removed in favor of the two functions above (`scale_…!` and `shift_…!`).

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,154 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+"""
+    scale_electric_potential_impurity_to_match_depletion!(sim::Simulation{T}, dep::RealQuantity,
+        U_min::RealQuantity = minimum(broadcast(c -> c.potential, sim.detector.contacts)),
+        U_max::RealQuantity = maximum(broadcast(c -> c.potential, sim.detector.contacts));
+        contact_id::Int = determine_bias_voltage_contact_id(sim.detector),
+        verbose::Bool = true,
+        reconverge_electric_potential::Bool = true,
+        kwargs...) where {T <: AbstractFloat}
+
+Rescales the impurity density of a [`Simulation`](@ref) in place so that its depletion voltage
```
tdixon97, line 220: I think you could add some formulas here (even if very simple), eg. showing its multiplicative not additive etc.

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,154 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+matches the target `dep`, **without a full field re-solve** and **without changing the bias**.
+
+The `impurity_density_model` is scaled by `f = dep / dep_sim`, where `dep_sim` is the current
```
tdixon97, line 223: Need to explicitly state the electric field now needs to be resolved?

src/Simulation/Simulation.jl
```diff
@@ -638,6 +640,20 @@ function update_till_convergence!( sim::Simulation{T,CS},
...
+    update_till_convergence!(sim, ElectricPotential; kwargs...)
+    mark_bulk_bits!(sim.point_types.data)
+    if has_depletion_handling(sim.point_types)
+        mark_undep_bits!(sim.point_types.data, sim.imp_scale.data)
+        if isdefined(sim.detector.semiconductor.impurity_density_model, :surface_imp_model)
+            mark_inactivelayer_bits!(sim.point_types.data)
+        end
+    end
+    nothing
+end
```
fhagemann: I don't really think that we need a function doing all of this, but a general `mark_bits!(sim)` might be nice. 

```suggestion
function mark_bits!(sim::Simulation)
    mark_bulk_bits!(sim.point_types.data)
    if has_depletion_handling(sim.point_types)
        mark_undep_bits!(sim.point_types.data, sim.imp_scale.data)
        if isdefined(sim.detector.semiconductor.impurity_density_model, :surface_imp_model)
            mark_inactivelayer_bits!(sim.point_types.data)
        end
    end
    nothing
end
```


Then, your functions can call both
```julia
if reresolve_electric_potential
    update_till_convergence!(sim, ElectricPotential; kwargs...)
    mark_bits!(sim)
end
```

This might then also allow to avoid code duplication everywhere else where we mark bits in the point types.

src/SolidStateDetectors.jl
```diff
@@ -77,6 +77,7 @@ export ConstantImpurityDensity, LinearImpurityDensity, ThermalDiffusionLithiumDe
 export LinBouleImpurityDensity, ParBouleImpurityDensity, LinExpBouleImpurityDensity, ParExpBouleImpurityDensity, SplineBouleImpurityDensity
 export NoChargeTrappingModel, BoggsChargeTrappingModel, ConstantLifetimeChargeTrappingModel, CombinedChargeTrappingModel
 export get_active_volume, is_depleted, estimate_depletion_voltage, set_point_type_depletion_handling!
+export scale_electric_potential_impurity_to_match_depletion!, shift_electric_potential_to_match_bias!
```
fhagemann: I thought about names again --> maybe we could highlight what the main thing the function does first (scale the impurity density or adjust the bias voltage) and then also mention the electric potential as part of the function name? Just to throw in some suggestions:
- `scale_impurity_and_electric_potential_to_match_depletion!`
- `adjust_bias_voltage_and_electric_potential!`

?

I'm not sure I like `shift` for the `bias` case, because it's still somewhat scaling the weighting potential contribution with the bias and it's not just a general offset.

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,154 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+    V = sim.detector.contacts[contact_id].potential
+    bias = _parse_value(T, bias, internal_voltage_unit)
+    if check_against_depletion_voltage 
+        dep = _parse_value(T, estimate_depletion_voltage(sim; contact_id, verbose, kwargs...), internal_voltage_unit) 
+        dep * bias <= 0 && throw(ArgumentError("The depletion voltage ($(dep)$(internal_voltage_unit)) and operating voltage ($(bias)$(internal_voltage_unit)) must have the same (non-zero) sign."))
+        abs(bias) <= abs(dep) && throw(ArgumentError("The operating voltage ($(bias)$(internal_voltage_unit)) must exceed the depletion voltage ($(dep)$(internal_voltage_unit)) in magnitude. The detector must be over-depleted to superpose the contributions of the bias voltage and impurity density to the electric potential."))
+    end
+
+    if ismissing(sim.weighting_potentials[contact_id]) || sim.weighting_potentials[contact_id].grid != sim.electric_potential.grid
+        _adapt_weighting_potential_to_electric_potential_grid!(sim, contact_id)
```
fhagemann, line 374: On the same note about naming (but does not necessarily have to be in this PR, just as a note to myself):
Probably a better name for this internal function would be something like
`_map_weighting_potential_onto_electric_potential_grid!`

If we were to make that name change, I would still allow for the old name but add a function deprecation and maybe remove it in the next minor version release.

src/Simulation/Depletion.jl
```diff
@@ -204,6 +204,154 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
...
+the electric potential** and **without changing the impurity density**.
+
+Since the electric potential is linear in the applied bias, the stored `sim.electric_potential` is
+modified by adding `(bias - V_old) · ϕV`, where `V_old` is the current contact potential and `ϕV` is
+the [`WeightingPotential`](@ref) of the bias contact.
+
+!!! note
+    The [`WeightingPotential`](@ref) of the bias contact (`contact_id`) may be modified by this
+    function: if it is missing or does not share the grid of the [`ElectricPotential`](@ref), it
+    is calculated/adapted onto that grid.
```
fhagemann: I like the verb `map` also here for the docstring.
```suggestion
    is calculated on or mapped onto that grid.
```

src/Simulation/Depletion.jl
fhagemann: +1 
(Or add another keyword argument to this function, e.g. `recalculate_electric_field`, with default set to `true`, so that it's automatically done, but can be disabled for whatever reason if needed)

hervasa2: I thick we should just make a note about it in the docstring. All functions which explicitly act on the electric potential in SSD leave the electric field unchanged. This should be the same here. At most I would put a warning (but it would have to be in all functions that calculate the electric potential) for sims with a populated electric field to recalculate it.

src/Simulation/Simulation.jl
hervasa2: I was also thinking along the same lines of using it elsewhere. But since these are often called together I grouped them. But I like your proposal.

src/SolidStateDetectors.jl
hervasa2: I agree about shift, but then the same would apply to scale in the first function. Technically you aren't scaling the electric potential as a whole, which is why I put "electric_potential_impurity". That being said I do like your opinions.

src/Simulation/Depletion.jl
tdixon97: Sure then what would help alot users would be a little block of code in the manual of how this is meant to be used

I guess the intended usage is after the electric potiental but before the electric field. Or can the function emit eg. a warning if the electric field was already computed? Or if the electric potiental was not?

gipert:
Hi all, thanks for this new feature! This kinda urgently required from the simflow side (https://github.com/legend-exp/legend-simflow/pull/272), but unfortunately @hervasa2 will be busy with detector production matters in the next weeks.

What is required to merge this and release a new SSD version? @tdixon97 and me are available to push this through, if needed.

fhagemann:
My biggest concern would be: if we export this functions, a new release would require to have:
- function names that we will not change until the next major breaking release
- argument and keyword arguments that we do not wish to rename
- ideally some up-to-date documentation that describes both what the function does and how to use it

If this is super urgent, I would propose:
Not exporting these functions (for now), essentially still treating them as internal function in the next release, but allow for you to import them in your simulation workflow via `using SolidStateDetectors: name_of_the_function`.
Then, I would also say that it's not too critical to have a final docstring or warnings to guide the user if e.g. the electric field needs to be recalculated.

If that is OK with you, I can take care of the missing comments later today, merge the PR with these function as internal functions and tag a new patch release.

Alternatively, you could add these functions to https://github.com/legend-exp/legend-simflow/blob/main/workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl, until this PR is mature enough to be released with these functions being exported, well-documented and handling errors/warning well.

gipert:
Your first proposal sounds great to me Felix! The alternative would also work for us but I believe these functions are already useful also outside the Simflow (for e.g. @hervasa2), so we would likely prefer to push this PR through.

fhagemann:
I agree that this should end up in SSD after all, but people should also need to know to use it properly and not be confused by any docstring or function names.

But then: let me tackle this (renaming, undoing the export, merging and tagging a release) later today, probably when it's evening in Europe.

hervasa2:
Thank you Felix. I am happy with whichever names you chose, but I strongly suggest we export these functions rather sooner than later. Not necessarily in this release, but it should not fall off our plate. Also please see my comment about NOT including `recalculate_electric_field`.

src/SolidStateDetectors.jl
fhagemann: Then, how about using `adjust` for both? This also would make it more obvious that these functions work in a very similar way

src/Simulation/Depletion.jl
fhagemann: --> added with the keyword argument `reconverge electric_potential`

src/SolidStateDetectors.jl
fhagemann: --> renamed to start with `adjust_`

src/Simulation/Depletion.jl
fhagemann: --> I added an `@info` about recalculating the electric field, that is only shown if the electric field was already calculated, and that can be omitted if `verbose` is set to `false`.

fhagemann:
Ok, I renamed the functions, added some `@info` in case the electric field was already calculated, and will keep them exported. It might make sense to also allow to pass simulation parameters to the re-convergence step (e.g. right now, users cannot select to reconverge the electric potential on a GPU), but this can be a future PR which should be non-breaking.

fhagemann (approved):

gipert:
Hi all, good to merge?

fhagemann:
Yes!

fhagemann:
I started tagging a new release, might maybe take 15-60min until it's registered in the Julia general registry:
https://github.com/JuliaRegistries/General/pull/159779

fhagemann:
Ok, 0.11.4 including this change is released now.

tdixon97:
Thanks @hervasa2 and @fhagemann this is a really helpful development!

PR #616, opened 2026-07-17
Point types documentation and FDD calculation

claudiaalvgar:
A new manual page is added to document the PointTypes behaviour. The six bit flags, their combined values, how they are used by ```is_depleted```, ```get_active_volume```, and ```estimate_depletion_voltage``` to exclude the inactive layer, how ```mark_inactivelayer_bits!``` propagates the inactive layer flag outward from the doped contact, and a detailed description of how the new ```get_FDD``` function extracts the two contours and computes the FDD thickness. 

Example plots show the inner contour overlaid on the point-type map and a heat-map of FDD thickness across the detector cross-section.
FDD_example

The new function ```get_FDD```, computes the Full Depletion Depth (FDD), the thickness of the inactive lithium-doped layer. For each inner contour point it returns the coordinates of the transition between the inactive layer and the active bulk (r_inner/z_inner), the coordinates of the nearest point on the doped contact surface (r_outer/z_outer), and the perpendicular thickness of the inactive layer. 
When the impurity model is ```PtypePNJunctionImpurityDensity``` or ```ThermalDiffusionLithiumDensity```, the thickness is computed analytically via ```ConstructiveSolidGeometry.distance_to_surface```, giving exact perpendicular distances for any surface orientation (vertical, horizontal, or diagonal). For other models it falls back to the Euclidean distance to the nearest contact-surface grid tick. Both Cylindrical and Cartesian variants are implemented.

The outer contour (r_outer/z_outer) is collected from grid ticks that carry ```inactive_contact_bit``` directly, which lie on the contact surface itself since SSD places grid ticks at contact boundaries. This means that for axis-aligned surfaces the Euclidean distance between inner and outer contour points equals the analytical thickness exactly; for diagonal surfaces (e.g. a cone face) a small residual difference remains because the nearest contact tick is not the exact perpendicular foot.
A helper function ```_dist_to_inactive_contact``` is added to extract the distance_to_contact from the impurity density model, or return nothing to trigger the Euclidean fallback. Tests for get_FDD are added to ```test_depletion.jl```


docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+Six single-bit flags are defined, each encoding one physical property:
+
+| Bit | Constant | Hex | Meaning |
+|-----|----------|-----|---------|
+| 0 | `update_bit` | `0x01` | `1` → point is updated in SOR; `0` → fixed point (contacts, exterior boundaries) |
+| 1 | `undepleted_bit` | `0x02` | `1` → point is undepleted; `0` → fully depleted |
+| 2 | `pn_junction_bit` | `0x04` | `1` → point lies inside the semiconductor volume |
+| 3 | `bulk_bit` | `0x08` | `1` → all points in the 3×3×3 neighbourhood (face-, edge-, and corner-adjacent) also have `pn_junction_bit` and `update_bit` set |
+| 4 | `inactive_layer_bit` | `0x10` | `1` → point is part of the inactive (dead) layer |
+| 5 | `inactive_contact_bit` | `0x20` | `1` → point is part of the contact adjacent to the inactive layer |
```
fhagemann: I really like this table! :)
I would argue that we would like to start counting at `1` instead of `0`, in the spirit of Julia's array-indexing also starting at 1.

```suggestion
| 1 | `update_bit` | `0x01` | `1` → point is updated in SOR; `0` → fixed point (contacts, exterior boundaries) |
| 2 | `undepleted_bit` | `0x02` | `1` → point is undepleted; `0` → fully depleted |
| 3 | `pn_junction_bit` | `0x04` | `1` → point lies inside the semiconductor volume |
| 4 | `bulk_bit` | `0x08` | `1` → all points in the 3×3×3 neighbourhood (face-, edge-, and corner-adjacent) also have `pn_junction_bit` and `update_bit` set |
| 5 | `inactive_layer_bit` | `0x10` | `1` → point is part of the inactive (dead) layer |
| 6 | `inactive_contact_bit` | `0x20` | `1` → point is part of the contact adjacent to the inactive layer |
```

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+using SolidStateDetectors: update_bit, undepleted_bit, pn_junction_bit,
+                            bulk_bit, inactive_layer_bit, inactive_contact_bit
+
+pt = sim.point_types.data[i, j, k]   # PointType (UInt8) at one grid point
+is_fixed       = pt & update_bit         == 0
+is_undepleted  = pt & undepleted_bit     > 0
+in_pn_junction = pt & pn_junction_bit    > 0
+in_bulk        = pt & bulk_bit           > 0
+in_inactive    = pt & inactive_layer_bit > 0
+```
```
fhagemann, line 35: We also have convenience functions for this, not sure if we should also mention them here:
https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/71bf61d05135e472fea7643abaa7a1cebcc187a8/src/ScalarPotentials/PointTypes.jl#L36-L39

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+| 2 | `pn_junction_bit` | `0x04` | `1` → point lies inside the semiconductor volume |
+| 3 | `bulk_bit` | `0x08` | `1` → all points in the 3×3×3 neighbourhood (face-, edge-, and corner-adjacent) also have `pn_junction_bit` and `update_bit` set |
+| 4 | `inactive_layer_bit` | `0x10` | `1` → point is part of the inactive (dead) layer |
+| 5 | `inactive_contact_bit` | `0x20` | `1` → point is part of the contact adjacent to the inactive layer |
+
+They can be queried with bitwise AND:
+
+```julia
+using SolidStateDetectors: update_bit, undepleted_bit, pn_junction_bit,
+                            bulk_bit, inactive_layer_bit, inactive_contact_bit
```
fhagemann: `inactive_contact_bit` is not used in this code example.

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Information from Point Types
+
+### Depletion and Active Volume
+
+`is_depleted(sim.point_types)` returns `true` when no **bulk** point remains undepleted outside the inactive layer.
+Concretely, it checks whether any point simultaneously satisfies all three conditions:
+
+```
+bulk_bit = 1   AND   undepleted_bit = 1   AND   inactive_layer_bit = 0
```
fhagemann: Consistent spacing:
```suggestion
bulk_bit = 1    AND   undepleted_bit = 1   AND   inactive_layer_bit = 0
```

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
+# Point Types
+
+After running `calculate_electric_potential!`, every grid point in `sim.point_types` is assigned a `PointType` value — a `UInt8` bitmask that encodes the physical role of that point in the simulation.
+Point types are used throughout the code to decide which points are updated during the successive over-relaxation (SOR), whether the detector is depleted, which points belong to the inactive layer, and what volume counts as active.
```
fhagemann: Link the "Simulation Algorithm" section here?

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+Because inactive layer cells always carry `undepleted_bit = 1`, they are automatically excluded from the active volume.
+
+### Depletion Voltage Estimation
+
+`estimate_depletion_voltage` applies the same inactive layer exclusion: only points with `pn_junction_bit = 1 AND inactive_layer_bit = 0` (and `bulk_bit = 1` in the final refinement step) are considered.
+The returned voltage is therefore the bias at which the **active bulk** outside the inactive layer first becomes fully depleted.
+
+### Inactive Layer Automatic Detection
+
+The inactive layer is the permanently undepleted region that forms near the lithium-doped contact of a p-type HPGe detector.
```
fhagemann: This could be more generic (assuming that people would also use this for highly-doped contacts in silicon detectors for example.

```suggestion
The inactive layer is the permanently undepleted region that forms near highly-doped contact of a semiconductor detector.
```
?

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+
+- **Outer contour** (`r_outer`, `z_outer`): the inner face of the doped contact, where the inactive layer meets the detector boundary.
+  Every grid tick inside the doped contact geometry carries `inactive_contact_bit = 1`.
+  For each inner contour point the nearest such contact tick (by Euclidean distance in the r-z plane) is selected as `r_outer`/`z_outer`.
+
+In both calculation paths the FDD measures the distance from the inner contour to the inner face of the contact; the contact's own physical thickness is never included:
+
+- **When the model stores a `distance_to_contact` function** (e.g. `ThermalDiffusionLithiumDensity` or `PtypePNJunctionImpurityDensity`): `thickness` is computed as `ConstructiveSolidGeometry.distance_to_surface` from the inner contour point to the contact geometry. For a point outside the contact this always returns the distance to the nearest (inner) surface of the contact, giving an exact analytical perpendicular distance regardless of surface orientation and regardless of how thick the contact is.
```
fhagemann: Is this correct?
I would naively assume that if `distance_to_contact` exists, this would be the function to use INSTEAD of `ConstructiveSolidGeometry.distance_to_surface` (?)


```suggestion
- **When the model stores a `distance_to_contact` function** (e.g. `ThermalDiffusionLithiumDensity` or `PtypePNJunctionImpurityDensity`): `thickness` is computed by evaluating `distance_to_contact` at the inner contour point. For a point outside the contact this always returns the distance to the nearest (inner) surface of the contact, giving an exact analytical perpendicular distance regardless of surface orientation and regardless of how thick the contact is.
```

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+- **Fallback**: Euclidean distance from the inner contour point to the nearest contact grid tick (r-z plane for cylindrical, 3D for Cartesian). Because the inner contour sits just outside the contact, the nearest tick is on the innermost layer of the contact for standard detector geometries, so the contact thickness is not counted here either.
+
+For **axis-aligned surfaces** (vertical bore wall, horizontal top/bottom face, vertical outer wall) the nearest contact tick lies exactly along the surface normal, so the Euclidean distance from `(r_inner, z_inner)` to `(r_outer, z_outer)` equals `thickness` exactly.
+For **diagonal surfaces** (e.g. a cone face) the nearest contact tick on the slanted surface is not the perpendicular foot, so the Euclidean distance can differ from `thickness` of the FDD (either direction, depending on grid snapping); use the `thickness` field for the correct value.
```
fhagemann: What is the `thickness` field? This is only explained later

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+T = Float64
+sim = Simulation{T}(SSD_examples[:IVCIlayer])
+sim.world = typeof(sim.world)(sim.world.intervals, ntuple(_ -> T(5e-5), 3))
+calculate_electric_potential!(sim, depletion_handling = true, use_nthreads = 4)
+all_fdd = get_FDD(sim)                             # full contour, all coordinates
+fdd_at_r = get_FDD(sim; r = 5u"mm")                # only inner-contour points nearest r = 5 mm
+````
+
+````@example fdd
+let
```
fhagemann: I would remove enclosing this in `let`. Or is there a specific reason why that is included?

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
     nothing
 end
 
+"""
+    _dist_to_inactive_contact(sim::Simulation)
```
fhagemann: Rename this to:

```suggestion
    _distance_to_inactive_contact(sim::Simulation)
```
?
`dist` could also be `distribution`

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+        return imp.surface_imp_model.distance_to_contact
+    elseif hasproperty(imp, :distance_to_contact)
+        return imp.distance_to_contact
+    else
+        return nothing
+    end
+end
+
+"""
+    get_FDD(sim::Simulation{T}; <keywords>)
```
fhagemann: It is rather unusual to use upper-case in functions in julia.
Can this become `get_fdd` or `get_full_depletion_depth` ?

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+
+Estimates the Full Depletion Depth (FDD), the thickness of the inactive layer.
+
+When the semiconductor uses a `PtypePNJunctionImpurityDensity` or `ThermalDiffusionLithiumDensity`
+impurity model, the thickness is computed via `ConstructiveSolidGeometry.distance_to_surface`
```
fhagemann: Same comment as above --> isn't this using `distance_to_contact`?

docs/src/man/point_types.md
```diff
@@ -0,0 +1,158 @@
...
+````@setup fdd
+using Plots
+using SolidStateDetectors
+using Unitful
+gr(fmt = :png) #hide
+
+T = Float64
+sim = Simulation{T}(SSD_examples[:IVCIlayer])
+sim.world = typeof(sim.world)(sim.world.intervals, ntuple(_ -> T(5e-5), 3))
+calculate_electric_potential!(sim, depletion_handling = true, use_nthreads = 4)
```
fhagemann, line 135: This takes quite some time to run, because the grid is very fine.

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+perpendicular distance to diagonal and curved surfaces.
+For other impurity models the thickness falls back to the Euclidean distance to the nearest
+contact-surface grid point (subject to grid-spacing discretization error on diagonal surfaces).
+
+!!! note "r_outer/z_outer and thickness"
+    `r_outer`/`z_outer` (and the Cartesian equivalents) are the nearest grid tick on the doped contact
+    surface (carries `inactive_contact_bit`).  Because SSD places grid ticks at
+    contact boundaries, these coordinates lie on the contact surface itself.
+
+    For diagonal surfaces, the nearest contact tick on the slanted surface may not be exactly along
```
fhagemann: I would argue that this not only applies to diagonal/inclined surfaces, but also in most cases where contact geometries are rotated.

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+
+At a given `r` (or `z`) there may be more than one entry — for example an inactive layer on
+both the top and bottom face of the detector at the same radius.
+
+## Returns
+* `r_inner`, `z_inner` (Cylindrical) or `x_inner`, `y_inner`, `z_inner` (Cartesian): 
+position of the inner boundary point (inactive layer / active bulk transition).
+* `r_outer`, `z_outer` / `x_outer`, `y_outer`, `z_outer`: nearest grid point on the doped contact surface
+(where the inactive layer meets the lithium-doped contact).
+* `thickness`: perpendicular distance from the inner boundary to the contact surface = FDD thickness.
```
fhagemann, line 456: Are these always vectors  / tuples, no matter if there is one or multiple values? Maybe also add that information here.

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+Returns an empty vector if no inactive layer is present in `sim.point_types`.
+
+"""
+function get_FDD(sim::Simulation{T, Cylindrical};
+                 r::Union{Nothing, RealQuantity} = nothing,
+                 z::Union{Nothing, RealQuantity} = nothing) where {T <: AbstractFloat}
+    @assert !ismissing(sim.point_types) "Please calculate the electric potential first using `calculate_electric_potential!(sim, depletion_handling = true)`"
+
+    pt      = sim.point_types.data
+    g       = sim.electric_potential.grid
```
fhagemann:
```suggestion
    g       = sim.point_types.grid
```
If you do this, the whole function only depends on `sim.point_types`.
You could then also define a `get_FDD` acting upon just `PointTypes{T,3,Cylindrical}` or `PointTypes{T,3,Cartesian}`, and then redefine this as
```julia
function get_FDD(sim::Simulation; kwargs...) 
    return get_FDD(sim.point_types; kwargs...)
end

# fallback, if point types are missing
function get_FDD(::Missing; kwargs...)
    throw(ArgumentError("Please calculate the electric potential first using `calculate_electric_potential!(sim, depletion_handling = true)`"))
end
```

I remember that accessing `sim.point_types` and its field multiple times might maybe be less performant (because `sim` is a mutable struct and `point_types` could in principle be missing. Not sure if that's still the case in julia-1.10 or newer, though.

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+    # Exact analytical distance from the contact geometry when available;
+    # falls back to nearest-outer-grid-point distance otherwise.
+    dist_fn = _dist_to_inactive_contact(sim)
+
+    result = NamedTuple{(:r_inner, :z_inner, :r_outer, :z_outer, :thickness), Tuple{T,T,T,T,T}}[]
+    for (i, k) in inner_ik
+        ri, zi   = T(r_ticks[i]), T(z_ticks[k])
+        min_d2   = typemax(T)
+        ro_best  = zero(T)
+        zo_best  = zero(T)
```
fhagemann: Is `zero(T)` the best default value? Maybe `NaN`?

src/Simulation/Depletion.jl
```diff
@@ -391,6 +391,223 @@ function adjust_bias_and_electric_potential!(sim::Simulation{T}, bias::RealQuant
...
+                zo_best = zo
+            end
+        end
+        thickness = isnothing(dist_fn) ? sqrt(min_d2) : dist_fn(CartesianPoint{T}(xi, yi, zi))
+        push!(result, (x_inner = xi, y_inner = yi, z_inner = zi, x_outer = xo_best, y_outer = yo_best, z_outer = zo_best, thickness = thickness))
+    end
+
+    if !isnothing(x)
+        x_val     = _parse_value(T, x, internal_length_unit)
+        x_nearest = T(x_ticks[argmin(abs.(x_ticks .- x_val))])
```
fhagemann: This seems to be doing that `SolidStateDetectors.searchsortednearest` does ?

src/SolidStateDetectors.jl
```diff
@@ -76,7 +76,7 @@ export ElectricFieldChargeDriftModel, ADLChargeDriftModel, ADL2016ChargeDriftMod
 export ConstantImpurityDensity, LinearImpurityDensity, ThermalDiffusionLithiumDensity, PtypePNJunctionImpurityDensity
 export LinBouleImpurityDensity, ParBouleImpurityDensity, LinExpBouleImpurityDensity, ParExpBouleImpurityDensity, SplineBouleImpurityDensity
 export NoChargeTrappingModel, BoggsChargeTrappingModel, ConstantLifetimeChargeTrappingModel, CombinedChargeTrappingModel
-export get_active_volume, is_depleted, estimate_depletion_voltage, set_point_type_depletion_handling!
+export get_active_volume, is_depleted, estimate_depletion_voltage, set_point_type_depletion_handling!, get_FDD
```
fhagemann: If we export this, I want to have a discussion about the naming ^^
See my comment above.

test/test_depletion.jl
```diff
@@ -77,3 +77,92 @@ T = Float32
...
+            p.r_outer > p.r_inner &&
+            p.r_outer < T(0.035)
+        end
+
+        @test !isempty(cone_face)
+        # Perpendicular distance is strictly less than the purely radial distance to the cone surface
+        @test all(p -> p.thickness < r_cone_at_z(p.z_inner) - p.r_inner, cone_face)
+        # Thickness matches the analytical perpendicular distance formula
+        @test all(p -> isapprox(p.thickness, perp_dist(p.r_inner, p.z_inner), atol = T(1e-4)), cone_face)
+    end
```
fhagemann: Why `let`?

test/test_depletion.jl
```diff
@@ -77,3 +77,92 @@ T = Float32
     @test_throws ArgumentError adjust_bias_and_electric_potential!(sim, 1000u"V", check_against_depletion_voltage = true, check_for_depletion = false, verbose = false)          # opposite sign
     @test_throws ArgumentError adjust_bias_and_electric_potential!(sim, dep_target / 2, check_against_depletion_voltage = true, check_for_depletion = false, verbose = false)    # |bias| < |dep|
 end
+
+@testset "Test Full Depletion Depth" begin
+    sim = Simulation{T}(SSD_examples[:IVCIlayer])
```
fhagemann, line 83: Maybe add a test here that throws the error in case the `point_types` were not calculated yet.

src/Simulation/Depletion.jl
fhagemann, line 456:
```suggestion
## Returns
* `r_inner`, `z_inner` (Cylindrical) or `x_inner`, `y_inner`, `z_inner` (Cartesian): 
  position of the inner boundary point (inactive layer / active bulk transition).
* `r_outer`, `z_outer` / `x_outer`, `y_outer`, `z_outer`: nearest grid point on the doped contact surface
  (where the inactive layer meets the lithium-doped contact).
* `thickness`: perpendicular distance from the inner boundary to the contact surface = FDD thickness.
```

Some formatting -> without the spaces it looks like this:

Image

docs/src/man/point_types.md
claudiaalvgar: You are right, this is a bit complicated way of explaining it since the code calls dist_fn(...) which calls distance_to_contact directly

docs/src/man/point_types.md
claudiaalvgar: No specific reason, there's only one block in the end so I deleted it

docs/src/man/point_types.md
claudiaalvgar: I increased the size, it's just to show how the surface refinement tool is used, I don't think we added an example of this

src/SolidStateDetectors.jl
claudiaalvgar: I changed to lower case but if you still prefer a different name, I'm happy to change it

fhagemann:
I just merged #624, so that should hopefully fix the failing windows test. Rerunning the tests for this PR now.

docs/src/man/point_types.md
fhagemann: Do we want convenience functions for `bulk_bit` and `inactive_contact_bit` as well?

docs/src/man/point_types.md
```diff
@@ -124,35 +131,33 @@ gr(fmt = :png) #hide
...
+    [pt.r_inner for pt in all_fdd], [pt.z_inner for pt in all_fdd],
+    markershape = :circle, markersize = 1,
+    color = :red, markerstrokewidth = 0)
+
+p2 = scatter(
+    [pt.r_inner * 1000 for pt in all_fdd],
+    [pt.z_inner * 1000 for pt in all_fdd],
+    marker_z = t_mm,
+    color = :heat, colorbar = true, colorbar_title = "thickness [mm]",
+    clims = (minimum(t_mm), maximum(t_mm)),
```
fhagemann:
```suggestion
    clims = extrema(t_mm),

**Review by fhagemann — COMMENTED** [review]
Looks good. Two minor comments and still the open question whether we want to keep calling it `get_fdd` or something else.

Based on the fact that we also call it `get_active_volume` or `calculate_capacitance`, I might be slightly in favor to avoid using the abbreviation `fdd` but call it `get_full_depletion_depth`. Thoughts?

File: docs/src/man/point_types.md (hunk 2 above)
Comment (claudiaalvgar) [reply docs/src/man/point_types.md:35]: I think we shouldn't since they don't have a concrete call site that needs them. The bitwise check is already shown in the docs in case it's needed

**Comment by claudiaalvgar** [comment]
In the last commit I changed ```get_fdd``` by ```get_full_depletion_depth```, I think it's more intuitive too, I used initially the abreviation because it's shorter

File: docs/src/man/point_types.md [hunk 23]
```diff
@@ -89,10 +89,10 @@ All consecutive points where `pn_junction_bit`, `undepleted_bit`, and `update_bi
 
 The contact carrying `inactive_contact_bit` is identified through the semiconductor impurity density model (`PtypePNJunctionImpurityDensity`).
 
-### Full Depletion Depth via `get_fdd`
+### Full Depletion Depth via `get_full_depletion_depth`
```
fhagemann: I would remove the function name from the title:
```suggestion
### Full Depletion Depth
```

test/test_depletion.jl
```diff
@@ -77,3 +77,94 @@ T = Float32
     @test_throws ArgumentError adjust_bias_and_electric_potential!(sim, 1000u"V", check_against_depletion_voltage = true, check_for_depletion = false, verbose = false)          # opposite sign
     @test_throws ArgumentError adjust_bias_and_electric_potential!(sim, dep_target / 2, check_against_depletion_voltage = true, check_for_depletion = false, verbose = false)    # |bias| < |dep|
 end
+
+@testset "Test Full Depletion Depth" begin
+    # Calling get_full_depletion_depth before the electric potential is calculated should throw ArgumentError
+    sim_empty = Simulation{T}(SSD_examples[:IVCIlayer])
+    @test_throws ArgumentError get_full_depletion_depth(sim_empty)
+
+    sim = Simulation{T}(SSD_examples[:IVCIlayer])
```
fhagemann: You don't need to create two simulations here, just check for the error to be thrown before calculating the electric potential:

```suggestion
    sim = Simulation{T}(SSD_examples[:IVCIlayer])
    @test_throws ArgumentError get_full_depletion_depth(sim_empty)
```

docs/src/man/point_types.md
fhagemann: Still not used in the example (?)

fhagemann (approved):
Just pushed the last changes. I will wait and see that the tests pass and then continue with merging this :)

PR #630, opened 2026-08-26
Support ArraysOfArrays v1

oschulz:
Adds ArraysOfArrays v1 to the compat bounds while keeping support for v0.5 and v0.6. All ArraysOfArrays API used by SolidStateDetectors, including `VectorOfVectors`, is available unchanged on v1 (`VectorOfVectors` stays a supported alias as of ArraysOfArrays v1.0.1).

Running CI against ArraysOfArrays v1 also needs the compatible RadiationDetectorSignals release (JuliaPhysics/RadiationDetectorSignals.jl#54), until then the resolver picks ArraysOfArrays v0.6.

Includes a patch version bump.

Created by generative AI.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -49,7 +49,7 @@ SolidStateDetectorsGeant4Ext = "Geant4"
 
 [compat]
 Adapt = "3, 4"
-ArraysOfArrays = "0.5, 0.6"
+ArraysOfArrays = "0.5, 0.6, 1"
 Clustering = "0.15"
 DataStructures = "0.17, 0.18, 0.19"
 Distributions = "0.24.5, 0.25"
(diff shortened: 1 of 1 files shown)


fhagemann:
Are you planning to release a new version with this right away?
Or should we release this together with #626 for example?

Project.toml
```diff
@@ -49,7 +49,7 @@ SolidStateDetectorsGeant4Ext = "Geant4"
 
 [compat]
 Adapt = "3, 4"
-ArraysOfArrays = "0.5, 0.6"
+ArraysOfArrays = "0.5, 0.6, 1"
```
fhagemann, line 52: Still keep the `VectorOfVectors` alias here ?
```suggestion
ArraysOfArrays = "0.5, 0.6, 1.0.1"
```

oschulz: Shouldn't be necessary, that was just an (incorrect) deprecation in ArraysOfArrays v1.0.0, so v1.0.1 is not strictly required.

oschulz:
Closes #631 .

oschulz:
https://github.com/JuliaRegistries/General/pull/166218

PR #632, opened 2026-09-02
Fix 615 fixed waveform

### LINKED ISSUE #615: Return fixed length array with add_baseline_and_extend_tail
In the function [`add_baseline_and_extend_tail`](https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e43fd4bcd3fb1e70a78651e505ac5312cc25884e/src/Event/Event.jl#L288-L305), the output padded waveforms are returned as Vectors. It would be nice to instead return as fixed-length Arrays, since they will all have the same length. This would be beneficial for future processing of waveforms since processing based on recorded data tends to assume a fixed-length waveform as determined by the DAQ. https://github.com/JuliaPhysics/SolidStateDetectors.jl/blob/e43fd4bcd3fb1e70a78651e505ac5312cc25884e/src/Event/Event.jl#L288-L305

gitpsii:
## Description

add_baseline_and_extend_tail` previously returned the waveform signal as a regular `Vector`, even though the waveform length is required to remain fixed for DAQ waveform processing.

This change wraps the allocated signal in a SizedVector with total_waveform_length as its static size.

### Testing

Added tests to verify that the returned signal is a SizedVector and has the expected length.

Fixes #615


fhagemann:
Thanks for the contribution:
I’ll leave the decision to @oschulz, who might have some opinion about using `ArraysOfArrays` here.

oschulz:
`total_waveform_length` is not known at compile time and can be quite large, I don't think we want to type-parameterize based on it.

gitpsii:
Thanks, that makes sense. I see that SizedVector{total_waveform_length would type-parameterize on a runtime value and could be problematic for large waveform lengths. What fixed-length array type/approach would you recommend for this function?

oschulz:
We could look into using FixedSizeArrays.

Project.toml
```diff
@@ -54,6 +55,7 @@ Clustering = "0.15"
 DataStructures = "0.17, 0.18, 0.19"
 Distributions = "0.24.5, 0.25"
 FillArrays = "0.9, 0.10, 0.11, 0.12, 0.13, 1"
+FixedSizeArrays = "1.3.0"
```
fhagemann: This seems to be the newest version of `FixedSizeArrays`, and it might be too restrictive. What's the lowest version of `FixedSizeArrays` that we could allow for, that has the functionality required here?
Could this just become `1` (same as `1.0.0`) instead?

test/Project.toml
```diff
@@ -21,6 +22,7 @@ TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
+FixedSizeArrays = "1.3.0"
```
fhagemann: We don't need to explicitly restrict the `compat` here in the `test/Project.toml`.
It will be based on whatever we define in the overall `Project.toml`


```suggestion
```

Project.toml
```diff
@@ -54,6 +55,7 @@ Clustering = "0.15"
 DataStructures = "0.17, 0.18, 0.19"
 Distributions = "0.24.5, 0.25"
 FillArrays = "0.9, 0.10, 0.11, 0.12, 0.13, 1"
+FixedSizeArrays = "1.0.0"
```
fhagemann, line 58:
```suggestion
FixedSizeArrays = "1"
```

fhagemann (commented):
@oschulz can you have a look at this and either approve or raise concerns/request changes?

I just have one tiny comment on the `Project.toml` formatting

PR #623, opened 2026-07-20
Bug fixes, correctness guards, performance and test coverage from an in-depth review

oschulz:
In-depth review of the package: verified bug fixes, new correctness guards, performance work, and a test-coverage push. 27 commits, each scoped and with regression tests; the full suite passes under CI-like conditions (Julia 1.12, forced bounds checks, Geant4 testset included). Solver changes are bit-identical unless stated.

## Bug fixes

- **Geometry/CSG**: surface normals of tapered `ConeMantle` (≈74° off for SI-scale tapers) and `EllipsoidMantle` (consumed by drift reflections and E-field plots); `ConeMantle` single-intersection sign; two-frame `frame_transformation` composed in the wrong order; partial-φ serialization crash; several parsing holes now raise `ConfigFileError`.
- **Fields/axes**: cylindrical E-field used a wrong denominator at the periodic φ-wrap (wrong `e_φ` on the last φ slice of every non-φ-symmetric map); `searchsortednearest` never snapped across the periodic seam; missing `multiplicity` method for periodic Cartesian axes; per-point allocations in the E-field builder.
- **SOR**: the periodic boundary condition on the compressed red-black dimension copied ghost rows into ghost rows — wrapped interior values never entered the update.
- **GDML export** (all verified against Geant4 readback): placement rotations were never emitted, so rotated primitives — including every partial-φ segment — collapsed onto φ = 0; polyhedra radii written as circumradius where GDML expects apothems (hexagons 4/3 too large in volume); boolean second-operand offsets lost or expressed in the wrong frame.
- **Detector construction/API**: doped-contact auto-detection could never run; `estimate_depletion_voltage` crash when both candidates lay in range; capacitance matrix could not store its documented `missing` entries; unknown trapping-model names silently disabled trapping.
- **Charge transport**: ADL2016 NaN velocities at zero field; unclamped `acos` NaN; `PtypePNJunctionImpurityDensity + offset` applied the offset twice; `cluster_detector_hits` skipped DBSCAN exactly for low-multiplicity events; floating-boundary drift corrupted shared step vectors across charges.

## Correctness guards (behavior changes)

- `Grid(sim)` rejects 2D and periodic φ-wedge grids for detectors without the corresponding φ-symmetry (the `CGD_CylGrid` example silently simulated its cuboid in 2D).
- `is_depleted` throws when the potential was computed without depletion handling (was warn + `true`, silently disabling the `estimate_depletion_voltage` safety check).
[...]

PR #634, opened 2026-09-07
Fix gauge invariance in electric potential depletion-handling

claudiaalvgar:
Some issues in the depletion-handling / SOR machinery all shared the same root cause: an implicit, unshiftable "0 V" reference baked into several places that should instead reference the detector's actual applied potentials. Fixed together because they're only fully verifiable together, the r0-axis fix alone still leaves gauge-dependent mismatches near threshold (see table below).

### `BoundaryConditionsCartesian.jl`
The `:infinite` boundary condition sets edge cells to decay toward literal 0, which is correct only when the lowest applied potential (`minimum_applied_potential`) equals 0V. Under gauge shifts, this decayed toward wrong values. Since the potential array itself now lives in the `gauge_ref_potential`-shifted frame (established once at setup, not per-iteration), reconciling it with `:infinite`'s actual target requires a second shift: `Δ = minimum_applied_potential - gauge_ref_potential`, applied to just the edge margin before the formula runs and undone after: `n → n - Δ → f·(n - Δ) → f·(n - Δ) + Δ`. This runs every SOR iteration, on narrow edge margins only.

### `BoundaryConditionsCylindrical.jl`
Same fix, applied to the r-axis and z-axis dispatchers for cylindrical grids, for the same reason.

### `PotentialCalculationSetupCartesian3D.jl` / `PotentialCalculationSetupCylindrical.jl`
Two independent fixes each:
- **Interior seed**: Changed from `zeros(T, ...)` to `fill(gauge_ref_potential, ...)` — a new struct field, the mean of the applied contact potentials (`0` for weighting-potential solves). An earlier version of this fix used `fill(minimum_applied_potential, ...)` instead, but `min` isn't antisymmetric under a simultaneous sign flip of bias and impurity density (`min` and `max` swap contacts), so two sign-mirrored, gauge-equivalent configurations could still seed inconsistently and converge to different fixed points. `gauge_ref_potential` (mean) is exactly antisymmetric under that flip, fixing this. `minimum_applied_potential` is retained separately as `:infinite`'s own decay target (see above) — the two references now serve different purposes and are not the same value.
- **`is_weighting_potential` guard** on `contact_potentials` (renamed from `contact_bias_voltages`): Weighting-potential solves use the 0/1 convention (target = 1, rest = 0), not the detector's real bias voltages. Without this guard, `minimum_applied_potential`/`gauge_ref_potential` would derive from real (possibly large) bias voltages even during weighting-potential solves, seeding incorrect values.

Also added: `estimate_depletion_voltage` (`Depletion.jl`) now combines *all* filtered depletion-voltage candidates via `max`/`min` instead of assuming exactly one via `only(...)`, fixing a crash on detectors with multiple candidates. The "shift back to real units" step for the electric potential now lives inside `ElectricPotentialArray` (both coordinate systems) rather than duplicated at each `Simulation.jl` call site.

### `CPU_innerloop.jl`
`r0_handling_depletion_handling` (Cylindrical) read an unwritten "ghost" neighbour at the r=0 axis that silently defaults to 0, acting as an implicit depletion floor regardless of gauge. Now duplicates the r-right (radially outward) neighbour into that slot instead — a value that moves with the actual solution rather than a fixed external constant.

### `convergence.jl`
Purely diagnostic, no solved values change: labels *why* the SOR loop exited (`:converged` / `:stagnated` / `:max_iterations`) and warns once (`maxlog=1`, separate `_id`s per cause) when stagnation is due to near-threshold depletion bistability vs. hitting floating-point precision. Near-threshold points can genuinely flip classification between iterations, so tight-tolerance convergence is not always reachable, by design.

### `PointTypes.jl`
`is_depleted` required `bulk_bit` (all 26 neighbours also `pn_junction_bit`), which structurally excludes any point one grid-step from a contact — exactly where a shrinking undepleted sliver ends up right before true full depletion. Switched to `pn_junction_bit & update_bit`, with no such neighbour requirement — the same pair `get_active_volume` already uses elsewhere in this file for the equivalent question.

### Tests
- **`test_gauge_invariance.jl`** (new): Validates fixes against closed-form analytical depletion voltages across ten testsets covering planar slabs, annular coax, and cylindrical geometries with various boundary conditions — including direct pointwise field-comparison checks (`epotΔ ≈ epot0 + Δ`) under both `depletion_handling` values, which probe the mean-seed fix directly rather than only through `is_depleted`.
- **`test_depletion.jl`**: Updated `r0_handling_depletion_handling` unit test to corrected output. Tolerances were initially loosened to account for observed near-threshold noise, but that noise was subsequently root-caused to `BEGe_01.yaml`'s `:infinite` boundaries sitting too close to the crystal (a 20mm buffer around a 40mm crystal).
- **`test_real_detectors.jl`**: Inverted Coax's post-depletion margin widened (0.5% → 15%) since the r0-axis fix now correctly detects a small undepleted region near the axis that used to be missed, raising the true depletion voltage. Hexagon's depletion-voltage check replaced with an is_depleted-based bracket: confirmed undepleted at -5.0V and depleted at -11.0V.
- **`test_refine_existing_potential.jl`**: grid-size assertions updated.

### Gauge-invariance sweep, Cylindrical (InvertedCoax)
T = Float64 refinement = [0.2, 0.1, 0.05]
:infinite boundary condition
| U (V), is_depleted | unfixed orig/swap | r0-only orig/swap | all-fixes orig/swap | unfixed match | r0-only match | all-fixes match |
|---|---|---|---|---|---|---|
| 1866.0 | false/false | false/false | false/false | ✅ | ✅ | ✅ |
| 1867.0 | true/false | false/false | false/false | ❌ | ✅ | ✅ |
| 2014.0 | true/false | false/false | false/false | ❌ | ✅ | ✅ |
| 2015.01 | true/false | false/false | true/true | ❌ | ✅ | ✅ |
| 2015.76 | true/false | true/false | true/true | ❌ | ❌ | ✅ |
| 2016.0 | true/false | true/false | true/true | ❌ | ❌ | ✅ |
| 2018.0 | true/false | false/true | true/true | ❌ | ❌ | ✅ |
|  2020.0 | true/false | true/true | true/true | ❌ | ✅ | ✅ |
| 2022–2050V | true/false | true/true | true/true | ❌ | ✅ | ✅ |

All-fixes matches ground truth at every single voltage tested, including exactly at and above the true threshold (2015.76V–2018V), where r0-only alone still leaves 3 gauge-dependent mismatches — direct evidence that the `:infinite`-boundary and interior-seed fixes are independently necessary, not redundant with the r0-axis fix.

| U (V), is_depleted | all-fixes orig/swap (seed mean, :fixed) |  all-fixes orig/swap (seed mean, :reflecting) |
|---|---|---|
| 2012.0 | false/false | false/false |
| 2014.0 | true/true | true/true |
| 2016.0 | true/true | true/true |
| 2018.0 | true/true | true/true |
| 2020.0 | true/true | true/true |

| U (V), is_depleted | unfixed  orig/swap (:fixed) |  unfixed orig/swap (:reflecting) |
|---|---|---|
| 1866.0 | false/false | false/false |
| 1867.0 | false/false | true/false |
| 1868.0 | false/false | true/false |
| 2014.0 | true/false | true/false |
| 2016.0 | true/false | true/false |

### Gauge-invariance sweep, Cartesian (InvertedCoax)

No r0-only column here: Cartesian grids never dispatch through `r0_handling_depletion_handling` at all, so that fix is structurally irrelevant to this case — only the `:infinite`-boundary and interior-seed fixes can do anything for Cartesian.
T = Float64 refinement = [0.2, 0.1, 0.05]
:infinite boundary condition
| U (V), is_depleted | unfixed orig/swap | all-fixes orig/swap (seed min) | all-fixes orig/swap (seed mean) | unfixed match | all-fixes match |
|---|---|---|---|---|---|
| 2025.51 | false/false | false/false | false/false | ✅ | ✅ |
| 2026.0 | false/false | false/false | false/false | ✅ | ✅ |
| 2026.12 | true/false | true/true | false/false | ❌ | ✅ |
| 2028.0 | true/false | true/true | false/false | ❌ | ✅ |
| 2029.0 | true/false | true/true | true/true | ❌ | ✅ |

All-fixes matches at every point. Confirms the Cylindrical result generalises: on a coordinate system where the r0 fix contributes nothing at all, the remaining two fixes alone are sufficient to fully restore gauge invariance here.

| U (V), is_depleted | all-fixes orig/swap (seed mean, :fixed) |  all-fixes orig/swap (seed mean, :reflecting) |
|---|---|---|
| 2020.0 | false/false | false/false |
| 2022.0 | false/false | false/false |
| 2023.0 | false/false | false/false |
| 2024.0 | false/false | true/true |
| 2025.0 | true/true | true/true |

Holding the algorithm fixed and varying only the boundary type shows the near-threshold transition point itself moving by several volts: `:reflecting` transitions around 2022-2024V, `:fixed` around 2024-2025V, and `:infinite` (table above) around 2028-2029V.

| U (V), is_depleted | unfixed  orig/swap (:fixed) |  unfixed orig/swap (:reflecting) |
|---|---|---|
| 2020.0 | true/false | false/false |
| 2022.0 | true/false | false/false |
| 2023.0 | true/false | true/false |
| 2024.0 | true/false | true/true |
| 2025.0 | true/false | true/true |

### Open questions / still needs verification

**For detectors using only `:infinite` boundaries, `estimate_depletion_voltage`'s numeric answer is sensitive to how much separation the grid gives the boundary from the crystal.** Confirmed directly: enlarging `BEGe_01.yaml`'s grid (20mm→300mm buffer) reduced a comparable instability from ~50V down to under 1V. This is still an open, structural limitation of `:infinite` (see the Faraday-cage test in `test_gauge_invariance.jl`) — not something a code fix removes, since the correct open-boundary reference is geometry/capacitance-dependent, not derivable from contact potentials alone. Giving `:infinite` enough separation mitigates it; nothing eliminates it outright.

### DIFF
--- a/src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
+++ b/src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
@@ -169,7 +169,19 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                                                                 (grid_boundary_factor_z_left, grid_boundary_factor_z_right))
         end
 
-        contact_bias_voltages::Vector{T} = if length(det.contacts) > 0 
+        # Weighting-potential solves don't use the detector's real bias
+        # voltages at all: by convention the target contact is fixed at 1
+        # and every other contact at 0. contact_bias_voltages must reflect
+        # that 0/1 convention here, not the real contact.potential values.
+        # Otherwise minimum_applied_potential/maximum_applied_potential
+        # used below for the interior seed and for the :infinite boundary's
+        # gauge-consistency shift, would be computed from voltages that
+        # have nothing to do with this solve. Mirrors the
+        # is_weighting_potential guard convergence.jl already uses for its
+        # own c_limit.
+        contact_bias_voltages::Vector{T} = if is_weighting_potential
+            T[0, 1]
+        elseif length(det.contacts) > 0
             T[contact.potential for contact in det.contacts]
         else
             T[0]
@@ -293,7 +305,14 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
             end
         end
 
-        potential = ismissing(potential_array) ? zeros(T, size(grid)...) : potential_array
+        # Seed the interior at minimum_applied_potential, not a hardcoded 0.
+        # A literal-0 start isn't gauge invariant -- it sits at a different
+        # point relative to the contacts depending on which one is
+        # grounded. Since depletion handling's clamp is bistable (issue
+        # #618), starting two gauge-equivalent solves from
+        # non-corresponding states can make them converge to different
+        # fixed points, even with a shift-invariant update rule.
+        potential = ismissing(potential_array) ? fill(minimum_applied_potential, size(grid)...) : potential_array
         point_types = ones(PointType, size(grid)...)
         set_point_types_and_fixed_potentials!( point_types, potential, grid, det, 
                 weighting_potential_contact_id = weighting_potential_contact_id,
--- a/src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
+++ b/src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
@@ -169,26 +169,27 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
                                                                 (grid_boundary_factor_z_left, grid_boundary_factor_z_right))
         end
 
-        # Weighting-potential solves don't use the detector's real bias
-        # voltages at all: by convention the target contact is fixed at 1
-        # and every other contact at 0. contact_bias_voltages must reflect
+        # Weighting-potential solves don't use the detector's real contact
+        # potentials at all: by convention the target contact is fixed at 1
+        # and every other contact at 0. contact_potentials must reflect
         # that 0/1 convention here, not the real contact.potential values.
-        # Otherwise minimum_applied_potential/maximum_applied_potential
-        # used below for the interior seed and for the :infinite boundary's
-        # gauge-consistency shift, would be computed from voltages that
-        # have nothing to do with this solve. Mirrors the
-        # is_weighting_potential guard convergence.jl already uses for its
-        # own c_limit.
-        contact_bias_voltages::Vector{T} = if is_weighting_potential
+        # Otherwise minimum_applied_potential/maximum_applied_potential/
+        # gauge_ref_potential used below for the interior seed and for the
+        # :infinite boundary's gauge-consistency shift, would be computed
+        # from voltages that have nothing to do with this solve. Mirrors
+        # the is_weighting_potential guard convergence.jl already uses for
+        # its own c_limit.
+        contact_potentials::Vector{T} = if is_weighting_potential
             T[0, 1]
         elseif length(det.contacts) > 0
             T[contact.potential for contact in det.contacts]
         else
             T[0]
         end
-        minimum_applied_potential::T = minimum(contact_bias_voltages)
-        maximum_applied_potential::T = maximum(contact_bias_voltages)
+        minimum_applied_potential::T = minimum(contact_potentials)
+        maximum_applied_potential::T = maximum(contact_potentials)
         bias_voltage::T = maximum_applied_potential - minimum_applied_potential
+        gauge_ref_potential::T = is_weighting_potential ? zero(T) : sum(contact_potentials) / length(contact_potentials)
         sor_consts = [sor_consts]
 
         medium_ϵ_r::T = medium.ϵ_r
@@ -305,20 +306,25 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
             end
         end
 
-        # Seed the interior at minimum_applied_potential, not a hardcoded 0.
+        # Seed the interior at gauge_ref_potential, not a hardcoded 0.
         # A literal-0 start isn't gauge invariant -- it sits at a different
         # point relative to the contacts depending on which one is
         # grounded. Since depletion handling's clamp is bistable (issue
         # #618), starting two gauge-equivalent solves from
         # non-corresponding states can make them converge to different
-        # fixed points, even with a shift-invariant update rule.
-        potential = ismissing(potential_array) ? fill(minimum_applied_potential, size(grid)...) : potential_array
+        # fixed points, even with a shift- and sign-flip update rule.
+        potential = ismissing(potential_array) ? fill(gauge_ref_potential, size(grid)...) : copy(potential_array)
         point_types = ones(PointType, size(grid)...)
-        set_point_types_and_fixed_potentials!( point_types, potential, grid, det, 
+        set_point_types_and_fixed_potentials!( point_types, potential, grid, det,
                 weighting_potential_contact_id = weighting_potential_contact_id,
                 use_nthreads = use_nthreads,
-                not_only_paint_contacts = Val(not_only_paint_contacts), 
+                not_only_paint_contacts = Val(not_only_paint_contacts),
                 paint_contacts = Val(paint_contacts)  )
+        # Shift the whole problem into the V_ref = 0 frame once, here, so the
+        # SOR (and its `:infinite` boundary decay) can run at its
+        # original, pre-gauge-fix literal-0 target instead of re-deriving a
+        # shifted frame on every iteration.
+        potential .-= gauge_ref_potential
         rbpotential = RBExtBy2Array( potential, grid )
         rbpoint_types = RBExtBy2Array( point_types, grid )
     end # @inbounds
@@ -337,6 +343,7 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
         bias_voltage,
         maximum_applied_potential,
         minimum_applied_potential,
+        gauge_ref_potential,
         grid_boundary_factors
     )
     return pcs
--- a/src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
+++ b/src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
@@ -364,7 +364,7 @@ function ElectricPotentialArray(pcs::PotentialCalculationSetup{T, Cartesian,  3,
             end
         end
     end
-    return pot
+    return pot .+ pcs.gauge_ref_potential
 end
 
 function ImpurityScaleArray(pcs::PotentialCalculationSetup{T, Cartesian, 3, Array{T, 3}})::Array{T, 3} where {T}
--- a/src/ScalarPotentials/PointTypes.jl
+++ b/src/ScalarPotentials/PointTypes.jl
@@ -116,7 +116,15 @@ is_depleted(sim.point_types)
 """
 function is_depleted(point_types::PointTypes)::Bool
     return if has_depletion_handling(point_types)
-        !any(b -> (bulk_bit & b > 0) && (undepleted_bit & b > 0) &&
+        # Checks pn_junction_bit & update_bit, not bulk_bit. bulk_bit only
+        # gets set when ALL 26 neighbors of a point are also pn_junction --
+        # so any point one grid step from a contact can never count as
+        # "bulk", even if it's genuinely undepleted. That's exactly where an
+        # undepleted sliver ends up right before full depletion: as bias
+        # approaches V_d, the undepleted region shrinks toward whichever
+        # contact it's pinching off against. So checking bulk_bit made
+        # is_depleted blind to tese cases.
+        !any(b -> (pn_junction_bit & b > 0) && (update_bit & b > 0) && (undepleted_bit & b > 0) &&
         (inactive_layer_bit & b == 0), point_types.data)
     else 
         @warn """Electric potential calculation was not run with depletion handling enabled. 
--- a/src/Simulation/Depletion.jl
+++ b/src/Simulation/Depletion.jl
@@ -135,8 +135,7 @@ function estimate_depletion_voltage(sim::Simulation{T},
     end
     bulk = findall((sim.point_types.data .& bulk_bit .> 0) .& (sim.point_types.data .& inactive_layer_bit .== 0))
     U_min_max = filter(in(Urng), _find_depletion_voltage_candidates(ϕρ, ϕV, bulk))
-    U2 = isempty(U_min_max) ? U : only(U_min_max)    
-    U = U > 0 ? max(U, U2) : min(U, U2)
+    U = U > 0 ? max(U, U_min_max...) : min(U, U_min_max...)
     if verbose
         @info "The depletion voltage is around $(round(U, digits = Int(ceil(-log10(tol))))) ± $(tol) $(internal_voltage_unit) applied to contact $(contact_id)."
     end
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -502,8 +502,7 @@ function apply_initial_state!(sim::Simulation{T, CS}, ::Type{ElectricPotential},
     sim.q_eff_fix = EffectiveChargeDensity(FixedEffectiveChargeDensityArray(pcs), grid)
     sim.ϵ_r = DielectricDistribution(DielectricDistributionArray(pcs), get_extended_midpoints_grid(grid))
     sim.point_types = PointTypes(PointTypeArray(pcs), grid, depletion_handling)
-    # pcs.potential lives in the V_ref = 0 (gauge-shifted) frame; shift back to real potentials once, here.
-    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs) .+ pcs.gauge_ref_potential, grid)
+    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs), grid)
     nothing
 end
 
@@ -636,9 +635,7 @@ function update_till_convergence!( sim::Simulation{T,CS},
     sim.imp_scale = ImpurityScale(ImpurityScaleArray(pcs), grid)
     sim.q_eff_fix = EffectiveChargeDensity(FixedEffectiveChargeDensityArray(pcs), grid)
     sim.ϵ_r = DielectricDistribution(DielectricDistributionArray(pcs), get_extended_midpoints_grid(grid))
-    # pcs.potential lives in the V_ref = 0 (gauge-shifted) frame for the
-    # whole SOR run; shift back to real potentials exactly once, here, after convergence.
-    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs) .+ pcs.gauge_ref_potential, grid)
+    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs), grid)
     sim.point_types = PointTypes(PointTypeArray(pcs), grid, depletion_handling)
     cf
 end
--- a/src/Simulation/Simulation.jl
+++ b/src/Simulation/Simulation.jl
@@ -502,7 +502,8 @@ function apply_initial_state!(sim::Simulation{T, CS}, ::Type{ElectricPotential},
     sim.q_eff_fix = EffectiveChargeDensity(FixedEffectiveChargeDensityArray(pcs), grid)
     sim.ϵ_r = DielectricDistribution(DielectricDistributionArray(pcs), get_extended_midpoints_grid(grid))
     sim.point_types = PointTypes(PointTypeArray(pcs), grid, depletion_handling)
-    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs), grid)
+    # pcs.potential lives in the V_ref = 0 (gauge-shifted) frame; shift back to real potentials once, here.
+    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs) .+ pcs.gauge_ref_potential, grid)
     nothing
 end
 
@@ -635,7 +636,9 @@ function update_till_convergence!( sim::Simulation{T,CS},
     sim.imp_scale = ImpurityScale(ImpurityScaleArray(pcs), grid)
     sim.q_eff_fix = EffectiveChargeDensity(FixedEffectiveChargeDensityArray(pcs), grid)
     sim.ϵ_r = DielectricDistribution(DielectricDistributionArray(pcs), get_extended_midpoints_grid(grid))
-    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs), grid)
+    # pcs.potential lives in the V_ref = 0 (gauge-shifted) frame for the
+    # whole SOR run; shift back to real potentials exactly once, here, after convergence.
+    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs) .+ pcs.gauge_ref_potential, grid)
     sim.point_types = PointTypes(PointTypeArray(pcs), grid, depletion_handling)
     cf
 end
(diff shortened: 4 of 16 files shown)


fhagemann:
Without having had a close look at the code, but looking at the description:
- **How do we find a good `V_ref`?**
  This seems to address gauge-invariance by absolute shifts in the electric potential, but what about if we were to flip the signs of all contact potentials and impurity densities?
  Assume we take the ICPC example detector, flip the sign of all of its bias voltage and convert it from p-type to n-type by flipping also the sign of the impurity density, I would expect to get different values for the depletion voltage (not only in the sign), because `V_ref` will be set to the mantle contact in one case, and to the point contact in the other case.

flip

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -293,7 +305,14 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
...
 
-        potential = ismissing(potential_array) ? zeros(T, size(grid)...) : potential_array
+        # Seed the interior at minimum_applied_potential, not a hardcoded 0.
+        # A literal-0 start isn't gauge invariant -- it sits at a different
+        # point relative to the contacts depending on which one is
+        # grounded. Since depletion handling's clamp is bistable (issue
+        # #618), starting two gauge-equivalent solves from
+        # non-corresponding states can make them converge to different
+        # fixed points, even with a shift-invariant update rule.
+        potential = ismissing(potential_array) ? fill(minimum_applied_potential, size(grid)...) : potential_array
```
fhagemann: I'm not sure if this is what we want because `minimum` is also not invariant with respect to sign-flips.

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
```diff
@@ -169,7 +169,19 @@ function PotentialCalculationSetup(det::SolidStateDetector{T}, grid::CartesianGr
...
+        # and every other contact at 0. contact_bias_voltages must reflect
+        # that 0/1 convention here, not the real contact.potential values.
+        # Otherwise minimum_applied_potential/maximum_applied_potential
+        # used below for the interior seed and for the :infinite boundary's
+        # gauge-consistency shift, would be computed from voltages that
+        # have nothing to do with this solve. Mirrors the
+        # is_weighting_potential guard convergence.jl already uses for its
+        # own c_limit.
+        contact_bias_voltages::Vector{T} = if is_weighting_potential
+            T[0, 1]
```
fhagemann: Then, we might want to rename this to `contact_potentials` (instead of `contact_bias_voltages`) if we are also using this for weighting potentials.

src/ScalarPotentials/PointTypes.jl
```diff
@@ -116,7 +116,15 @@ is_depleted(sim.point_types)
...
-        !any(b -> (bulk_bit & b > 0) && (undepleted_bit & b > 0) &&
+        # Checks pn_junction_bit & update_bit, not bulk_bit. bulk_bit only
+        # gets set when ALL 26 neighbors of a point are also pn_junction --
+        # so any point one grid step from a contact can never count as
+        # "bulk", even if it's genuinely undepleted. That's exactly where an
+        # undepleted sliver ends up right before full depletion: as bias
+        # approaches V_d, the undepleted region shrinks toward whichever
+        # contact it's pinching off against. So checking bulk_bit made
+        # is_depleted blind to tese cases.
+        !any(b -> (pn_junction_bit & b > 0) && (update_bit & b > 0) && (undepleted_bit & b > 0) &&
```
fhagemann, line 127: Back in the days (#407), we included the `bulk_bit` check because we were getting individual grid points very close to the surface that would be wrongly marked as undepleted. How can we avoid those messing up the `is_depleted` result?

test/test_gauge_invariance.jl
fhagemann, line 1: I would also like to see some tests where all signs are swapped (bias voltage, impurity density) to cross-check if the electrical potential is also sign-flipped.

test/test_gauge_invariance.jl
```diff
@@ -0,0 +1,322 @@
+using SolidStateDetectors
+using Test
```
fhagemann:
```suggestion
# This file is a part of SolidStateDetectors.jl, licensed under the MIT License (MIT).

using Test
using SolidStateDetectors
```

test/test_gauge_invariance.jl
```diff
@@ -0,0 +1,322 @@
...
+    )
+
+    function electric_potential_planar_inf(U::Real, ground_offset::Real)
+        sim = Simulation{T}(deepcopy(cfg))
+        det = sim.detector
+        det = SolidStateDetector(det, contact_id = 1, contact_potential = T(ground_offset))
+        det = SolidStateDetector(det, contact_id = 2, contact_potential = T(U + ground_offset))
+        sim.detector = det
+        apply_initial_state!(sim, ElectricPotential, Grid(sim))
+        calculate_electric_potential!(sim, depletion_handling = false, refinement_limits = refinement_limits, use_nthreads = 4, verbose = false)
```
fhagemann: Read in `test_utils.jl` at the beginning and used the `timed_` equivalent functions here (and elsewhere in this file).

test/test_real_detectors.jl
```diff
@@ -31,8 +31,14 @@ T = Float32
...
-    # Check wether detector is depleted, 10V above the previously calculated depletion voltage
-    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = ustrip(deplV + deplV*0.005))
+    # Check whether the detector is depleted. `deplV` is a quick linear
+    # estimate that assumes uniform, full impurity density. The real
+    # threshold, with depletion_handling on, sits noticeably higher: the
+    # r0_handling_depletion_handling fix now correctly detects a small
+    # undepleted region near the r=0 axis that used to be missed, which
+    # raises the true depletion voltage. So the old ~0.5% margin above
+    # `deplV` is no longer enough -- 15% above `deplV` is confirmed depleted.
+    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = ustrip(deplV * 1.15))
```
fhagemann, line 41: I'm not sure i like this... In an ideal world, `estimate_depletion_voltage` should give a good estimate and not something that is 15% off..

test/test_real_detectors.jl
```diff
@@ -149,7 +155,20 @@ end
     signalsum *= inv(ustrip(SolidStateDetectors._convert_internal_energy_to_external_charge(sim.detector.semiconductor.material)))
     @info signalsum
     @test isapprox( signalsum, T(2), atol = 5e-3 )
-    @test isapprox(timed_estimate_depletion_voltage(sim, verbose = false, check_for_depletion = false), T(-13.15)*u"V", atol = 1.0u"V") 
+    # estimate_depletion_voltage's only(...) call assumes a single filtered
```
fhagemann: Why are we getting two candidates?
I can see this happening if there were one positive and one negative candidate, but for two negative candidates, we should always return the more extreme one..

test/test_depletion.jl
```diff
@@ -47,20 +62,25 @@ T = Float32
...
-    @test abs(estimate_depletion_voltage(sim, check_for_depletion = false, verbose = false) - dep_sim) < 5u"V"
+    @test abs(estimate_depletion_voltage(sim, check_for_depletion = false, verbose = false) - dep_sim) < 75u"V"
 
     # Finally, compare to fresh simulation which is changed manually
     sim_fresh = Simulation{T}(joinpath(@__DIR__, "test_config_files/BEGe_01.yaml"))
     sim_fresh.detector = SolidStateDetector(sim_fresh.detector, contact_id = id, contact_potential = bias_target)
     sim_fresh.detector = SolidStateDetector(sim_fresh.detector, sim.detector.semiconductor.impurity_density_model)
     timed_calculate_electric_potential!(sim_fresh, refinement_limits = 0.01, depletion_handling = true)
-    @test abs(estimate_depletion_voltage(sim_fresh, check_for_depletion = false, verbose = false) - dep_sim) < 5u"V"
+    @test abs(estimate_depletion_voltage(sim_fresh, check_for_depletion = false, verbose = false) - dep_sim) < 75u"V"
```
fhagemann: I'm also not sure I like the fact that the values are now 75V apart instead of 5V..

claudiaalvgar: I don't like this either, that's why I wanted to keep this as a draft. For Cartesian, unfixed code's gauges agree again above ~2030V too — not sure if that's a fluke or expected (agreement at high bias may just mean we've passed both gauges' thresholds, regardless of correctness — same pattern as r0-only Cylindrical above 2020V). I'll dig into the rest so we can reach a conclusion of whether this is a change we want to have or not

fhagemann:
Thanks!
Maybe we can also find an analytical example that's a bit more complicated than the infinite plate capacitor or inifinite coaxial detector, that might allow us to test what's actually correct for pinch-off or close-to-contact undepletion, to be included in the decision which solution we want to go for.

src/ScalarPotentials/PointTypes.jl
hervasa2: In theory we could leave it as is as long as the grid size close to the surface is small enough. Sure it may lag the true depletion voltage by a bit, but would be more robust. Maybe we can quantify this effect with a reasonable grid size? I think for explorations of the Li profile and depletion voltage the grid size should be small enough by construction?

fhagemann: I'm not sure we have to worry about the Li profile, because that part should be covered by the `inactive_layer_bit` part in the next line. I think this becomes a more relevant issue for coarse grids.

test/test_real_detectors.jl
fhagemann: Should be fixed with #623, but I would still be interested why the changes in this PR result in two candidates instead of just one. Is this related to removing the `bulk_bit` check from `is_depleted`?

src/PotentialCalculation/PotentialCalculationSetup/PotentialCalculationSetupCartesian3D.jl
claudiaalvgar: I fixed: the interior seed now uses gauge_ref_potential = mean(contact potentials), which is odd under sign flip (unlike min), and directly verified in tests. One thing to take into account is: :infinite's own boundary decay ended up staying on min, not mean because mean measurably corrupts the near-boundary field for realistic single-bias-direction detectors (confirmed on the isochrone-style geometry). So :infinite's truncation itself is not sign-flip-invariant but the cryostat test added shows that's an unavoidable property of any boundary approximation

test/test_gauge_invariance.jl
claudiaalvgar: I added new tests for this in the new commits

test/test_real_detectors.jl
claudiaalvgar: I think what's happening is that at deplV*1.005 (the old 0.5% margin), the detector is not depleted with the current fixes. This isn't a numerical issue in ```estimate_depletion_voltage```, it's that deplV is an approximate linear estimate, while the true depletion threshold with ```depletion_handling``` on sits measurably higher once ```r0_handling``` correctly catches a previously-missed undepleted sliver near the axis

src/Simulation/Depletion.jl
```diff
@@ -135,8 +135,7 @@ function estimate_depletion_voltage(sim::Simulation{T},
     end
     bulk = findall((sim.point_types.data .& bulk_bit .> 0) .& (sim.point_types.data .& inactive_layer_bit .== 0))
```
fhagemann, line 136: `estimate_depletion_voltage` still seems to be restricted to `bulk_bits`, whereas this restriction was removed in `is_depleted`.
Could this be why the results of `estimate_depletion_voltage` and manual `is_depleted` checks used to be only ~10V apart and are now 75V apart.

Can we adjust `estimate_depletion_voltage` to give more realistic estimates, e.g. by dealing differently with `bulk_bits`?

src/Simulation/Simulation.jl
```diff
@@ -502,7 +502,8 @@ function apply_initial_state!(sim::Simulation{T, CS}, ::Type{ElectricPotential},
     sim.q_eff_fix = EffectiveChargeDensity(FixedEffectiveChargeDensityArray(pcs), grid)
     sim.ϵ_r = DielectricDistribution(DielectricDistributionArray(pcs), get_extended_midpoints_grid(grid))
     sim.point_types = PointTypes(PointTypeArray(pcs), grid, depletion_handling)
-    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs), grid)
+    # pcs.potential lives in the V_ref = 0 (gauge-shifted) frame; shift back to real potentials once, here.
+    sim.electric_potential = ElectricPotential(ElectricPotentialArray(pcs) .+ pcs.gauge_ref_potential, grid)
```
fhagemann: Maybe this is something to add in `ElectricPotentialArray` instead of here?

fhagemann:
```julia
function ElectricPotentialArray(pcs::PotentialCalculationSetup{T, Cylindrical, 3, Array{T, 3}})::Array{T, 3} where {T}
    pot::Array{T, 3} = Array{T, 3}(undef, size(pcs.grid))
    for iz in axes(pot, 3)
        irbz::Int = rbidx(iz)
        for iφ in axes(pot, 2)
            irbφ::Int = iφ + 1
            idxsum::Int = iz + iφ
            for ir in axes(pot, 1)
                irbr::Int = ir + 1
                rbi::Int = iseven(idxsum + ir) ? rb_even::Int : rb_odd::Int
                pot[ir, iφ, iz] = pcs.potential[ irbz, irbφ, irbr, rbi ]
            end
        end
    end
    for iz in axes(pot, 3)
        p_r0 = mean(pot[1,:,iz])
        pot[1,:,iz] .= p_r0
    end
    # pcs.potential lives in the V_ref = 0 (gauge-shifted) frame; shift back to real potentials once, here.
    return pot .+ pcs.gauge_ref_potential
end

File: src/Simulation/Depletion.jl (hunk 9 above)
Comment (claudiaalvgar) [reply src/Simulation/Depletion.jl:136]: I checked this and I think this is not the issue. ```BEGe_01``` :infinite boundaries only had a 20mm buffer beyond the 40mm crystal. The Faraday-cage/sign-flip tests in ```test_gauge_invariance.jl``` show that  :infinite's truncation-edge behaviour is only accurate once it has real separation from the region of interest. So the fix in my commits was to give ```BEGe_01.yaml``` the separation :infinite needs and all four comparisons now pass at 10V

File: test/test_real_detectors.jl (hunk 6 above)
Comment (fhagemann) [inline test/test_real_detectors.jl:41]: Can we somehow address to make `estimate_depletion_voltage` better in this sense? 15% tolerance is pretty bad in my opinion..

File: test/test_real_detectors.jl [hunk 11]
```diff
@@ -149,7 +155,15 @@ end
...
-    @test isapprox(timed_estimate_depletion_voltage(sim, verbose = false, check_for_depletion = false), T(-13.15)*u"V", atol = 1.0u"V") 
+    @test (timed_estimate_depletion_voltage(sim, verbose = false, check_for_depletion = false); true)
+    id = SolidStateDetectors.determine_bias_voltage_contact_id(sim.detector)
+    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = T(-5.0))
+    timed_calculate_electric_potential!(sim, depletion_handling = true)
+    @test !is_depleted(sim.point_types)
+
+    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = T(-11.0))
+    timed_calculate_electric_potential!(sim, depletion_handling = true)
+    @test is_depleted(sim.point_types)
```
fhagemann, line 166: Also, what used to be a depletion voltage of `-13.15V` is now between `-5V` and `-11V`. Do we understand this, and does this make sense?

test/test_config_files/BEGe_01.yaml
```diff
@@ -17,8 +17,8 @@ grid:
         left: periodic
         right: periodic
     z:
-      from: -20
-      to: 60
+      from: -300
+      to: 300
```
fhagemann, line 21: I know that these changes were done to move the `:infinite` boundary further away from the crystal. But is there any way that we can also get a reasonable boundary condition without having to extend the world this much?

fhagemann (commented):
Commenting on the changes on the tests first, to make sure that this is what we want.
We might also want to include the findings in the new PR #639 here.

test/test_real_detectors.jl
fhagemann: @claudiaalvgar could you check if with the updates to `estimate_depletion_voltage` in PR #639, the two methods agree again (and we don't have to increase the value from `estimate_depletion_voltage` by 15% to make this work).

claudiaalvgar:
I ran the combined PR634+639 (julia version 1.11.5): on the real-detector config, Float32 now gives 2030.11 V against Float64's 2030.006 V — so PR639's fix holds fully when combined with PR634, and the old 15% safety margin isn't needed anymore; on the BEGe config, though, the U_est/U_alt comparison in test_depletion.jl is unmoved by adding PR639 — both still fail the <5V check, but this failure would also be true if we don't consider PR634 nor PR639 and use Float64 instead of Float32. I put the results on the table attached for future reference
Screenshot 2026-09-17 at 14 58 27

test/test_real_detectors.jl
claudiaalvgar: -13.15V came from the previous code, where ```:infinite``` always decayed toward a hardcoded 0 on both the real solve and the weighting-potential solve used in ```estimate_depletion_voltage``` so no mismatch. This PR's gauge-invariance change makes the real solve's ```:infinite``` decay target depend on the bias sign, while the weighting potential's stays fixed at 0. For this detector's negative bias, that mismatch corrupts ```estimate_depletion_voltage``` to ≈-70 to -75V (in both Float32 and Float64). ```:reflecting```'s boundary formula is a plain mirror-copy of the neighbouring grid cell with no reference-potential term at all, so  running this same code with ```:reflecting``` instead gives ≈-12.16V/-12.19V, close to the old -13.15V

test/test_config_files/BEGe_01.yaml
claudiaalvgar: The enlargement here is compensating for the same weighting-potential mismatch inside ```adjust_impurity_and_electric_potential_to_match_depletion!/adjust_bias_and_electric_potential!```, which pushes dep_sim ≈30V off target on that small world. For now, the validated fix could be using ```:reflecting``` : no reference-potential term, so it can't develop this mismatch

claudiaalvgar:
Update: I rebased this branch since the changes were needed for these fixes to apply cleanly. PR634 added ```gauge_ref_potential``` (a mean-based SOR interior seed) for gauge-consistent convergence, but before the last commits, left the :infinite boundary itself decaying toward ```minimum_applied_potential``` unconditionally — a reference that's exactly translation-invariant but sign-convention dependent, so for detectors where the swept contact happens to be negative (BEGe_01, HexagonalPrism) the boundary decays toward the wrong value and the depletion-voltage estimate becomes unusable (tens of volts off, or a failed bisection); enlarging the :infinite world was tried first since it also masks the symptom (pushing the truncation edge far enough away dilutes the wrong reference's error), but it doesn't fix the actual cause, only hides it for whatever world size happens to be tested, and would have to be re-applied to every current and future detector config hitting this pattern rather than being fixed once. With the new commits, ```boundary_ref_potential``` decays toward 0 whenever exactly one contact is non-zero, the physically correct grounded reference, matching the previous convention — and falls back to the ```old minimum_applied_potential``` otherwise, so nothing regresses where that convention doesn't hold. The one cost, specific to the :infinite boundary, is that this value-based rule can't be made exactly translation-invariant under an artificial relabeling shift (Δ = -U in ```test_gauge_invariance.jl```, which swaps which contact reads 0).
Together with ```gauge_ref_potential```, which stops the nonlinear depletion clamp from landing on a different answer for what should be the same physical setup, no matter what boundary type is used, PR634 as a whole now gives correct, consistent depletion-voltage estimates for detectors that used to either fail outright or quietly return the wrong number

PR #636, opened 2026-09-09
Replace `CITATION.bib` with `CITATION.cff`

fhagemann:
Follow-up of #633:

This is replacing the CURRENT `CITATION.bib` with a more generic `CITATION.bib`:
- with the [initial JINST publication](https://doi.org/10.1088/1748-0221/16/08/P08007), describing the state of the software in v0.5 as `preferred-citation`, and
- the [EPJC publication on reduced charge-collection layers](https://doi.org/10.1140/epjc/s10052-026-15508-3) as additional `reference`.

Citation metadata are valid according to schema version 1.2.0 (tested using `cffconvert --validate`).


fhagemann:
I received the ORCID identifiers from everyone, this should be complete now.

PR #637, opened 2026-09-10
Update Downgrade workflow after julia-1.13 release

fhagemann:
With the release of julia-1.13.0, the Downgrade tests seem to fail because the [lowest allowed compat for `Unitful`](https://github.com/JuliaRegistries/General/blob/master/U/Unitful/Compat.toml) for julia-1.13 (v1.12.3) does not compile under julia-1.13 (see https://github.com/JuliaPhysics/Unitful.jl/issues/815).

Temporary fix: fix the `Unitful` compat for Downgrade tests on `1` manually to 1.25.1, which is the lowest version that compiles under julia-1.13. 

This can be removed once things are sorted out, either in julia-1.13 itself, or in the `Compat.toml` of `Unitful` in the General registry.

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: ['1.10', '1']
+        version: ['1.10', '1.12', '1']
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


fhagemann:
Restricting KernelAbstractions  to a minimum of `0.8.1` instead of `0.8.0` to have this fix: 
https://github.com/JuliaGPU/KernelAbstractions.jl/pull/300

PR #639, opened 2026-09-16
Consistent depletion voltage estimate

### LINKED ISSUE #618: Depletion handling: clamp bistability and the hidden 0 V floor
> [!CAUTION]
> **Correction (2026-07-21):** the central ground-truth claim of this issue — that the full-impurity *linear* solution at 1.05 × V_d shows no interior extremum, making the `6d576c13` "undepleted" result a spurious fixed point — is **wrong and retracted**. Direct linear solves (full impurities, `depletion_handling = false`) show the pocket above the point contact is **real ICPC pinch-off**, persisting to ≈ 1.08–1.12 × the estimated V_d; it is also reproduced in Cartesian coordinates. `6d576c13` reports genuine physics, and the current (ghost-anchored) code suppresses it — the ±0.5 % test brackets validate that artifact. The flawed original ground truth came from an adapted-weighting-potential superposition that is not faithful near the pocket. See the [correction comment](https://github.com/JuliaPhysics/SolidStateDetectors.jl/issues/618#issuecomment-5038538907) and discussion below for the corrected analysis and agreed direction (adopt `6d576c13`, re-anchor the depletion tests, initialize depletion handling from the converged linear solution).
>
> The parts of this issue that remain valid: the clamp's plateau fixed points and history dependence, the 0 V-floor fragility (including its gauge non-invariance), and the reformulation proposals.

Proposal generated as part of full AI review of SolidStateDetectors codebase:

From the 2026-07 review on the `improvements` branch; line numbers refer to that branch.
This documents the most consequential *algorithmic* finding of the review: `handle_depletion`
admits stable spurious "undepleted" states, and the thing currently preventing them in the
common cases is an accident — a never-written ghost value acting as a 0 V potential
floor at the cylindrical axis.

## How `handle_depletion` works

`src/PotentialCalculation/SuccessiveOverRelaxation/SOR.jl:57–83`, applied per grid
point inside the SOR update on CPU and GPU. Given the new potential proposal, the
six neighbor potentials, and the point's impurity term
`imp_contribution = q_eff_imp * volume_weight` (negative for p-type):

```julia
vmin = min(neighbor_potentials...);  vmax = max(neighbor_potentials...)
neighbor_relevant_extremum = is_ptype ? vmin : vmax
imp_scale = (neighbor_relevant_extremum - new_potential) / imp_contribution
imp_scale = clamp(imp_scale, 0, 1)
if vmax == vmin imp_scale = 0 end
new_potential += imp_scale * imp_contribution
# overshoot clamp: never cross the neighbor extremum
if (is_ptype && new_potential < vmin) || (!is_ptype && new_potential > vmax)
    new_potential = neighbor_relevant_extremum
end
```

The rationale (docstring, same file): in a depleted detector the potential is
monotone between contacts, so no interior point may become a local extremum. Where
the full impurity term would create one, the impurity density is scaled down
(`imp_scale < 1`) and the point is treated as undepleted; `is_depleted(point_types)`
reports `imp_scale == 1` everywhere.

## The accidental 0 V floor at r = 0

The cylindrical red-black layout is (z, φ, r). For an axis point the
radially-inward neighbor — slot 5 of the neighbor tuple — reads the sub-axis ghost
row, which **no boundary condition ever writes**: it stays at its initial value 0.
`r0_handling_depletion_handling`
(`src/PotentialCalculation/SuccessiveOverRelaxation/CPU_innerloop.jl:236–250`)
keeps that slot in the tuple, so every axis point's extremum check runs against
`min(real neighbors…, 0)` — an implicit 0 V floor (ceiling, for n-type). The
Cartesian variant is a no-op: Cartesian grids have no such anchor anywhere.

This looks exactly like a stale-index bug, and the review first "fixed" it
(commit `6d576c13`: replace slot 5 by the outward neighbor). That fix made the
suite fail in a physically revealing way, which led to the analysis below and a
documented revert (`f87dc4ee`, plus a unit test pinning the tuple behavior in
`test/test_depletion.jl:7`).

## What the failed fix revealed: erosion fixed points

With the ghost masked, the inverted-coax example (`test_real_detectors.jl:26–37`,
estimated depletion voltage ≈ 1871 V) still reported **undepleted at 1.05× its
depletion voltage**. Ground truth says otherwise: because the no-depletion-handling
problem is linear, the true full-impurity solution at any bias is the superposition
`epot(V_b) = V_b · wpot + epot_ρ` (weighting potential of the bias contact, plus
the all-contacts-grounded impurity-only solution). Computing both on the refined
grid and scanning the superposition at 1.05× V_d shows no interior extremum — the
potential minimum sits on the 0 V contact. Depleted. The "undepleted" flag is a
spurious *fixed point of the iteration*, not a property of the physics:

* Wherever the local impurity term exceeds the curvature margin of the
  neighborhood (|imp_contribution| ≳ vmax − vmin), the clamped update degenerates
  to `new_potential = vmin` (p-type). The update becomes a **morphological
  erosion**: each point copies the minimum of its neighbors.
* Erosion's stable states are flat plateaus pinned at the lowest value on the
  region's rim. Flatness then keeps `imp_scale ≈ 0` (and `vmax == vmin` forces
  exactly 0), which keeps the impurity term out of the update, which keeps the
  plateau flat: self-consistent, and it survives arbitrarily many iterations and
  further grid refinement.
* The true solution (imp_scale = 1, curved potential) is *also* a fixed point.
  Which basin the solver lands in depends on history — thin, nearly-flat channels
  (the IVC axis channel above the point contact) get trapped in the spurious one.

The ghost 0 is what breaks this at the axis: `vmin` is 0 there regardless of any
plateau value, so axis points are never pinned to a plateau neighbor, keep
(essentially) their full impurity term, and follow the true Poisson update; correct
values then propagate radially outward and drain axis-adjacent plateaus.

A second attempted fix — forcing `imp_scale = 1` at the axis — cured 1.05× but
broke the opposite test: at 0.995× V_d the detector must report *undepleted*, and
the legitimate signature of that at the axis is precisely "the impurity term would
push the point below the 0 V floor". The floor plays both roles; the current
behavior (revert + in-place documentation) is the only variant of the three that
passes 0.995× → undepleted, 1.005× → depleted, 1.05× → depleted.

## Why the current state is still fragile

1. **The floor hard-codes 0 V.** It is exact only while some contact is grounded
   at 0 V (true for every shipped example). A bias scheme with both contacts away
   from 0 (split bias, shifted reference) makes 0 wrong in both directions:
   legitimate potentials below 0 near a negative contact read as undepletion, and
   the anchor sits at the wrong level for the extremum check.
2. **The anchor exists only on the cylindrical axis.** Cartesian simulations and
   off-axis flat regions (thin gaps between segmented contacts, plateaus formed
   mid-refinement) run the same clamp with no floor at all; the erosion fixed
   points are reachable there. (No failing Cartesian example was constructed
   during the review, but it is the identical code path minus the accident that
   protects the cylindrical axis.)
3. **`vmax == vmin ⇒ imp_scale = 0`** declares exactly-flat neighborhoods
   undepleted, locking constant-initialized regions until rounding noise breaks
   the tie.

## Proposed reformulation

Ordered by effort; 1 is a contained change, 2–3 are the real fix.

1. **Replace the implicit 0 by the applied-potential extrema.** For a depleted
   detector the maximum principle bounds the potential by the applied contact
   potentials, so `floor = minimum(applied potentials)`,
   `ceiling = maximum(applied potentials)` are the correct generalization of the
   ghost's role — and they can be applied uniformly at *every* point (extend the
   extremum check by the bound), not just the axis. This fixes fragility 1,
   removes the cylindrical/Cartesian asymmetry (2), and anchors every channel
   whose erosion would drag it to a bound. It does **not** remove the bistability
   for plateaus strictly inside the bounds.
2. **Make the spurious plateaus non-stationary.** Options that keep the current
   structure: (a) when the overshoot clamp fires, mix a small fraction of the
   *unclamped* full-impurity proposal into the update, so exact plateaus drift
   instead of self-reproducing; (b) periodically re-anneal — reset
   `imp_scale = 1` in flagged regions every N iterations and let the clamp
   re-derive it; genuine undepleted regions re-form, spurious ones converge away;
   (c) after convergence, re-evaluate flagged regions once with full impurities
   and check for an actual interior extremum (the superposition criterion above),
   warning or re-iterating on mismatch.
3. **Longer term:** what is being solved is a free-boundary/complementarity
   problem (either full impurity and the potential respects the bounds, or the
   potential sits on the bound and the charge is reduced). That is the classical
   obstacle problem, and **projected SOR** (Cryer) solves its LCP form with a
   convergence guarantee — a natural fit since the solver is already SOR. This
   would replace the heuristic clamp entirely and make the depleted solution the
   unique attractor.

Any change here alters physics-facing behavior, so it should be validated the way
the review validated the revert: the ±0.5 % depletion-voltage brackets for the
example detectors, plus superposition ground truth on refined grids, plus at least
one case with no 0 V contact and one Cartesian flat-channel case (which today have
no safety net).

## Reproducing the spurious state

*(Edited: the snippet originally posted here was flawed — without weighting
potentials at proper refinement, `estimate_depletion_voltage` returns ≈ 2156 V
instead of ≈ 1871 V and every variant reports depleted at all probe points.
Verified snippet below; see the comments for the full variant matrix.)*

With slot 5 masked in `r0_handling_depletion_handling` (the one-line change of
`6d576c13`), the settings of the "Inverted Coax" testset report **undepleted at
1.005× and even 1.05× V_d**; the current code reports depleted at both:

```julia
using SolidStateDetectors, Unitful
T = Float32
simulate!(sim, convergence_limit = 1e-6, refinement_limits = [0.2, 0.1, 0.05, 0.03, 0.02, 0.01], verbose = false)
deplV = estimate_depletion_voltage(sim, check_for_depletion = false)   # 1871.4 V
id = SolidStateDetectors.determine_bias_voltage_contact_id(sim.detector)
for f in (0.995, 1.005, 1.05)
    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = f * ustrip(deplV))
    calculate_electric_potential!(sim, depletion_handling = true)
    @show f, is_depleted(sim.point_types)
end
```

Relevant history on the branch: `6d576c13` (attempted fix), `f87dc4ee` (revert
with documented semantics), `test/test_depletion.jl:7` (tuple pin),
`SOR.jl:31–56` (docstring incl. the existing fixed-charge caveat, which this
analysis reinforces).

### LINKED ISSUE #640: Result of `estimate_depletion_handling` depends on precision type
When running the tests in `test/test_real_detectors.jl` with `Float64` instead of `Float32`, some tests regarding the depletion state and the depletion voltage fail:

```
julia> include("test/utils.jl"); include("test_real_detectors.jl")
Precompiling SolidStateDetectors...
┌ Info: 	In electric field calculation: Keyword `n_points_in_φ` not set.
└ 		Default is `n_points_in_φ = 36`. 2D field will be extended to 36 points in φ.
[ Info: Detector simulation done
[ Info: 2.0
Convergence:  Time: 0:00:04 (8 iterations)
[ Info: Depletion voltage: 2030.006416947449 V
Inverted Coax: Test Failed at ~/Software/SolidStateDetectors.jl/test/test_real_detectors.jl:28
  Expression: isapprox(deplV, 1871 * u"V", atol = 10.0 * u"V")
   Evaluated: isapprox(2030.006416947449 V, 1871 V; atol = 10.0 V)

... 

Inverted Coax: Test Failed at ~/Software/SolidStateDetectors.jl/test/test_real_detectors.jl:33
  Expression: !(is_depleted(sim.point_types))
```
(Note that these tests both pass using `Float32`)

hervasa2:
## Summary

`estimate_depletion_voltage` now uses the analytic local-extremum scan as the main method and only falls back to bisection when the scan does not give a unique candidate in the search range. Before, bisection always ran first and the scan result was only used to correct it.

The idea behind the scan: the electric potential is linear in the bias voltage,

ϕ(U) = ϕρ + U·ϕV

where ϕρ is the potential from the impurity density alone and ϕV is the weighting potential of the bias contact. For a bulk point x₀ and a neighbour x, the difference

ϕ(x₀) − ϕ(x) = δ1 + U·δ2,   with δ1 = ϕρ(x₀) − ϕρ(x), δ2 = ϕV(x₀) − ϕV(x)

changes sign exactly once, at U = −δ1/δ2. Intersecting these thresholds over all neighbours gives, for every bulk point, the range of U in which it is a local maximum (n-type) or local minimum (p-type). The union of all these ranges is where the detector is undepleted. Its outer bounds are returned, and the depletion voltage is the bound on the side of the bias polarity. No iteration is needed, and the result is exact for the given grid.

Note that neighbours with ϕV(x) ≈ ϕV(x₀) have an ordering that does not depend on U. These are now handled explicitly: depending on the sign of δ1 they rule out the point being a local minimum or maximum for any U. Before they were skipped, which could wrongly keep a point as a candidate. **In many of these cases an Inf would survive to the final Umax/min vector, causing the function to return no candidates in range. This was particularly problematic with Float32, where `_find_depletion_voltage_candidates` would return (-Inf,Inf) in most tested cases.** 

The bisection fallback also got stricter. Besides `ϕmax − ϕmin < |U|` it now checks that no bulk point is a local extremum of ϕ(U) via the new `_has_local_extremum`.

## Changes

- **`estimate_depletion_voltage`**: scan first, bisection only as fallback. Type-stable result (`U::T`).
- **`_find_depletion_voltage_candidates`**: handle neighbours with equal weighting potential, return only the outer bounds instead of per-point vectors, add docstring.
- **`_find_depletion_voltage_by_bisection`**: the old bisection loop moved into its own function, now with the local-extremum check.
- **`_has_local_extremum`**: new helper, checks the face neighbours of every bulk point.

## Testing

Added tests that check that the bisection and candidate results match within 5 V.
To further test the performance, the example IC's impurity profile was scaled. The calculated depletion voltage should be linear with this scaling.

###  Float32:

The old code clearly flip-flops between an incorrectly calculated bisection result (lacked local extremum check) and the candidate result. 
Screenshot 2026-09-16 at 18 13 24

At low scalings the flip-flop effect is even more pronounced. In this case the bisection loses power at very low scalings in old and new code. However candidates are always returned for the new code, leading to a better overall performance even at very low impurities with Float32.

Screenshot 2026-09-16 at 18 15 37

###  Float64:

The final result is identical for Float64. But the bisection underestimates the candidate depletion voltage in all cases.

Screenshot 2026-09-16 at 18 20 17

In the old code the maximum of the bisection and candidate was chosen causing the bisection to take over at low scalings. Now the candidate result is used by default, only falling back on a bisection if it fails (no such cases found so far).

Screenshot 2026-09-16 at 18 17 10


fhagemann:
Thanks! Let me have a look at this later this week, once I find the time.
Will you take care of updating the tests (seems like some depletion voltage values change, but that seems to be expected)?

fhagemann:
This PR seems to be closely related to #634, so we might want to coordinate the efforts in both PRs.

src/Simulation/Depletion.jl
```diff
@@ -229,7 +229,7 @@ function _find_depletion_voltage_candidates(ϕᵨ::AbstractArray{T, 3}, ϕᵥ::A
             δ2 = ϕᵥ₀ - ϕᵥ[x]
             # min -> p-type: Find U range that solves δϕ = δ1 + U*δ2 < 0  
             # max -> n-type: Find U range that solves δϕ = δ1 + U*δ2 > 0 
-            if isapprox(ϕᵥ₀, ϕᵥ[x]) # ordering independent of U
+            if isapprox(ϕᵥ₀, ϕᵥ[x], rtol = eps(T)) # ordering independent of U
```
fhagemann: Is this intended? The default for `rtol` is `sqrt(eps(T))` I believe, and this would make the `rtol` smaller (?)

src/Simulation/Depletion.jl
```diff
@@ -119,80 +119,122 @@ function estimate_depletion_voltage(sim::Simulation{T},
...
+            @info """The local extremum scan returned no candidates (or no unique candidate). Using alternate bisection calculation.
+            The depletion voltage is $(round(U_bisection, digits = Int(ceil(-log10(tol))))) ± $(tol) $(internal_voltage_unit) applied to contact $(contact_id)."""
+        end
+        U_bisection
+    else
+        U_cand = only(U_cand)
+        if verbose
+            @info "The depletion voltage is $(round(U_cand, digits = 2)) $(internal_voltage_unit) applied to contact $(contact_id)."
+        end
+        U_cand
```
fhagemann, line 138: I would argue that, if we have multiple `depletion_voltage_candidates`, we should just take the most extreme in this case, instead of defaulting to bisection here.

PR #623 suggests using something like this instead:

```suggestion
    U_cand = filter(in(Urng), _find_depletion_voltage_candidates(ϕρ, ϕV, bulk))
    U::T = if length(U_cand) == 0 
        inside = findall((simDV.point_types .& 4 .> 0) .& (simDV.point_types .& inactive_layer_bit .== 0))
        U_bisection = _find_depletion_voltage_by_bisection(ϕρ, ϕV, inside, bulk, Umin, Umax, tol)
        if verbose
            @info """The local extremum scan returned no candidates. Using alternate bisection calculation.
            The depletion voltage is $(round(U_bisection, digits = Int(ceil(-log10(tol))))) ± $(tol) $(internal_voltage_unit) applied to contact $(contact_id)."""
        end
        U_bisection
    else
        U_cand = mean(Urng) > 0 ? max(U_cand...) : min(U_cand...)
        if verbose
            @info "The depletion voltage is $(round(U_cand, digits = 2)) $(internal_voltage_unit) applied to contact $(contact_id)."
        end
        U_cand
```

fhagemann:
Nice, the tests seem to pass with the `r0_boundary_handling` fix!
We might want to look into running the tests also with `Float64` in the future (maybe just one of them can run on `Float64` instead of `Float32`), but that does not have to be part of this PR.

fhagemann (approved):

test/test_real_detectors.jl
```diff
@@ -25,13 +25,13 @@ T = Float32
...
     deplV = timed_estimate_depletion_voltage(sim, verbose = false, check_for_depletion = false)
     @info "Depletion voltage: $deplV"
-    @test isapprox(deplV, 1871*u"V", atol = 10.0*u"V") 
+    @test isapprox(deplV, 2030*u"V", atol = 10.0*u"V") 
     id = SolidStateDetectors.determine_bias_voltage_contact_id(sim.detector)
-    # Check wether detector is undepleted, 10V below the previously calculated depletion voltage
-    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = ustrip(deplV - deplV*0.005))
+    # Check wether detector is undepleted, 1% below the previously calculated depletion voltage
+    sim.detector = SolidStateDetector(sim.detector, contact_id = id, contact_potential = ustrip(deplV - deplV*0.01))
     timed_calculate_electric_potential!(sim, depletion_handling = true)
```
hervasa2, line 32: Maybe we want to simulate all 3 in the same custom grid?

fhagemann: In this PR?

fhagemann: Seems like I can keep the 0.5% if I add an additional refinement step to the tests here (bringing it closer to the same grid, without taking forever to run).

PR #643, opened 2026-09-19
Re-enable the 19-contact Coax testset

### LINKED ISSUE #622: Re-enable the 19-contact Coax testset (currently fully commented out)
(Draft from the 2026-07 code review; posted on behalf of the review.)

The true-coaxial example with 19 contacts has its entire testset commented out
in `test/test_real_detectors.jl`, presumably because it was too slow at the
original resolution. As a result the many-contact code paths (weighting
potentials for a large contact count, per-contact grouping, the associated
plotting-free bookkeeping) are exercised nowhere in the suite — the review found
this is the only example exercising that regime.

Proposal: re-enable it on a deliberately coarse grid with loose tolerances
(e.g. only a coarse `max_tick_distance`, no refinement, asserting just
convergence, depletion status and total signal-charge sanity rather than tight
physics numbers). The point is that the 19-contact path runs at all, not that it
is precise. If even the coarse version is too slow for CI, running it only on
one CI job (nightly / one OS) would still be better than nothing.

fhagemann:
Addressing #622:
Previously commented out to make the tests finished faster, I would like to see if we can re-enable the 19-contact Coax testset when using a loose convergence requirement for fast field calculation.

EDIT:
~Also: julia-1.10 tests were only run on 1 thread instead of 4 (I think this had to do with this being the codecov test and it actually being faster on 1 thread than on multiple), but I'm also increasing the number of threads to 4 here to see what the effect on the test runtime is.~


fhagemann:
Running the codecov tests on julia-1.10 with 4 threads seems to make it noticeably longer to run.
=> undoing this change

